PR #5, opened 2022-05-13
Initial commit

dmorokov:
1. Test Automation Framework skeleton in Java
2. Gradle config with most likely deps for tests
3. Template Test with place holders for further functionality
4. Logging props
5. Base main classes for testing and other utility needs
6. Code style in accordance with google-formatter-plugin
7. Also this PR is a verification step that EPAM team has accesses and permissions

### DIFF
--- a/dumper-integration-tests/redshift-tests/build.gradle
+++ b/dumper-integration-tests/redshift-tests/build.gradle
@@ -1,0 +1,49 @@
+buildscript {
+    repositories {
+        mavenCentral()
+    }
+    dependencies {
+        classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.18'
+    }
+}
+
+plugins {
+    id 'java'
+    id "com.google.protobuf" version "0.8.18"
+}
+
+sourceCompatibility = 1.8
+
+repositories {
+    mavenCentral()
+    maven {
+        url "https://redshift-maven-repository.s3-website-us-east-1.amazonaws.com/release"
+    }
+}
+
+dependencies {
+    implementation 'org.codehaus.groovy:groovy-all:3.0.5'
+    implementation 'org.testng:testng:7.5'
+    implementation 'org.slf4j:slf4j-api:2.0.0-alpha5'
+    implementation 'org.slf4j:slf4j-jdk14:2.0.0-alpha5'
+    implementation 'com.amazon.redshift:redshift-jdbc42:2.1.0.6'
+    implementation 'commons-io:commons-io:2.11.0'
+    implementation 'org.apache.commons:commons-collections4:4.4'
+    implementation 'com.google.guava:guava:31.0.1-jre'
+    implementation 'com.opencsv:opencsv:5.6'
+    implementation 'com.google.protobuf:protobuf-gradle-plugin:0.8.18'
+    implementation 'com.google.protobuf:protobuf-java:3.20.1'
+}
+
+test {
+    useTestNG {
+        preserveOrder true
+        systemProperty 'java.util.logging.config.file', 'src/main/resources/logging.properties'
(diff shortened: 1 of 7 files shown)


integration-tests/build.gradle
```diff
@@ -0,0 +1,31 @@
+plugins {
+    id 'groovy'
```
gregflack, line 2: Why you add groovy plugin - I don't see any .groovy files to run. This won't affect on building files it will use anyway installed groovy and gradle version on machine that will execute code

dumper-integration-tests/redshift-tests/src/main/java/com/google/sql/SqlUtil.java
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright 2021 Google LLC
```
gregflack, line 2: please update

integration-tests/build.gradle
```diff
@@ -0,0 +1,31 @@
...
+repositories {
+    mavenCentral()
+    maven {
+        url "https://redshift-maven-repository.s3-website-us-east-1.amazonaws.com/release"
+    }
+}
+
+dependencies {
+    implementation 'org.codehaus.groovy:groovy-all:3.0.5'
+    implementation 'org.testng:testng:7.5'
```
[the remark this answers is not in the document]
dmorokov: Does that limit the scope of lib usage as it does `<scope>test</scope>` maven pom.xml option?

integration-tests/build.gradle
dmorokov: I added it because of the possibility of working with json in separate Groovy files.

dumper-integration-tests/redshift-tests/src/main/java/com/google/sql/SqlUtil.java
dmorokov: I opened a random Java class of Dumper package, copied the licence agreement from there and used it for my classes.
Please, tell me if we need to use any other header.

integration-tests/build.gradle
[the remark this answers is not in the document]
gregflack: yes, correct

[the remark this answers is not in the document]
dmorokov: Then we don't want it, because we use TestNG annotations in Test scope like `@Test` or `@BeforeClass`.
So, because of that I used just `implementation` here.

dumper-integration-tests/redshift-tests/build.gradle
```diff
@@ -0,0 +1,30 @@
...
+repositories {
+    mavenCentral()
+    maven {
+        url "https://redshift-maven-repository.s3-website-us-east-1.amazonaws.com/release"
+    }
+}
+
+dependencies {
+    implementation 'org.codehaus.groovy:groovy-all:3.0.5'
+    implementation 'org.testng:testng:7.5'
```
paolomorandini, line 16: `dumper` uses JUnit:
https://github.com/google/dwh-migration-tools/blob/844447c4fe60504580e816bae621b9e204ed75d4/dumper/buildSrc/src/main/groovy/dwh-migration-dumper.java-common-conventions.gradle#L110

Is there any specific need for `testng`?

dmorokov: Yes, it's way easier to configure test suites, pass arguments and run in parallel if it's needed.
Also extended scope of `@Before<>` annotations and `@DataProvider` come very handy.
JUint works well with Unit tests and TestNG with integration.
Unless Junit is a strong requirement, I'd go with TestNG.

bdw-g (approved):
No objections from my side

PR #8, opened 2022-05-16
[IO] Ensure the target path directory exists before trying to write it

bdw-g:
The Netezza dumper would attempt to create files in a subdirectory, but the underlying virtual file system would throw an exception if there was an attempt to make a file in a subdirectory that does not exist.

This patch ensures that the parent directory exists, and should thereby avoid that bug.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/io/FileSystemByteSink.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/io/FileSystemByteSink.java
@@ -39,10 +39,6 @@ public class FileSystemByteSink extends ByteSink {
 
     @Override
     public OutputStream openStream() throws IOException {
-        Path parent = path.getParent();
-        if (!Files.isDirectory(parent)) {
-            Files.createDirectory(parent);
-        }
         return Files.newOutputStream(path);
     }
 
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/io/FileSystemByteSink.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/io/FileSystemByteSink.java
@@ -39,6 +39,10 @@ public class FileSystemByteSink extends ByteSink {
 
     @Override
     public OutputStream openStream() throws IOException {
+        Path parent = path.getParent();
+        if (!Files.isDirectory(parent)) {
+            Files.createDirectory(parent);
+        }
         return Files.newOutputStream(path);
     }
 
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/netezza/NetezzaMetadataConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/netezza/NetezzaMetadataConnector.java
@@ -172,6 +172,7 @@ public class NetezzaMetadataConnector extends AbstractJdbcConnector implements M
         // Not documented.
         out.add(new JdbcSelectTask("nz.v_objects.csv", "SELECT * FROM system.._v_objects"));
 
+        // TODO; these might be placed in a ParallelTaskGroup?`
         // these neeed to be filtered on WHERE = dbname, or else which DB md table will contain SYSTEM data too
         for (String db : dbs) {
             // The benefit of having this reduces the amount of data in the zip file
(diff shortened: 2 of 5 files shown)


paolomorandini (approved):

bdw-g:
Let's delay a merge until we can get a positive confirmation that this solves the issue

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/io/FileSystemByteSink.java
```diff
@@ -39,6 +39,10 @@ public FileSystemByteSink(@Nonnull Path path) {
 
     @Override
     public OutputStream openStream() throws IOException {
+        Path parent = path.getParent();
+        if (!Files.isDirectory(parent)) {
+            Files.createDirectory(parent);
```
Hackerpilot, line 44: Should this be `createDirectories` instead, to create more than one level of parent directory if needed?

genemerrill:
@bdw-g In order to confirm with customer, should I build based on `bdw-g:netezza-connector-fix` and deliver to customer to try?

Since there is not a constant stream of NZ metadata dumps happening, I would actually be okay with merging and telling this customer to try it out. But happy to make a custom build if that is what is required here.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/io/FileSystemByteSink.java
shevek-google: Right now, I'm more worried about this than having an active test. A unit test on FileSystemByteSink would totally work, though.

shevek-google (changes requested):
Question: To match the semantics of Files.asByteSink(File), should the createDirectories() be done in FileSystemOutputHandle.as[Temporary]ByteSink() ?

bdw-g:
> Question: To match the semantics of Files.asByteSink(File), should the createDirectories() be done in FileSystemOutputHandle.as[Temporary]ByteSink() ?

`Files.asByteSink(File)` really does nothing of the kind: https://source.corp.google.com/piper///depot/google3/third_party/java_src/google_common/current/java/com/google/common/io/Files.java;l=196?ct=os

... but having said that I think I would agree that this could be someplace else? Though I will say - most of this code seems to be deferring actual IO to the latest possible moment - that seems to be the reason for the many reasons of indirection - so 'someplace else' will also mean 'some time prior to starting writng'..

bdw-g:
> @bdw-g In order to confirm with customer, should I build based on `bdw-g:netezza-connector-fix` and deliver to customer to try?
> 
> Since there is not a constant stream of NZ metadata dumps happening, I would actually be okay with merging and telling this customer to try it out. But happy to make a custom build if that is what is required here.

Yes, a custom build would do.

bdw-g:
@shevek-google does the current change work for you?

shevek-google (changes requested):
I'd just like FileSystemOutputHandle to call createDirectories before creating the Sink, please.

The current change where there's a new task, I think, is not guaranteed to run synchronously or in order, as there isn't such a guarantee.

bdw-g:
> I'd just like FileSystemOutputHandle to call createDirectories before creating the Sink, please.
> 
> The current change where there's a new task, I think, is not guaranteed to run synchronously or in order, as there isn't such a guarantee.

Creating a sink is not (normally) a throwable operation, nor even an IO operation, but creating directories is, as is opening the output stream. (Which is indeed why I placed the operation there in the first place).

E.g. `Files.asByteSink` does not throw, nor does `FileSystems.asByteSink`, neither `MoreFiles.asByteSink`. So having the `OutputHandle.asByteSink` throw would be really odd.

And in fact, the `MetadataDumper` does run tasks sequentially (this might as well be 'guaranteed'); the only exception is `ParallelTaskGroup`, which is something you'd have to create explicitly in a connector.

But... this does suggest that the ideal place to handle this, is in `AbstractTask.run()`, since that already expects to run a throwable operation, and has the `TaskRunContext` / `OutputHandleFactory`, and is responsible for creating the `OutputHandle` and associated `ByteSink`. So having it invoke `createDirectories()` explicitly would be okay, I think.

It's quite unclear to me why we should even have the `OutputHandle` as a distinct thing from `ByteSink`, but there we are.

A propos, we could also implement this without the `ZipFileSystem` backend, using the commons-compress `ZipFile` on a `FileChannel` (memory map).

shevek-google:
I think memory-mapped file channels are fixed size, so one can't add a file to it.

bdw-g:
> I think memory-mapped file channels are fixed size, so one can't add a file to it.

In practice you'd write to a `FileChannel` and you can definitely append to those :-)

genemerrill:
FYI here's my attempt to create a build for a customer to validate the fix; https://drive.google.com/file/d/14V3aN_EdIKePrljPBUbXajgVmU-iKKCL/view?usp=sharing

Could we publish as a preview or beta release in the future via the GitHub project?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/io/FileSystemByteSink.java
bdw-g: Code was changed (and we now use `Files.createDirectories` instead :-)

shevek-google (changes requested):
Can we please make asByteSink() and friends throw the exception?

shevek-google:
Please see #8

paolomorandini:
#34 closes this (if approved)

PR #4, opened 2022-05-13
Add a python exemplary client tool.

xiashang0624:
Note that the macros_replacement_map setting is moved away from the main config file as an optional setting.  User needs to pass the file as a flag to enable this feature. e.g., "python main.py -m macros_replacement_map.yaml"

The README file will be added in a follow-up PR.

Closes #9

PR #9, opened 2022-05-17
Exemplary python client

shevek-google:
* Shared, mutable pull-request
* Closes #4


shevek-google:
@xiashang0624  I think our best approach from here is a small rearchitecture of the macro processing, as we have two classes involved, and we don't expose a good user API, but other than that, this is becoming solid.

genemerrill:
This branch results in a Inconsistent Tabs error, but Xia's fork version (pull/4) runs fine and looks ready to merge to me.

PR #13, opened 2022-05-21
Improve file path reading in the client tool.

xiashang0624:
This PR has the following changes:
* Move Input/output dir from the config.yaml file to args so that the behavior is not affected by where the user runs the binary.  Default path: client/input and client/output. 
* Add an optional arg that allows user to specify the path to the config file.
* I found a bug in the macro substitution about file name matching for sub-dirs in the input/output and fixed it.
* Improve the progress logs in the console to make it more readable to users.


client/README.md
```diff
@@ -73,11 +73,11 @@ bin/dwh-migration-client
 This tool can also perform macros substitution before/after the translation job
 through an option flag.
 
-To enable macros substitution, pass the arg '-m macros.yaml' when
+To enable macros substitution, pass the arg '-m client/macros.yaml' when
 running the tool:
 
 ```
-bin/dwh-migration-client -m macros.yaml
+bin/dwh-migration-client -m client/macros.yaml
```
paolomorandini, line 80: :+1: thank you

paolomorandini (commented):
Another possible approach could be to change
```
bin/dwh-migration-client
```
to to a 
```
(cd client; python3 main.py...)
```
So we don't need to handle two possible `pwd` for users?

client/README.md
```diff
@@ -62,7 +62,7 @@ the output directory.
 
 ## Run a translation job
 
-Simply run the following commands in Python3:
+Simply run the following commands in using Python3:
```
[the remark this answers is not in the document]
xiashang0624: Done. Reverted the change.  Thanks for catching this.

genemerrill (approved):
wfm

xiashang0624:
> wfm

client/macro_processor.py
```diff
@@ -125,7 +128,7 @@ def postprocess_file(self, tmp_path: str, output_path: str):
         print("Postprocessing into %s" % output_path)
         with open(tmp_path) as tmp_fh:
             text = tmp_fh.read()
-        text = self.postprocess_text(text, output_path)
+        text = self.postprocess_text(text, self.__get_path_in_tmp_output(tmp_path))
```
shevek-google, line 131: The path passed here was meant to be advisory, so the user knows which file is being processed in case that matters to their processing. It would seem to make more sense to pass the authoritative input/output path than the temp path. What do you think? Are we using the path in some way that makes us want to expose the temp path to the user? The user shouldn't be doing any I/O on that file. They might, however, want to read a sibling file which isn't in the temp directory.

client/macro_processor.py
```diff
@@ -184,7 +203,13 @@ def __parse_macros_config_file(self) -> Dict[str, Dict[str, str]]:
...
                 extension of ".sql".
         """
-        assert isfile(self.yaml_file_path), "Can't find a file at \"%s\"." % self.yaml_file_path
+        # Try to read the yaml file from either the root dir or the root/client dir.
+        if not isfile(self.yaml_file_path):
+            maybe_file_path = os.path.join("client", self.yaml_file_path)
+            if not isfile(maybe_file_path):
+                raise ValueError("Can't find a file at \"%s\"." % self.yaml_file_path)
+            else:
+                self.yaml_file_path = maybe_file_path
```
shevek-google, line 212: I'm a bit iffy on this happening down here. This should have happened at a higher level.

shevek-google (dismissed):

client/macro_processor.py
xiashang0624: Makes sense.  Done. I updated this part a little to stay away from the temp path. 
The tricky thing is that we need to use the file path that is relative to the config input/output directory (not cwd) so that the wildcard matching works for sub-directories, e.g, subdir_1/subdir_2/*.sql.

client/macro_processor.py
xiashang0624: Done. Now all the args and file-path level validations happen at the very beginning stage, via a validate_args() method.

client/main.py
```diff
@@ -28,14 +28,18 @@
                          'the program will revert the substitutions for all the output query files in a '
                          'post-processing step.  The replacement does not apply for files with extension of '
                          '.zip, .csv, .json.')
+parser.add_argument('-c', '--config', type=str, default="config.yaml",
+                    help='Path to the config.yaml file. (default: config.yaml)')
```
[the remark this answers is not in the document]
xiashang0624: Done.

client/config_parser.py
```diff
@@ -92,10 +96,14 @@ def parse_config(self, config_file: str = 'config.yaml') -> TranslationConfig:
...
         config.location = translation_config_input["location"]
         config.translation_type = translation_config_input[self.__TRANSLATION_TYPE]
-        config.input_directory = self.__DEFAULT_INPUT_DIR if self.__INPUT_DIR not in translation_config_input \
+        input_directory = self.__DEFAULT_INPUT_DIR if self.__INPUT_DIR not in translation_config_input \
             else translation_config_input[self.__INPUT_DIR]
+        config.input_directory = validate_path(input_directory)
         config.output_directory = self.__DEFAULT_OUTPUT_DIR if self.__OUTPUT_DIR not in translation_config_input \
             else translation_config_input[self.__OUTPUT_DIR]
+        if os.getcwd().endswith(REPO_ROOT_DIR) and config.output_directory == self.__DEFAULT_OUTPUT_DIR:
+            config.output_directory = join(CLIENT_DIR, self.__DEFAULT_OUTPUT_DIR)
```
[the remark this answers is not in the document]
xiashang0624: Agreed.  I updated this part so that users can run the tool from any directories.  If cwd is not the root, user should specify the path info through args. Otherwise, there will be warnings and the follow-up validations will likely to fail.

client/config_parser.py
```diff
@@ -16,7 +16,11 @@
 import os
 
 from argparse import Namespace
+<<<<<<< HEAD
+from os.path import exists, abspath
+=======
 from object_mapping_parser import ObjectMappingParser
+>>>>>>> main
```
paolomorandini, line 23: Unresolved conflict.

paolomorandini (changes requested):

client/config_parser.py
xiashang0624: I missed that. Thanks for catching it. Done.

ajwelch4 (changes requested):
If there's no rush to merge, I want to propose a slightly different way to deal with the confusion around calling the binary, PWD and the location of the default input/output/config files.

I'll check out this branch in my own repo and build on it to demonstrate.

client/batch_sql_translator.py
```diff
@@ -56,36 +56,37 @@ def start_translation(self):
...
-            print("Start post-processing by reverting the macros substitution...")
+            print("\nStart post-processing by reverting the macros substitution...")
             self.preprocessor.postprocess(local_output_dir, self.config.output_directory)
 
-        print("Finished post-processing. The output query files are in %s" % self.config.output_directory)
+        print("Finished postprocessing. The outputs are in %s\n" % self.config.output_directory)
 
-        if self.config.clean_up_tmp_files:
-            print("Cleaning up tmp files...")
+        if self.config.clean_up_tmp_files and exit(self.tmp_dir):
```
ajwelch4, line 81: I'm guessing this should have been `os.path.exists` not `exit`?

ajwelch4 (changes requested):

ajwelch4:
Here are my proposed changes to this branch:

https://github.com/xiashang0624/dwh-migration-tools/compare/client-path-fix...ajwelch4:client-path-fix

First, we should instruct the user to use a [virtualenv](https://docs.python.org/3/library/venv.html) to prevent dependency conflicts between our tool and the tools installed at the system-level by the system package manager. This goes without saying to Python folks as it's just good hygiene but not all of our customers will know this. It's particularly important in our case because we are asking our users to install `gcloud` and this is usually installed at the system-level.  `gcloud` is written in Python and likely uses many of the same Google Cloud Python libs that we are and we can't count on users having the same version of `gcloud` and its transitive dependencies installed as we do. Most folks use [Pipenv](https://pipenv.pypa.io/en/latest/) or [Poetry](https://python-poetry.org/) nowadays to manage virtualenvs but the simplest way to convey the recommendation is to show the old fashioned way:

```shell
python3 -m venv venv
source venv/bin/activate
```

Pip, Pipenv and Poetry would also take care of downloading our tool if we were publishing it to [PyPI](https://pypi.org/) but we are not so `git clone` it is:

```shelll
git clone git@github.com:google/dwh-migration-tools.git
```

At this point, our PWD looks like this:

```shell
tree -L 1 

.
├── dwh-migration-tools
└── venv
```

If the user doesn't plan on editing the tool and just wants to install it and get going:

```shell
pip3 install ./dwh-migration-tools/client
```

The above will invoke `pip`s dependency resolver to install a set of concrete dependencies from the set of [abstract dependencies](https://pipenv.pypa.io/en/latest/advanced/#pipfile-vs-setup-py) listed in `setup.py`.:

https://github.com/ajwelch4/dwh-migration-tools/blob/client-path-fix/client/setup.py#L23-L33

Using abstract dependencies in `setup.py` will allow us to support library use cases and PyPI distribution someday if we so chose. I would recommend tests before chosing so ;-). PSO is already investigating how to string together this tool, DTS, DVT, etc. so the library use case may be useful.

If the user wants to ensure they get the same exact concrete dependencies as us, they can run:

```shell
pip3 install -r ./dwh-migration-tools/client/requirements.txt
pip3 install ./dwh-migration-tools/client
```

Re: populating `requirements.txt`. The Python world has largely moved to lock files for concrete dependency management and Pipenv and Poetry would generate a lock file for us from the abstract dependencies. Instead, we will generate `requirements.text` from the abstract dependencies in `setup.py` by running:

```shell
make requirements
```

If the user wants to edit the tool, they can install it in editable mode using `pip`'s `-e` option:

```shell
pip3 install -e ./dwh-migration-tools/client
```

Another benefit of using `setup.py` is we can let it generate and install an executable for us instead of hard coding one at `dwh-migration-tools/bin.dwh-migration-client`. The `setup.py` generated executable will be put in the correct virtualenv or non-virtualenv location so that it's on the shell `PATH` and can be called without having to remember where the `bin` directory is. It will also respect `pip`'s `-e` editable mode option. We tell `setup.py` to do this for us via `console_scripts`:

https://github.com/ajwelch4/dwh-migration-tools/blob/client-path-fix/client/setup.py#L36-L38

FInally, instead of telling users to edit the config files and input dir in the source tree directly, we provide them a simple `init` command that sets up the project scaffolding for them in a directory of their choosing:

```shell
dwh-migration-client init my_project
```

Our PWD now looks like this:

```shell
tree -L 1 

.
├── dwh-migration-tools
├── my_project
└── venv
```

And if we `cd` into `my_project` we see:

```shell
tree

.
├── config.yaml
├── input
│   ├── 1.sql
│   ├── 2.sql
│   ├── google-examples.sql
│   └── subdir
│       └── 3.sql
├── macros.yaml
└── object_mapping.json
```

Executing the tanslation job is done via:

```shell
dwh-migration-client translate
```

or

```shell
dwh-migration-client translate --config config.yaml --input input --output output -m macros.yaml -o object_mapping.json     
```

or some variant thereof.

client/batch_sql_translator.py
xiashang0624: Good catch. Done. Thanks!

xiashang0624:
> Here are my proposed changes to this branch:
> 
> [xiashang0624/dwh-migration-tools@client-path-fix...ajwelch4:client-path-fix](https://github.com/xiashang0624/dwh-migration-tools/compare/client-path-fix...ajwelch4:client-path-fix)
> 
> First, we should instruct the user to use a [virtualenv](https://docs.python.org/3/library/venv.html) to prevent dependency conflicts between our tool and the tools installed at the system-level by the system package manager. This goes without saying to Python folks as it's just good hygiene but not all of our customers will know this. It's particularly important in our case because we are asking our users to install `gcloud` and this is usually installed at the system-level. `gcloud` is written in Python and likely uses many of the same Google Cloud Python libs that we are and we can't count on users having the same version of `gcloud` and its transitive dependencies installed as we do. Most folks use [Pipenv](https://pipenv.pypa.io/en/latest/) or [Poetry](https://python-poetry.org/) nowadays to manage virtualenvs and store them in `~/.local/share/virtualenvs` but the simplest way to convey the recommendation is to show the old fashioned way:
> 
> ```shell
> python3 -m venv venv
> source venv/bin/activate
> ```
> 
> Pip, Pipenv and Poetry would also take care of downloading our tool if we were publishing it to [PyPI](https://pypi.org/) but we are not so `git clone` it is:
> 
> ```
> git clone git@github.com:google/dwh-migration-tools.git
> ```
> 
> At this point, our PWD looks like this:
> 
> ```shell
> tree -L 1 
> 
> .
> ├── dwh-migration-tools
> └── venv
> ```
> 
> If the user doesn't plan on editing the tool and just wants to install it and get going:
> 
> ```shell
> pip3 install ./dwh-migration-tools/client
> ```
> 
> The above will invoke `pip`s dependency resolver to install a set of concrete dependencies from the set of [abstract dependencies](https://pipenv.pypa.io/en/latest/advanced/#pipfile-vs-setup-py) listed in `setup.py`:
> 
> https://github.com/ajwelch4/dwh-migration-tools/blob/client-path-fix/client/setup.py#L23-L33
> 
> Using abstract dependencies in `setup.py` will allow us to support library use cases and PyPI distribution someday if we so choose. I would recommend tests before choosing so ;-). PSO is already investigating how to string together this tool, DTS, DVT, etc. so the library use case may be useful.
> 
> If the user wants to ensure they get the same exact concrete dependencies as us, they can run:
> 
> ```shell
> pip3 install -r ./dwh-migration-tools/client/requirements.txt
> pip3 install ./dwh-migration-tools/client
> ```
> 
> Re: populating `requirements.txt`. The Python world has largely moved to lock files for concrete dependency management and Pipenv and Poetry would generate a lock file for us from the abstract dependencies. Instead, we will generate `requirements.text` from the abstract dependencies in `setup.py` by running:
> 
> ```shell
> make requirements
> ```
> 
> If the user wants to edit the tool, they can install it in editable mode using `pip`'s `-e` option:
> 
> ```shell
> pip3 install -e ./dwh-migration-tools/client
> ```
> 
> Another benefit of using `setup.py` is we can let it generate and install an executable for us instead of hard coding one at `dwh-migration-tools/bin/dwh-migration-client`. The `setup.py` generated executable will be put in the correct virtualenv or non-virtualenv location so that it's on the shell `PATH` and can be called without having to remember where the `bin` directory is. It will also respect `pip`'s `-e` editable mode option. We tell `setup.py` to do this for us via `console_scripts`:
> 
> https://github.com/ajwelch4/dwh-migration-tools/blob/client-path-fix/client/setup.py#L36-L38
> 
> `setup.py` lives in the client root directory. All other Python files have been moved into the `dwh_migration_client` package. It's pretty standard for the root directory of a Python project to only contain `setup.py`, markdown files, config files for formatters, linters, type checkers etc.
> 
> Finally, instead of telling users to edit the config files and input dir in the source tree directly, we provide them a simple `init` command that sets up the project scaffolding for them in a directory of their choosing:
> 
> ```shell
> dwh-migration-client init my_project
> ```
> 
> Our PWD now looks like this:
> 
> ```shell
> tree -L 1 
> 
> .
> ├── dwh-migration-tools
> ├── my_project
> └── venv
> ```
> 
> And if we `cd` into `my_project` we see:
> 
> ```shell
> tree
> 
> .
> ├── config.yaml
> ├── input
> │   ├── 1.sql
> │   ├── 2.sql
> │   ├── google-examples.sql
> │   └── subdir
> │       └── 3.sql
> ├── macros.yaml
> └── object_mapping.json
> ```
> 
> Executing the translation job is done via:
> 
> ```shell
> dwh-migration-client translate
> ```
> 
> or
> 
> ```shell
> dwh-migration-client translate --config config.yaml --input input --output output -m macros.yaml -o object_mapping.json     
> ```
> 
> or some variant thereof.

Thanks for the review and I learnt a lot!

These are great suggestions and personally I REALLY like all of them from the perspective of engineering excellency. 

First, I don't think this PR is conflicting with your suggestions so I would recommend to handle them in a following PR, unless you disagree, of course.

Here are my additional thoughts:

If the target audiences of this tool are experienced Python engineers, we should just do everything you suggested.

But I believe we are trying to make this tool so simple that people with almost no Python or coding experience can pick it up and run a translation job within just a few minutes. For this reason, I think we should mention `virtualenv` at the bottom of README in a "Common Questions" or "Recommendations" section instead of in the beginning.

The current **basic** user journey has 4 simple steps: people download this package from GH -> go to the root (dwh-migration-client) -> pip3 install requirements -> run the tool.

After a successful run, user should be happy and they can keep reading the rest of the doc about other features we provide.

I am not in favor of adding additional steps in the **basic** user journey. But I think we probably could replace "pip3 install client/requirements.txt" with "pip3 install dwh-migration-tools/client" after adding a setup.py file (in a follow-up PR).

And about "after user installs `./dwh-migration-tools/client`, they can just call `dwh-migration-tools` from any directory without memorizing the path to the bin".  I don't think this will help with the issue this PR is trying to solve about the default path to the config files.  For example, if the user runs `dwh-migration-tools` from `Users/xxx/Desktop`, they still need to override the path to "config.yaml", "input", etc, right?  So I think this PR is still needed.  What do you think?

And about the logic of macros pre/post processing, in the new structure, do you think if we could also move `macro_processor.py` to the config folder with the other config files?  This is because we also expect users to change that python file, e.g., maybe they want to use a different postprocessing logic.  And this is for advanced users. They probably can figure out what file to change so it's not a high-priority.

client/batch_sql_translator.py
```diff
@@ -56,36 +56,37 @@ def start_translation(self):
...
+        print("\nDownloading outputs...")
         gcs_util.download_directory(local_output_dir, self.config.gcs_bucket, join(self.gcs_path, "output"))
 
         if self.preprocessor is not None:
-            print("Start post-processing by reverting the macros substitution...")
+            print("\nStart post-processing by reverting the macros substitution...")
             self.preprocessor.postprocess(local_output_dir, self.config.output_directory)
 
-        print("Finished post-processing. The output query files are in %s" % self.config.output_directory)
+        print("Finished postprocessing. The outputs are in %s\n" % self.config.output_directory)
```
paolomorandini, line 79: Can we use f-strings (PEP 498)?

paolomorandini: My fault, I should have checked which version introduced this. Thank you.

xiashang0624:
> @xiashang0624 Thanks for the detailed response! I agree with most of it. Much of this can be moved to follow-on PRs as suggested, so I've broken out my response below accordingly.
> 
> # This PR
> ## Checking PWD is repo root if CLI arg defaults are used
> > And about "after user installs `./dwh-migration-tools/client`, they can just call `dwh-migration-tools` from any directory without memorizing the path to the bin".  I don't think this will help with the issue this PR is trying to solve about the default path to the config files.  For example, if the user runs `dwh-migration-tools` from `Users/xxx/Desktop`, they still need to override the path to "config.yaml", "input", etc, right?  So I think this PR is still needed.  What do you think?
> 
> I don't think the tool should check what directory it is in. I think it should validate the existence of the paths passed as CLI args (either by default or explicitly).
> 
> I don't know of any other tool that requires the user to be in a specific directory because of default settings. Every other tool I know just assumes the user is aware of the defaults and validates that they exist if they are file paths.
> 
> Also, the check as currently implemented isn't foolproof:
> 
> ```python
>     if not os.getcwd().endswith(REPO_ROOT_DIR):
>         if not args.config:
>             print("Warning: you are not running the binary from the repo's root dir \"%s\" and you did not "
>                   "pass a config file path through \"--config\". Trying to use the default config path \"%s\"."
>                   % (REPO_ROOT_DIR, DEFAULT_CONFIG_PATH))
>         if not args.input:
>             print("Warning: you are not running the binary from the repo's root dir \"%s\" and you did not "
>                   "pass an input path through \"--input\". Trying to use the default input path \"%s\"." %
>                   (REPO_ROOT_DIR, DEFAULT_INPUT_PATH))
>         if not args.output:
>             print("Warning: you are not running the binary from the repo's root dir \"%s\" and you did not "
>                   "pass an output path through \"--output\". Trying to use the default output path \"%s\"." %
>                   (REPO_ROOT_DIR, DEFAULT_OUTPUT_PATH))
> ```
> 
> If someone does `git clone git@github.com:google/dwh-migration-tools.git dwh-migration-tools-2`, the warning is wrong. Also, although less likely, the lack of warning is wrong if someone runs the tool from a directory named `dwh-migration-tools` that is not the repo.
> 
> Here's a proposed implementation of simply validating the CLI args that builds on this branch: [xiashang0624/dwh-migration-tools@client-path-fix...ajwelch4:aj-validate-paths](https://github.com/xiashang0624/dwh-migration-tools/compare/client-path-fix...ajwelch4:aj-validate-paths).
> 
> # Follow-on PRs
> ## User journey: getting started
> > But I believe we are trying to make this tool so simple that people with almost no Python or coding experience can pick it up and run a translation job within just a few minutes. For this reason, I think we should mention `virtualenv` at the bottom of README in a "Common Questions" or "Recommendations" section instead of in the beginning.
> > The current **basic** user journey has 4 simple steps: people download this package from GH -> go to the root (dwh-migration-client) -> pip3 install requirements -> run the tool.
> > After a successful run, user should be happy and they can keep reading the rest of the doc about other features we provide.
> > I am not in favor of adding additional steps in the **basic** user journey. But I think we probably could replace "pip3 install client/requirements.txt" with "pip3 install dwh-migration-tools/client" after adding a setup.py file (in a follow-up PR).
> 
> I am basically in agreement with everything above. In particular, I agree mentioning virtualenv is overkill. Just thought I should point out not doing this will likely cause issues some day. I've removed mention of it from the docs completely.
> 
> Where I disagree slightly is I think the user journey should be the following:
> 
> These 2 commands only need to be run once (and they would be 1 command if we distributed to PyPI):
> 
> ```shell
> git clone git@github.com:google/dwh-migration-tools.git
> pip3 install dwh-migration-tools/client
> ```
> 
> These 3 commands are the main user journey for a translation project:
> 
> ```
> dwh-migration-client init my_project
> cd my_project
> dwh-migration-client translate
> ```
> 
> This is a workflow that is very common. For example [DBT](https://github.com/dbt-labs/dbt-core), which is arguably one of the most popular CLI tools used by non-programmers who only know SQL, [uses this workflow](https://docs.getdbt.com/guides/getting-started/learning-more/getting-started-dbt-core#create-a-project).
> 
> The tradeoff as I see it is the user has to copy and paste one additional command (`dwh-migration-client init my_project`) but what they get is a clean, unintimidating project scaffolding. And they can create additional isolated translation projects without having to `git clone` and `pip install` each time.
> 
> I have a proposed implementation I will submit as a separate PR. See the sequential list of follow-on PR diffs below.
> 
> ## User journey: macro_processor.py
> > And about the logic of macros pre/post processing, in the new structure, do you think if we could also move `macro_processor.py` to the config folder with the other config files?  This is because we also expect users to change that python file, e.g., maybe they want to use a different postprocessing logic.  And this is for advanced users. They probably can figure out what file to change so it's not a high-priority.
> 
> This is a good point. We could just tell the user to fork the repo but that is more work for the user. `macro_processor.py` is basically a plugin. The user may want to plug in other types of processors to process esoteric BI/ETL tool exports for example. I've already started working on processor functionality similar to the ksh ingester and inplace writer from the CW binary. I think we should create a `Processor` abstract base class (or `Protocol` but I don't think they are supported on all versions of Python 3.0+) that the user can inject implementations of into the client.
> 
> Having all the file handling logic in `MacroProcessor` makes this difficult. I think `MacroProcessor` should be limited to the functionality of `MapBasedExpander` more or less. The file handling logic should be abstracted away into a `ProcessorPipeline` class. `ProcessorPipeline` would take a list of `Processor` implementations.
> 
> I have a proposed implementation. It uses a `processors.yaml` file to configure `ProcessorPipeline`. However, given each processor will likely require vastly different config, a Python file may be better than a YAML file. I will submit the proposed implementation as a separate PR. See the sequential list of follow-on PR diffs below.
> 
> ## Docs
> We may have to update the cloud.google.com docs if we take this approach? I'm happy to do that.
> 
> ## Sequential list of follow-on PR diffs
> * Formatter: [ajwelch4/dwh-migration-tools@aj-validate-paths...ajwelch4:black-formatter](https://github.com/ajwelch4/dwh-migration-tools/compare/aj-validate-paths...ajwelch4:black-formatter)
> * Linter: [ajwelch4/dwh-migration-tools@black-formatter...ajwelch4:pylint](https://github.com/ajwelch4/dwh-migration-tools/compare/black-formatter...ajwelch4:pylint)
> * Type checker: [ajwelch4/dwh-migration-tools@pylint...ajwelch4:mypy](https://github.com/ajwelch4/dwh-migration-tools/compare/pylint...ajwelch4:mypy)
> * User journey: getting started: [ajwelch4/dwh-migration-tools@mypy...ajwelch4:user-journey-getting-started](https://github.com/ajwelch4/dwh-migration-tools/compare/mypy...ajwelch4:user-journey-getting-started)
> * User journey: plugins: [ajwelch4/dwh-migration-tools@user-journey-getting-started...ajwelch4:user-journey-plugins](https://github.com/ajwelch4/dwh-migration-tools/compare/user-journey-getting-started...ajwelch4:user-journey-plugins)

Thanks for the explanation and the proposed change. It looks good to me. I merged your commit with a few minor changes:
* Refactored the args validation methods into a validation.py file.
* Added a section in README with all options.
* Deleted the `Print usage message if no args are supplied` step because `bin/dwh-migration-tool` is still the "quick-start" command in the current documentations.

For the follow-up PRs about user journey, the 'dwh-migration-client init my_project' step also makes sense to me now.  Thanks again for the explanation!  And thank you for drafting the PR!

About the abstract processors class, that was actually the initial plan ;) But we ended up with a simple `macros_processor.py` file because we (Shevek and I) expect people to read that file and we hope people (with minimum coding experience) is comfortable with modifying that python file directly.

But I really like the idea of providing other kinds of standard processing logics that users can just use to compose their own pre/post-processing pipeline.

ajwelch4 (changes requested):
Other than the old URL, LGTM!

client/batch_sql_translator.py
```diff
@@ -105,7 +105,7 @@ def __generate_gcs_path(self) -> str:
     def __get_ui_link(self) -> str:
         """Returns the http link to the offline translation page for this project.
         """
-        return ("https://console.cloud.google.com/bigquery/migrations/offline-translation?projectnumber=%s" %
+        return ("https://console.cloud.google.com/bigquery/migrations/offline-translation?project=%s" %
```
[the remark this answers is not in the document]
xiashang0624: Right. Updated.  Thanks

client/config.yaml
```diff
@@ -14,15 +14,12 @@
 
 gcp_settings:
   # GCP project number.
-  project_number: 'XXXXXXXXXX'
+  project_number: 'translation-test-external'
```
paolomorandini, line 17: I'm a bit surprised, `project_number` would make me thing only numeric IDs are allowed here

ajwelch4: I was actually thrown off by this as well.

paolomorandini (approved):

PR #15, opened 2022-05-24
Add support for ONM in the client tool

xiashang0624:
* The ONM can be enabled as an optional args like -o object_mapping.json.
* The json follows same format of the UI journey as documented in https://cloud.devsite.corp.google.com/bigquery/docs/output-name-mapping#json_file_format.  Users can just use the same json file they used in creating a translation job through the UI.


genemerrill (approved):
ran locally, wfm

yzhvictor (approved):

PR #17, opened 2022-05-26
Add support for default_database and schema_search_path

ajwelch4:
This PR allows users to specify `default_database` and `schema_search_path` in the `config.yaml` file.

I was planning on using this in the codelab I am working on.

### DIFF
--- a/client/batch_sql_translator.py
+++ b/client/batch_sql_translator.py
@@ -156,7 +156,11 @@ class BatchSqlTranslator:
             gcs_source_path=gcs_input_path,
             gcs_target_path=gcs_output_path,
             source_dialect=self.get_input_dialect(),
-            target_dialect=target_dialect
+            target_dialect=target_dialect,
+            source_env=bigquery_migration_v2.types.SourceEnv(
+                default_database=self.config.default_database,
+                schema_search_path=self.config.schema_search_path
+            )
         )
         if self.config.object_name_mapping_list:
             translation_config.name_mapping_list = self.config.object_name_mapping_list
--- a/client/batch_sql_translator.py
+++ b/client/batch_sql_translator.py
@@ -156,7 +156,7 @@ class BatchSqlTranslator:
             gcs_source_path=gcs_input_path,
             gcs_target_path=gcs_output_path,
             source_dialect=self.get_input_dialect(),
-            target_dialect=target_dialect,
+            target_dialect=target_dialect
         )
 
         if self.config.default_database or self.config.schema_search_path:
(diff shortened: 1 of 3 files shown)


client/config_parser.py
```diff
@@ -103,6 +109,10 @@ def parse_config(self, config_file: str = 'config.yaml') -> TranslationConfig:
             else translation_config_input[self.__INPUT_DIR]
         config.output_directory = self.__DEFAULT_OUTPUT_DIR if self.__OUTPUT_DIR not in translation_config_input \
             else translation_config_input[self.__OUTPUT_DIR]
+        config.default_database = self.__DEFAULT_DEFAULT_DATABASE if self.__DEFAULT_DATABASE not in translation_config_input \
```
xiashang0624, line 112: I think the default_database and schema_search_path are optional fields in the API.
Maybe only assign them if these fields showed up in the config.yaml file.

client/config_parser.py
```diff
@@ -58,11 +60,15 @@ def __init__(self, argument: Namespace):
...
     __INPUT_DIR = "input_directory"
     __OUTPUT_DIR = "output_directory"
+    __DEFAULT_DATABASE = "default_database"
+    __SCHEMA_SEARCH_PATH = "schema_search_path"
     __CLEAN_UP = "clean_up_tmp_files"
 
     # Config default values
     __DEFAULT_INPUT_DIR = "input"
     __DEFAULT_OUTPUT_DIR = "output"
+    __DEFAULT_DEFAULT_DATABASE = "__DEFAULT_DATABASE__"
```
xiashang0624, line 70: Since these are optional fields, I don't think we need a default value here.

client/config.yaml
```diff
@@ -29,5 +29,12 @@ translation_config:
   translation_type: Translation_Teradata2BQ
   location: 'us'
 
+  # The default database name to fully qualify SQL objects when their database name is missing.
```
xiashang0624, line 32: Maybe highlight this as an optional field?  What about putting [Optional field] here?

client/config.yaml
```diff
@@ -29,5 +29,12 @@ translation_config:
...
+  default_database: __DEFAULT_DATABASE__
+
+  # The schema search path. When SQL objects are missing schema name, translation engine will search through this list to find the value.
```
xiashang0624, line 35: Same as above. Maybe highlight this field is optional.

client/batch_sql_translator.py
```diff
@@ -156,7 +156,11 @@ def create_migration_workflow(self, gcs_input_path: str, gcs_output_path: str) -
             gcs_source_path=gcs_input_path,
             gcs_target_path=gcs_output_path,
             source_dialect=self.get_input_dialect(),
-            target_dialect=target_dialect
+            target_dialect=target_dialect,
+            source_env=bigquery_migration_v2.types.SourceEnv(
```
xiashang0624, line 160: Maybe add this field only when config.default_database or config.schema_search_path exist?

xiashang0624 (approved):
Thanks for drafting this PR! LGTM.
The comments are minor.

PR #23, opened 2022-06-06
Add extraction support for pid, label, and endtime for redshift-logs …

cyulysses-corp:
…connector.

xid to pid is one-to-many, and pid to queryid is one-to-many, so adding pid doesn't change total number of records or grouping.

However, changed how STL_QUERY is joined with STL_QUERYTEXT and now they are joined additionally on userid, xid, pid.

Have confirmed in group discussion that redshift-logs connector is currently not used by any downstream services so the changes shouldn't cause any breakage.

Tested manually with a real Redshift instance and the extracted results are as expected.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftLogsConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftLogsConnector.java
@@ -82,9 +82,11 @@ public class RedshiftLogsConnector extends AbstractRedshiftConnector implements 
         // DDL TEXT is simple ...
         // min() as there is no ANY() or SOME()
         String queryTemplateDDL
-                = " SELECT min(L.userid) as UserId, min(L.starttime) as StartTime, L.xid as xid, "
-                + " LISTAGG(CASE WHEN LEN(RTRIM(L.text)) = 0 THEN text ELSE RTRIM(L.text) END) WITHIN GROUP (ORDER BY L.sequence) as SqlText "
-                + " FROM STL_DDLTEXT L WHERE ## GROUP BY L.xid ";
+            = " SELECT any_value(L.userid) as UserId, any_value(L.starttime) as StartTime,"
+            + " any_value(L.endtime) as EndTime, any_value(trim(label)) as Label, L.xid as xid,"
+            + " L.pid as pid, LISTAGG(CASE WHEN LEN(RTRIM(L.text)) = 0 THEN text ELSE RTRIM(L.text)"
+            + " END) WITHIN GROUP (ORDER BY L.sequence) as SqlText FROM STL_DDLTEXT L WHERE ##"
+            + " GROUP BY L.xid, L.pid ";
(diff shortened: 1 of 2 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftLogsConnector.java
```diff
@@ -82,19 +82,26 @@ public void addTasksTo(List<? super Task<?>> out, ConnectorArguments arguments)
...
         // min() as there is no ANY() or SOME()
         String queryTemplateDDL
-                = " SELECT min(L.userid) as UserId, min(L.starttime) as StartTime, L.xid as xid, "
-                + " LISTAGG(CASE WHEN LEN(RTRIM(L.text)) = 0 THEN text ELSE RTRIM(L.text) END) WITHIN GROUP (ORDER BY L.sequence) as SqlText "
-                + " FROM STL_DDLTEXT L WHERE ## GROUP BY L.xid ";
+            = " SELECT any_value(L.userid) as UserId, any_value(L.starttime) as StartTime,"
+            + " any_value(L.endtime) as EndTime, any_value(trim(label)) as Label, L.xid as xid,"
+            + " L.pid as pid, LISTAGG(CASE WHEN LEN(RTRIM(L.text)) = 0 THEN text ELSE RTRIM(L.text)"
+            + " END) WITHIN GROUP (ORDER BY L.sequence) as SqlText FROM STL_DDLTEXT L WHERE ##"
+            + " GROUP BY L.xid, L.pid ";
```
paolomorandini, line 89: This formatting of the query is a bit difficult to read

paolomorandini, line 89: This formatting of the query is a bit difficult to read

bdw-g: Agreed

bdw-g (commented):
Minor, but wasn't `RedshiftLogsConnector` supposed to be deprecated?

shevek-google (commented):
This connector is deprecated because it does not, and I think can not work. What is the reason we are pursuing this change, please?

cyulysses-corp:
> This connector is deprecated because it does not, and I think can not work. What is the reason we are pursuing this change, please?

Because the non-working part is the query text concatenation, where as the other "log metadata" fields are intact. The redshift rawlogs-connector's output is un-aggregated - true that the metadata without query text can also be extracted from the rawlogs connector output just by deduplication, but it would greatly help with prototyping if we can save effort from that additional implementation for now.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftLogsConnector.java
cyulysses-corp: Done.

bdw-g (approved):
Reads nicer, thanks

PR #25, opened 2022-06-07
add location description for allowed regions

yzhvictor:


client/config.yaml
```diff
@@ -27,6 +27,9 @@ translation_config:
   # Translation_Oracle2BQ, Translation_HiveQL2BQ, Translation_SparkSQL2BQ, Translation_Snowflake2BQ,
   # Translation_Netezza2BQ, Translation_AzureSynapse2BQ, Translation_Vertica2BQ.
   translation_type: Translation_Teradata2BQ
+
+  # The location needs to be one of these: us, eu, asia-northeast1, asia-southeast1, europe-west2, asia-south1,
+  # australia-southeast1, southamerica-east1.
```
paolomorandini, line 32: should we give more context on the difference between `eu` and `europe-west2`?

xiashang0624 (approved):

client/config.yaml
yzhvictor: Should be fine, it matches the location of GCS buckets, it should be obvious to users.

paolomorandini (approved):

PR #33, opened 2022-06-10
Use black, the canonical python code formatter.

ajwelch4:
This PR adds `black` and `isort` to `requirements_dev.txt`. They can be called by running `make format`. See `CONTRIBUTING.md` for more info. `pylint` and `mypy` PRs to follow.


client/CONTRIBUTING.md
```diff
@@ -1 +1,21 @@
...
+## Install
+
+Install the client dependencies (including dev dependencies) in a 
+virtualenv.
+
+```shell
+git clone git@github.com:google/dwh-migration-tools.git
+cd dwh-migration-tools
+python3 -m venv venv
+source venv/bin/activate
```
shevek-google, line 12: This is excellent; could well be wrapped into a shell script rather than a C&P. Perhaps this could even be wrapped into bin/dwh-migration-client so that it's a total no-brainer.

shevek-google (approved):
The python formatter is dead. Long live the new python formatter.

client/Makefile
```diff
@@ -0,0 +1,6 @@
+SHELL := /bin/bash -ex
+
+.PHONY: format
+format:
+	black .
+	isort .
```
mjquinn-google, line 6: Does `black` have a mode that allows us to check whether anything is in need of formatting, similar to Spotless's `spotlessCheck` target? I'm thinking of a potential future in which we may want to check formatting/linting in CI.

mjquinn-google (approved):

client/CONTRIBUTING.md
shevek-google: Argh, I want to comment and approve, and I can't do that any more.

client/Makefile
ajwelch4: Yes: https://mail.google.com/mail/u/0/?zx=85jeogl5600k#inbox/FMfcgzGpGTBZgVPrLqxvHFWPfBpGDbrT

I thought of adding a GH action. Is that what we plan on using when folks submit a PR against this repo? 

In any case, we can add that in a future PR after pylint and mypy have been added as well.

ajwelch4: Sorry wrong link: https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#writeback-and-reporting

mjquinn-google: Ok sounds good. Not sure if there's any near-term intent to use Github Actions, but as you say, we can address that in a future PR.

PR #39, opened 2022-06-24
Use mypy, the canonical Python type checker.

ajwelch4:
Note that `mypy` is an optional type checker.  Users can still modify their forks without worrying about whether their modifications type check. It is however recommended to run `make check` before submitting a PR back upstream.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ build
 .gradle
 .idea
 venv
+.mypy_cache
+
--- a/client/Makefile
+++ b/client/Makefile
@@ -9,5 +9,9 @@ format:
 lint:
 	pylint *.py
 
+.PHONY: typecheck
+typecheck:
+	mypy --strict .
+
 .PHONY: check
-check: format lint
+check: format lint typecheck
--- a/client/batch_sql_translator.py
+++ b/client/batch_sql_translator.py
@@ -46,12 +46,11 @@ class BatchSqlTranslator:  # pylint: disable=too-many-instance-attributes
         output_directory: str,
         preprocessor: Optional[MacroProcessor] = None,
         object_name_mapping_list: Optional[ObjectMappingParser] = None,
-    ):
+    ) -> None:
         self.config = config
         self._input_directory = input_directory
         self._output_directory = output_directory
         self.client = bigquery_migration_v2.MigrationServiceClient()
-        self.gcs_path = None
         self.preprocessor = preprocessor  # May be None
         self._object_name_mapping_list = object_name_mapping_list
         self.tmp_dir = join(dirname(self._input_directory), self.__TMP_DIR_NAME)
@@ -66,7 +65,7 @@ class BatchSqlTranslator:  # pylint: disable=too-many-instance-attributes
     # This directory will be deleted (by default) after the job finishes.
     __TMP_DIR_NAME = ".tmp_processed"
 
-    def start_translation(self):
+    def start_translation(self) -> None:
         """Waits until the workflow finishes by calling the Migration Service API every
         5 seconds.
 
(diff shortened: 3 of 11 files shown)


bdw-g (commented):
Looks good, but I see that type annotation syntax is new since version 3.5.
I want to verify that we think this is a reasonable version to require

paolomorandini (approved):

Hackerpilot (approved):

shevek-google:
Note that Ubuntu Bionic LTS is still supported and uses Python 3.6.5-3.

Ubuntu Xenial and Trusty are in extended security maintenance only.

So, this is probably OK, but a bit tight for my liking.

ajwelch4:
@shevek-google According to the [README](https://github.com/googleapis/python-storage#unsupported-python-versions) for the GCS client lib:

> Python == 3.6: the last released version which supported Python 3.6 was google-cloud-storage 2.0.0, released 2022-01-12.
>
> Python == 3.5: the last released version which supported Python 3.5 was google-cloud-storage 1.32.0, released 2020-10-16.
>
> Python == 2.7: the last released version which supported Python 2.7 was google-cloud-storage 1.44.0, released 2022-01-05.

So if we're willing and able to downgrade the GCS client lib then we can support older versions of Python.#39 

@xiashang0624 I think you upgraded the GCS client lib? Do we need `2.3.0` or can we downgrade?

ajwelch4:
Also, I believe the type annotation syntax has been supported since 3.0. The `typing` module has only been available in the stdlib since 3.5. But there is a [backport](https://pypi.org/project/typing/) if we want to support older 3.x versions. See [here](https://stackoverflow.com/a/31029363) for more info. So I think we can support this PR for 3.0+.

I should have done [this](https://github.com/google/dwh-migration-tools/pull/39/files#diff-cfab7219da3bb36313aff4a78ea8a1e3219509ac34cfd2774021d92642ff6776L15-R63) in a separate PR. I just noticed that `pyproject.toml` was out of sync with @xiashang0624 's change to `README.md`.  I can take it out and we can address it in a separate PR that includes a downgrade of the GCS client lib if @xiashang0624 deems a downgrade worthwhile.

xiashang0624:
> Also, I believe the type annotation syntax has been supported since 3.0. The `typing` module has only been available in the stdlib since 3.5. But there is a [backport](https://pypi.org/project/typing/) if we want to support older 3.x versions. See [here](https://stackoverflow.com/a/31029363) for more info. So I think we can support this PR for 3.0+ but ideally I'd rather not go beyond 3.6.
> 
> I should have done [this](https://github.com/google/dwh-migration-tools/pull/39/files#diff-cfab7219da3bb36313aff4a78ea8a1e3219509ac34cfd2774021d92642ff6776L15-R63) in a separate PR. I just noticed that `pyproject.toml` was out of sync with @xiashang0624 's change to `README.md`. I can take it out and we can address it in a separate PR that includes a downgrade of the GCS client lib if @xiashang0624 deems a downgrade worthwhile.



> @shevek-google According to the [README](https://github.com/googleapis/python-storage#unsupported-python-versions) for the GCS client lib:
> 
> > Python == 3.6: the last released version which supported Python 3.6 was google-cloud-storage 2.0.0, released 2022-01-12.
> > Python == 3.5: the last released version which supported Python 3.5 was google-cloud-storage 1.32.0, released 2020-10-16.
> > Python == 2.7: the last released version which supported Python 2.7 was google-cloud-storage 1.44.0, released 2022-01-05.
> 
> So if we're willing and able to downgrade the GCS client lib then we can support older versions of Python.
> 
> @xiashang0624 I think you upgraded the GCS client lib? Do we need `2.3.0` or can we downgrade?

I only used the '2.3.0' version during development so I'm not sure. Actually I don't think we have tested the compatibility matrix of Python versions vs. versions of dependencies for this tool yet. And I think this should be a task before we can declare an older version as "supported".  For example, I think `dataclasses` is also a Python 3.7+ feature https://docs.python.org/3.7/library/dataclasses.html#

xiashang0624:
> Also, I believe the type annotation syntax has been supported since 3.0. The `typing` module has only been available in the stdlib since 3.5. But there is a [backport](https://pypi.org/project/typing/) if we want to support older 3.x versions. See [here](https://stackoverflow.com/a/31029363) for more info. So I think we can support this PR for 3.0+ but ideally I'd rather not go beyond 3.6.
> 
> I should have done [this](https://github.com/google/dwh-migration-tools/pull/39/files#diff-cfab7219da3bb36313aff4a78ea8a1e3219509ac34cfd2774021d92642ff6776L15-R63) in a separate PR. I just noticed that `pyproject.toml` was out of sync with @xiashang0624 's change to `README.md`. I can take it out and we can address it in a separate PR that includes a downgrade of the GCS client lib if @xiashang0624 deems a downgrade worthwhile.

I think what we want to do here is to test the tool using different Python versions. The GCS usage is pretty standard here so I think we can probably use an earlier version.

xiashang0624:
This PR looks good given tool currently supports Python 3.7+, which is documented in the README already. It appears that we want to support an earlier Python version.  I created a separate task to follow-up: http://b/237219199

xiashang0624 (approved):

genemerrill (approved):
worked for me

PR #40, opened 2022-06-27
Use a more conventional Python project structure.

ajwelch4:
The main purpose of this PR is to use `setup.py` to install the client and its concrete dependencies from a set of [abstract dependencies](https://pipenv.pypa.io/en/latest/advanced/#pipfile-vs-setup-py).

Using abstract dependencies in `setup.py` will allow us to more easily support multiple Python versions, library use cases and PyPI distribution someday if we so choose. PSO is already investigating how to string together this tool, DTS, DVT, etc. so the library use case may be useful.

Another benefit of using `setup.py` is we can let it generate and install an executable for us via `console_scripts` instead of hard coding one at `dwh-migration-tools/bin/dwh-migration-client`. The `setup.py` generated executable will be put in the correct virtualenv or non-virtualenv location so that it's on the shell `PATH` and can be called without having to remember where the `bin` directory is. It will also respect `pip`'s `-e` editable mode option. 

Finally, re: populating `requirements.txt`. The Python world has largely moved to lock files for concrete dependency management and Pipenv and Poetry would generate a lock file for us from the abstract dependencies. Since I don't think we will use either of those tools, I've added a `make requirements` hack to build `requirements.txt` and `requirements_dev.txt`.


shevek-google:
I am happy to take this on (relative) trust on the assumption that AJ's team is the main user/tester of this code. What does anyone else think? Aside from that, the test case would have to be a VM which ran up this code as a build/install test?

ajwelch4:
[tox](https://tox.wiki/en/latest/) or [nox](https://nox.thea.codes/en/stable/) can be used to install and test across Python versions.

ajwelch4:
I think we should take that work on as part of b/237219199.

mjquinn-google (approved):
Makes sense, not familiar with the particulars in the python world, but I recognize the pattern from the JS world.

xiashang0624 (approved):

PR #41, opened 2022-06-27
Use a proper logger.

ajwelch4:
This allows us to support log message timestamps, levels and eventually thread name.


xiashang0624 (approved):

genemerrill (approved):
wfm on AJ's branch

PR #43, opened 2022-06-28
Use Python f-strings.

ajwelch4:
If we have settled on Python 3.6+ or 3.7+ (b/237219199) then we can and should use f-strings.

### DIFF
--- a/client/dwh_migration_client/batch_sql_translator.py
+++ b/client/dwh_migration_client/batch_sql_translator.py
@@ -81,8 +81,8 @@ class BatchSqlTranslator:  # pylint: disable=too-many-instance-attributes
             self.preprocessor.preprocess(self._input_directory, local_input_dir)
 
         gcs_path = self.__generate_gcs_path()
-        gcs_input_path = join("gs://%s" % self.config.gcs_bucket, gcs_path, "input")
-        gcs_output_path = join("gs://%s" % self.config.gcs_bucket, gcs_path, "output")
+        gcs_input_path = join(f"gs://{self.config.gcs_bucket}", gcs_path, "input")
+        gcs_output_path = join(f"gs://{self.config.gcs_bucket}", gcs_path, "output")
         logging.info("Uploading inputs to gcs ...")
         gcs_util.upload_directory(
             local_input_dir, self.config.gcs_bucket, join(gcs_path, "input")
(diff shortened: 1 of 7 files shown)


mjquinn-google (approved):

shevek-google:
I wonder if the idea of b/237219199 was to support at least BACK to 3.6, not FORWARDS from 3.6 only. What effort does this create if we ever want to support 3.5? Is there any LTS release which still ships 3.5?

Hackerpilot (approved):

ajwelch4:
@shevek-google I will look into whether any distro LTS releases support 3.5. FWIW which may not be much, the [2020 JetBrains survey](https://www.jetbrains.com/lp/python-developers-survey-2020/) pegs Python <=2.7 at 6% and Python <= 3.5 at 2.8%.

xiashang0624 (approved):
LGTM. Another thing I learnt about Python.  Thanks!

shevek-google:
OK, I accept the JetBrains survey, with a slight side-order of them being very new-code/developer focused, not old-sysadmin focused. So if there's ALSO no LTS release still running 3.5 then I approve this. Thank you.

bdw-g (approved):
Go nuts :-)

ajwelch4:
> OK, I accept the JetBrains survey, with a slight side-order of them being very new-code/developer focused, not old-sysadmin focused. So if there's ALSO no LTS release still running 3.5 then I approve this. Thank you.

@shevek-google Agreed that the JetBrains survey likely significantly underrepresents sysadmins.

[google-cloud-bigquery-migration](https://github.com/googleapis/python-bigquery-migration/blob/main/setup.py#L46) requires Python >=3.6. So we would have to start there if we want to support <= 3.5.

Also, I analyzed [Repology](https://repology.org/project/python/versions) for Python:

- [Python <= 2.7](https://docs.google.com/spreadsheets/d/1ejl0QYH6YSbsQSQ38hRJk43B0G5mRUxDkN6_IyqEEDE/edit#gid=1096926535&fvid=1049884159)
- [Python 3.0 - 3.7](https://docs.google.com/spreadsheets/d/1ejl0QYH6YSbsQSQ38hRJk43B0G5mRUxDkN6_IyqEEDE/edit#gid=1096926535&fvid=1204682823)

I think Repology in combo with the JetBrains data suggests if we were to ever have a problem it would likely be with 2.7 before 3.5. It seems both are mainly available in repos for EOL releases or soon to be EOL releases that have 3.6+ as well.

shevek-google:
Thank you for the analysis.

PR #45, opened 2022-07-01
Hive: get fields from storage descriptor if available

paolomorandini:
This should help mitigate the exception: `Storage schema reading not supported`, it also implies one less RPC (per table).

To reproduce the error:
```sql
create table some_tbl (x int) row format serde 'org.apache.hadoop.hive.serde2.OpenCSVSerde';
```

### DIFF
--- a/dumper/lib-ext-hive-metastore/src/main/java/com/google/edwmigration/dumper/ext/hive/metastore/HiveMetastoreThriftClient_v2_3_6.java
+++ b/dumper/lib-ext-hive-metastore/src/main/java/com/google/edwmigration/dumper/ext/hive/metastore/HiveMetastoreThriftClient_v2_3_6.java
@@ -124,6 +124,8 @@ public class HiveMetastoreThriftClient_v2_3_6 extends HiveMetastoreThriftClient 
             @Nonnull
             @Override
             public List<? extends Field> getFields() throws Exception {
+                // If we already have a non-null Storage Descriptor let's get the columns from it, we do one less remote call and we also avoid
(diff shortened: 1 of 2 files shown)


bdw-g (approved):
Looks good, though I have no idea why this would work :-)

mjquinn-google (approved):
Makes sense to me.

dumper/lib-ext-hive-metastore/src/main/java/com/google/edwmigration/dumper/ext/hive/metastore/HiveMetastoreThriftClient_v2_3_6.java
```diff
@@ -123,31 +124,42 @@ public String getLocation() {
             @Nonnull
             @Override
             public List<? extends Field> getFields() throws Exception {
+                if (table.getSd() != null) {
+                    List<com.google.edwmigration.dumper.ext.hive.metastore.thrift.api.v2_3_6.FieldSchema> cols = table.getSd().getCols();
+                    if (cols != null) {
+                        return cols.stream().map(this::toField).collect(Collectors.toList());
+                    }
+                }
```
shevek-google, line 132: I'd love a comment saying why this condition is here, and why there are two paths.

shevek-google (approved):

dumper/lib-ext-hive-metastore/src/main/java/com/google/edwmigration/dumper/ext/hive/metastore/HiveMetastoreThriftClient_v2_3_6.java
paolomorandini: Done.

PR #50, opened 2022-07-06
Don't use name mangling.

ajwelch4:
Using a double underscore for "private" methods and attributes in Python is very rare. What the double underscore actually does is mangle the name of the method or attribute during bytecode compilation to include the class name as a prefix. This makes debugging and reflection more difficult.

Instead, the convention in Python is to use a single underscore for "private" methods and attributes. Both the single and double underscore don't actually make anything private. There is no way to truly make something private in Python. In this regard, you will hear some Python people say ["We are all consenting adults."](https://python-guide-chinese.readthedocs.io/zh_CN/latest/writing/style.html#we-are-all-consenting-adults). I guess this has been revised to [responsible users](https://github.com/realpython/python-guide/issues/525) in recent years.

[This SO post](https://stackoverflow.com/questions/7456807/python-name-mangling) has more details, most of which I agree with.


mjquinn-google (approved):
I consent.

Hackerpilot (approved):

PR #51, opened 2022-07-07
Use a proper marshaling library to deserialize and validate configs.

ajwelch4:
This PR uses [marshmallow](https://marshmallow.readthedocs.io/en/stable/index.html), a popular standalone data marshaling  library, to deserialize and validate `config.yaml`, `macros.yaml` and `object_mapping.json`. Benefits include:
- More thorough validation.
- Replaces tedious conditional logic and asserts with a declarative approach.
- Replaces asserts, which can be compiled away, with exceptions.
- Catches  `ValidationError` exceptions and logs an improved error message.


shevek-google (approved):
I'm in favour of it; this is tricky enough that I have proof-of-correctness questions. Who else has eyes on this?

client/dwh_migration_client/config.py
```diff
@@ -0,0 +1,114 @@
...
+# limitations under the License.
+"""Parses the config file into a Config object."""
+
+import logging
+from dataclasses import asdict, dataclass
+from pprint import pformat
+from typing import List, Optional
+
+import yaml
+from marshmallow import Schema, ValidationError, fields, post_load
```
xiashang0624, line 22: Not surprisingly, this is also my first time learning what magic this package offers. And the change LGTM after ~10 mins of Google search on it.

A general/naive question,  what is the rule of thumb when we should consider to add a new package to replace some existing logics that are implemented in the old fashioned way? A new open-source package offers some level of improvements (otherwise it shouldn't exist in the first place) but it also raises the bars of maintenance/readability because we are asking readers to spend another ~10min learning this package.  And I think it also raises the risk of compatibility issues due to a new dependency added to this tool.

I am not saying we shouldn't add this change, just curious about the "best practice" in the open source Python  world.

client/setup.py
```diff
@@ -23,8 +23,9 @@
         "protobuf>=3.20.1",
         # gRPC,
         "grpcio>=1.46.1",
-        # PyYML package to parse yaml config files.
+        # Used to marshal config file data.
         "PyYAML>=6.0",
+        "marshmallow<=3.14.1",
```
xiashang0624, line 28: why the version of marshmallow needs to be less than a specific version?

client/setup.py
```diff
@@ -23,8 +23,9 @@
         "protobuf>=3.20.1",
         # gRPC,
         "grpcio>=1.46.1",
-        # PyYML package to parse yaml config files.
+        # Used to marshal config file data.
```
xiashang0624, line 26: move to the next line?

client/dwh_migration_client/object_name_mapping.py
```diff
@@ -0,0 +1,111 @@
...
+        data = json.load(file)
+    try:
+        object_name_mapping_list: ObjectNameMappingList = (
+            ObjectNameMappingListSchema().load(data)
+        )
+    except ValidationError as error:
+        logging.error(
+            "Invalid object name mapping file: %s: %s.",
+            object_name_mapping_file_path,
+            error,
```
xiashang0624, line 103: I am curious what kind of error messages we will get from marshmallow.  In the old logic, we basically just prepared the error message indicating how to fix the error in the json file.

client/setup.py
ajwelch4: Can't upgrade beyond `3.14.1` until wee drop support for Python `3.6`.

client/setup.py
ajwelch4: I put it above `PyYML` because `PyYML` and `marshmallow` are both doing the work. I should probably change it to something like "Used to read and validate config file data".

client/dwh_migration_client/object_name_mapping.py
ajwelch4: Something along the lines of:

Required field `location` not specified:
```
2022-07-07 03:25:56,198: ERROR: Invalid config file:/usr/local/google/home/ajwelch/Documents/code/integrations/dwh-migration/dwh-migration-tools/client/config.yaml: {'translation_config': {'location': ['Missing data for required field.']}}.
```

Invalid fields in the config file (I accidentally pasted an old config form before we moved input and output to cli args):
```
2022-07-07 05:08:08,778: ERROR: Invalid config file: /usr/local/google/home/ajwelch/Documents/code/integrations/dwh-migration/dwh-migration-tools/client/config.yaml: {'translation_config': {'input_directory': ['Unknown field.'], 'output_directory': ['Unknown field.']}}.
```

But we can make these whatever we want.

xiashang0624 (approved):

PR #49, opened 2022-07-04
Hive: test dumper with an in-memory Hive metastore

paolomorandini:
Command:
```
./gradlew :dumper:app:{cleanTest,test} --tests HiveMetadataConnectorTest -Dtest.verbose=true -Dorg.gradle.java.home=/usr/lib/jvm/java-1.8.0-openjdk-amd64
```

PR #53, opened 2022-07-08
Update the url to the specific job panel

xiashang0624:
This will update the url to the job detail panel directly


paolomorandini:
#49 will also fix the out of memory issues

shevek-google (approved):

paolomorandini (approved):

PR #56, opened 2022-07-08
Add language to functions.csv for Redshift

dawidxc:


shevek-google (approved):

shevek-google (changes requested):
Needs adding to the enum in the format description.

Add a field to 

https://github.com/google/dwh-migration-tools/blob/main/dumper/lib-dumper-spi/src/main/java/com/google/edwmigration/dumper/plugin/lib/dumper/spi/RedshiftMetadataDumpFormat.java#L174

shevek-google (approved):
I'll approve this because while I suspect the case mismatch in the header enum is still a bug, it's not a new bug, so you're not responsible for it.

bdw-g (approved):

PR #57, opened 2022-07-08
Create alpha branch using manually generated v2alpha client library

gabbybarbieri:


client/dwh_migration_client/batch_sql_translator.py
```diff
@@ -24,7 +24,7 @@
 from os.path import dirname, join
 from typing import Optional
 
-from google.cloud import bigquery_migration_v2
+from google.cloud import bigquery_migration_v2alpha
```
shevek-google, line 27: from google.cloud import bigquery_migration_v2_alpha as bigquery_migration_v2

and don't edit any of the code below.

shevek-google (changes requested):
Is it possible to import it as a source tree rather than as a tar.gz? I think that plays better with github, and lets us see what changed.

client/dwh_migration_client/batch_sql_translator.py
gabbybarbieri: Good idea!

gabbybarbieri:
> Is it possible to import it as a source tree rather than as a tar.gz? I think that plays better with github, and lets us see what changed.

Yes it is; I created https://github.com/google/dwh-migration-tools/pull/58 as a sample. It definitely does need a level of refinement from someone who has more python-knowledge than I; especially in regards to namespaces/modules/etc. as this has the source under `client/google/cloud/` rather than something like `<alpha_dir_name>/google/cloud/` or `client/<alpha_dir_name>/google/cloud`.

yzhvictor:
Just to confirm, since we have https://github.com/google/dwh-migration-tools/pull/58, should this one be closed?

PR #59, opened 2022-07-15
Delete a type that is not supported in Python 3.7

xiashang0624:


xiashang0624:
Verified using Python 3.7

yzhvictor (approved):

client/dwh_migration_client/macro_processor.py
```diff
@@ -260,7 +260,7 @@ def _parse_macros_config_file(self) -> Dict[str, Dict[str, str]]:
 
     def _get_all_regex_pattern_mapping(
         self, file_path: str, use_reversed_map: bool = False
-    ) -> Tuple[Dict[str, str], Pattern[str]]:
+    ):
```
ajwelch4, line 263: We should probably change this to something if possible. Which part of the type signature didn't check?

ajwelch4 (changes requested):
We should probably add the minimum Python version we plan on supporting to the `mypy` section of the `pyproject.toml` file so it catches this: https://mypy.readthedocs.io/en/stable/config_file.html#platform-configuration.

client/dwh_migration_client/macro_processor.py
xiashang0624: I don't know which part of this type is not supported in Python 3.7. Actually I am pretty confused about this. So Python doesn't technically enforce type but it's recommended to have type annotations.  And clearly there are unexpected behaviors at different Python 3 versions.

Since we don't have automatic integration tests yet, for future PRs, I think we should at least manually verify the e2e logic at both Python 3.7 and Python 3.10 versions.  WDYT?

BTW, this issue is raised by a big customer, Ford.  I think we may want to submit this quick fix, especially given this is a private method.

xiashang0624:
> We should probably add the minimum Python version we plan on supporting to the `mypy` section of the `pyproject.toml` file so it catches this: https://mypy.readthedocs.io/en/stable/config_file.html#platform-configuration.

Done.

client/dwh_migration_client/macro_processor.py
ajwelch4: Unless the customer is running `mypy` after a code change (which I doubt but possible), my guess is it failed to import one of the types. Do we have the error message? If not, I should be able to test today and figure it out.

ajwelch4: Just looked at the stacktrace. This is the issue:
```
from re import Pattern
```
Should be:
```
from typing import Pattern
```
`re.Pattern` did not support subscripting until 3.9: https://docs.python.org/3/library/typing.html#typing.Pattern.

ajwelch4: You should be able to add the type signature back after that.

xiashang0624: If we replace "re" with "typing" in the import, have you verified that both 3.7 and 3.10 still work e2e with the --macros flag enabled?

I am currently working on another high priority issue, I can verify this next week as well.

ajwelch4: I will create a branch with with a quick and dirty integration test and use nox to test it on all supported versions of Python.

ajwelch4: https://github.com/google/dwh-migration-tools/pull/62

shevek-google: Is this conversation resolved, and is this PR now mergeable?

ajwelch4: No. This is the real fix IMO: https://github.com/google/dwh-migration-tools/pull/62

ajwelch4: I would close this and merge #62

ajwelch4:
Closing in favor of #62

PR #60, opened 2022-07-15
not all files may contain all macros.

stankiewicz:
As user I produce macros globally to all SQL files - global macros.
There may be a chance that one or few files don't have any parameter and those files should not fail processing.


ajwelch4 (approved):
Thanks for catching this! I would have done `if not reg_pattern_map:` but other than that LGTM.

PR #61, opened 2022-07-15
Guard SVV_DISKUSAGE behind --metadata-stats

dawidxc:


bdw-g (approved):

shevek-google:
In the medium term, I'd love to unguard this if we can prove that it doesn't have significant cost, the way the table stats do.

PR #64, opened 2022-07-21
Teradata dump more tables

okerbez:


shevek-google (commented):
Are you planning always to run with --metadata-stats ? If so, should these be guarded within the following if() block?

The ones that prove to be valuable and cheap in the long run, I'd like to pull out, but I start cautious. What's your opinion?

okerbez:
> Are you planning always to run with --metadata-stats ? If so, should these be guarded within the following if() block?
> 
> The ones that prove to be valuable and cheap in the long run, I'd like to pull out, but I start cautious. What's your opinion?

Sounds fair, I moved new tables inside if, and will consider your comment in future changes

okerbez:
> Are you planning always to run with --metadata-stats ? If so, should these be guarded within the following if() block?
> 
> The ones that prove to be valuable and cheap in the long run, I'd like to pull out, but I start cautious. What's your opinion?

Sounds fair, I moved new tables inside if, and will consider your comment in future changes

shevek-google (approved):
Thank you.

PR #85, opened 2022-08-26
Change `--assessment` and `--metadata-stats` flags spec to Void

frsv:
Currently following examples give results which go against the documentation:
* `--assessment` is parsed as false
* `--assessment any_possible_value` is parsed as false instead of failing

As per discussion, set it to OptionSet<Void>, making any arguments after the flag noop because of joptsimple bug.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
@@ -132,7 +132,7 @@ public class ConnectorArguments extends DefaultArguments {
     private final OptionSpec<String> optionWarehouse = parser.accepts(OPT_WAREHOUSE, "Virtual warehouse to use once connected (for providers such as Snowflake)").withRequiredArg().ofType(String.class);
     private final OptionSpec<String> optionDatabase = parser.accepts(OPT_DATABASE, "Database(s) to export").withRequiredArg().ofType(String.class).withValuesSeparatedBy(',').describedAs("db0,db1,...");
     private final OptionSpec<String> optionSchema = parser.accepts(OPT_SCHEMA, "Schemata to export").withRequiredArg().ofType(String.class).withValuesSeparatedBy(',').describedAs("sch0,sch1,...");
-    private final OptionSpec<Boolean> optionAssessment = parser.accepts(OPT_ASSESSMENT, "Whether to create a dump for assessment (i.e., dump additional information).").withOptionalArg().withValuesConvertedBy(BooleanValueConverter.INSTANCE).defaultsTo(Boolean.FALSE);
+    private final OptionSpec<Void> optionAssessment = parser.accepts(OPT_ASSESSMENT, "Whether to create a dump for assessment (i.e., dump additional information).");
     private final OptionSpec<String> optionUser = parser.accepts(OPT_USER, "Database username").withRequiredArg().describedAs("admin");
     private final OptionSpec<String> optionPass = parser.accepts(OPT_PASSWORD, "Database password, prompted if not provided").withOptionalArg().describedAs("sekr1t");
     private final OptionSpec<String> optionRole = parser.accepts(OPT_ROLE, "Database role").withRequiredArg().describedAs("dumper");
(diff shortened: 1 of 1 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -416,7 +416,7 @@ public List<String> getSchemata() {
     }
 
     public boolean isAssessment() {
-        return BooleanUtils.isTrue(getOptions().valueOf(optionAssessment));
+        return getOptions().has(optionAssessment);
```
dawidxc, line 419: I think `has` just checks, whether the option is there, not whether the boolean has value `true`.
https://github.com/jopt-simple/jopt-simple/blob/master/net.sf.joptsimple/src/main/java/joptsimple/OptionSet.java#L94

I would suggest:
```
return getOptions().valueOf(optionAssessment);
```

frsv: Apparently, none of this is enough. Argument parser is interesting, allowing multiple values combination and evaluating options lazely.

I tried to go with:
```
public boolean isAssessment() {
        OptionSet options = getOptions();
        if (options.has(optionAssessment)) {
            return options.hasArgument(optionAssessment) ? options.valueOf(optionAssessment) : true;
        }
        return false;
    }
```

But it also doesn't work as expected, correctly evaluating ` `, `--assessment`, `--assessment [no,false,0]`, but accepting everything else as true, without failing.

I decided to make the flag to expect required argument. It doesn't allows to use `--assessment` as is, but at least it fails for something like `--assessment folse`.

bdw-g:
There's no way to have it as a 'default boolean flag' , like you have in getopt?

bdw-g:
actually, I think you can do this with an `Option<Void>`

shevek-google (changes requested):
As previous commenter; I would do this as an Option<Void> and use has() instead of valueOf()

frsv:
If it is just `getOptions().has(assessmentOption)`, then `--assessment no` will be treated as true.
`OptionSpec<Void>` also allows to pass any argument after the flag.
I would prefer it to be more defensive. I debugged the code to understand why it doesn't fail on unexpected arguments and treats it as true.

So my investigation:
1. [BooleanValueConverter](https://github.com/google/dwh-migration-tools/blob/f5e2c42f7e50f7db26d5e41f4d9abd4e73004339/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/DefaultArguments.java#L69) raises a `ValueConversionException` when unexpected argument arrives.
2. Exception is rethrown as `OptionException ` by [AbstractOptionSpec](https://github.com/jopt-simple/jopt-simple/blob/master/net.sf.joptsimple/src/main/java/joptsimple/AbstractOptionSpec.java#L104)
3. [ArgumentAcceptingOptionSpec](https://github.com/jopt-simple/jopt-simple/blob/master/net.sf.joptsimple/src/main/java/joptsimple/ArgumentAcceptingOptionSpec.java#L289) ignores it and as flag has optional argument, proceeds further.

Only `ReflectionException` and `ValueConversionException` are caught at step 1. Changing the exception in converter to any other helps to fail fast.

shevek-google:
If you remove the withOptionalArg() from assessmentOption then it doesn't take an argument at all, and its presence alone can be tested with has().

frsv:
It will work, I considered it. There are two reasons why I want to keep it like this:
* In the documentation we refer to `--assessment yes` and `--metadata-stats yes`. I don't think adjust the documentation is hard, but imho it's faster to fix the code
* Not allowlisted boolean values are silently accepted, which looks like a bug

If you don't like these changes, I'll switch to keeping it as a flag, not accepting values.

shevek-google:
I think, given our long view, the small number of customers reliant on the current behaviour, and that joptsimple hasn't addressed this issue yet, we should switch to a flag now.

frsv:
Makes sense. I've changed OptionSpec for this two flags to <Void> type, making any flags after them noop

shevek-google (approved):

PR #86, opened 2022-08-26
Make macro-unexpand case-insensitive

xiashang0624:
During translation:

* all the schema names in the inputs are converted to the lower-case.
* all SQL commands (e.g., select, from, join, etc) are converted to the upper-case.

This PR makes the macro unexpansion case-insensitive.

### DIFF
--- a/client/dwh_migration_client/macro_processor.py
+++ b/client/dwh_migration_client/macro_processor.py
@@ -23,7 +23,7 @@ import shutil
 from argparse import Namespace
 from os.path import abspath, dirname, isfile, join
 from pprint import pformat
-from typing import Dict, Pattern, Tuple
+from typing import AnyStr, Dict, Pattern, Tuple
 
 import yaml
 from marshmallow import Schema, ValidationError, fields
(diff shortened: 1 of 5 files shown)


shevek-google (approved):

client/dwh_migration_client/macro_processor.py
```diff
@@ -273,6 +275,8 @@ def _get_all_regex_pattern_mapping(
         for file_map_key, token_map in macro_subst_maps.items():
             if fnmatch.fnmatch(file_path, file_map_key):
                 for key, value in token_map.items():
-                    reg_pattern_map[re.escape(key)] = value
+                    reg_pattern: AnyStr = "{0}{1}".format(self._case_insensitive_re_pattern, re.escape(key.lower())) \
+                        if use_reversed_map else re.escape(key)
+                    reg_pattern_map[reg_pattern] = value
         all_patterns = re.compile("|".join(reg_pattern_map.keys()))
```
ajwelch4, line 281: I get the following when I run `make test`:
```
dwh_migration_client/macro_processor.py:281: DeprecationWarning: Flags not at the start of the expression '(?i)my_table|(?i)mac' (truncated)
```
Is it possible to concat `(?i)` to the beginning of the compound pattern passed to `re.compile` instead each individual pattern?

ajwelch4 (changes requested):

client/dwh_migration_client/macro_processor.py
xiashang0624: Yep, done.  Thanks.

ajwelch4 (approved):

PR #92, opened 2022-09-08
Add columns for Teradata logs dumping with '--assessment' flag

abogdanov93:


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsConnector.java
```diff
@@ -273,9 +405,11 @@ public void addTasksTo(List<? super Task<?>> out, @Nonnull ConnectorArguments ar
         ZonedIntervalIterable intervals = ZonedIntervalIterable.forConnectorArguments(arguments);
         LOG.info("Exporting query log for " + intervals);
         SharedState state = new SharedState();
+        boolean isAssessment = arguments.isAssessment();
         for (ZonedInterval interval : intervals) {
             String file = ZIP_ENTRY_PREFIX + DateTimeFormatter.ISO_OFFSET_DATE_TIME.format(interval.getStartUTC()) + ".csv";
-            out.add(new TeradataLogsJdbcTask(file, state, logTable, queryTable, condition, interval).withHeaderClass(Header.class));
+            out.add(new TeradataLogsJdbcTask(file, state, logTable, queryTable, condition, interval,
+                    isAssessment).withHeaderClass(isAssessment ? HeaderForAssessment.class : Header.class));
```
bdw-g, line 412: I think this should probably better be a subclass (`TeradataAssessmentLogsJdbcTask`) rather than fiddling with booleans and changing arrays.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/Teradata14LogsConnector.java
```diff
@@ -111,12 +111,12 @@ String getSql(@Nonnull Predicate<? super String> predicate) {
...
 
         public LogQueryFactory(String targetPath, SharedState state, String logTable, String queryTable, String condition, ZonedInterval interval) {
-            super(targetPath, state, logTable, queryTable, condition, interval);
+            super(targetPath, state, logTable, queryTable, condition, interval, false);
         }
 
         @Override
         @Nonnull
-        String getSql(@Nonnull Predicate<? super String> predicate) {
+        String getSql(@Nonnull Predicate<? super String> predicate, String[] expressions) {
```
bdw-g, line 119: I think you only have to add this, because you added the parameters to the TeradataLogsConnector.
If you instead created a `TeradataAssessmentLogsJdbcTask`  - ie. add this customization as a subclass - then yo don't have to add these (unused) parameters here.

bdw-g (commented):
Minor comments; I think that if you're going to create a substantially different query, you should probably be using a different task subclass.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/Teradata14LogsConnector.java
abogdanov93: Agree, this parameter should not present in this class.

I removed `expressions` from this method, here is no longer needed. It is overridden only in `TeradataLogsConnector` class now, could you please take another look?

bdw-g (approved):
Much nicer, thank you!

paolomorandini (approved):

PR #96, opened 2022-09-16
move STL_WLM_QUERY to raw logs connector (Redshift assessment dump)

dominik-klimczak:
STL_WLM_QUERY can be big and has a timestamp, so raw logs connector is a correct place to dump it.

### DIFF
--- a/dumper/lib-dumper-spi/src/main/java/com/google/edwmigration/dumper/plugin/lib/dumper/spi/RedshiftRawLogsDumpFormat.java
+++ b/dumper/lib-dumper-spi/src/main/java/com/google/edwmigration/dumper/plugin/lib/dumper/spi/RedshiftRawLogsDumpFormat.java
@@ -80,4 +80,14 @@ public interface RedshiftRawLogsDumpFormat {
             commit_queue_elapsed, commit_exec_time;
         }
     }
+
+    interface WlmQuery {
+        String ZIP_ENTRY_PREFIX = "wlm_query_";
+
+        enum Header {
+            userid, xid, task, query, service_class, slot_count, service_class_start_time,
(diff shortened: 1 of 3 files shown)


dumper/lib-dumper-spi/src/main/java/com/google/edwmigration/dumper/plugin/lib/dumper/spi/RedshiftRawLogsDumpFormat.java
```diff
@@ -80,4 +80,14 @@ enum Header {
...
         }
     }
+
+    interface WlmQuery {
+        String ZIP_ENTRY_PREFIX = "wlm_query_";
+
+        enum Header {
+            userid, xid, task, query, service_class, slot_count, service_class_start_time,
+            queue_start_time, queue_end_time, total_queue_time, exec_start_time, exec_end_time,
+            total_exec_time, service_class_end_time, final_state, query_priority;
```
paolomorandini, line 90: Is there an opportunity to generate the above SQL projection by `Joiner.on(", ").join(Header.values())` or similar?

paolomorandini (approved):
Looks good

dumper/lib-dumper-spi/src/main/java/com/google/edwmigration/dumper/plugin/lib/dumper/spi/RedshiftRawLogsDumpFormat.java
dominik-klimczak: Good idea, but I should refactor also other tables as well, so I do it in another PR.

PR #104, opened 2022-09-27
Change processing --assessment flag for Teradata query logs

abogdanov93:
1. Change base log table to `dbc.QryLogV` in order to get field `SpoolUsage`
2. Save byte array as base64 format instead of `toString()`

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsConnector.java
@@ -23,20 +23,23 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Predicates;
 import com.google.common.io.ByteSink;
 import com.google.edwmigration.dumper.application.dumper.annotations.RespectsArgumentAssessment;
+import com.google.edwmigration.dumper.plugin.ext.jdk.progress.RecordProgressMonitor;
 import com.google.errorprone.annotations.ForOverride;
+import java.io.IOException;
+import java.io.Writer;
+import java.nio.charset.StandardCharsets;
 import java.sql.Connection;
 import java.sql.SQLException;
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
+import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import javax.annotation.Nonnull;
+import org.apache.commons.csv.CSVFormat;
+import org.apache.commons.csv.CSVPrinter;
 import org.apache.commons.lang3.StringUtils;
 import com.google.edwmigration.dumper.application.dumper.ConnectorArguments;
 import com.google.edwmigration.dumper.application.dumper.MetadataDumperUsageException;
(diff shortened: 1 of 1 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsConnector.java
```diff
@@ -61,6 +64,8 @@
 import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.jdbc.core.ResultSetExtractor;
 
+import static com.google.edwmigration.dumper.application.dumper.connector.teradata.TeradataLogsConnector.TeradataAssessmentLogsJdbcTask.ASSESSMENT_DEF_LOG_TABLE;
```
bdw-g, line 67: You have a static import for a variable that you define in this file?

bdw-g (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsConnector.java
abogdanov93: Yes, I think it works this way, because this constant is defined in inner static class (that is basically the same as normal java class), that's why if you use it outside of this class you have to import it.

I agree that it looks weird, but instead I can either use it with class prefix (`TeradataAssessmentLogsJdbcTask.ASSESSMENT_DEF_LOG_TABLE`) or define in upper class `TeradataLogsConnector` (same as `DEF_LOG_TABLE` is defined).

So I've decided to move it inside `TeradataAssessmentLogsJdbcTask`, in order usage looks better and imports are usually hidden.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsConnector.java
```diff
@@ -393,6 +398,39 @@ public TeradataAssessmentLogsJdbcTask(@Nonnull String targetPath, SharedState st
...
+                        : new RecordProgressMonitor(getName());
+                     Writer writer = sink.asCharSink(StandardCharsets.UTF_8).openBufferedStream();
+                     CSVPrinter printer = format.print(writer)) {
+                    final int columnCount = rs.getMetaData().getColumnCount();
+                    while (rs.next()) {
+                        monitor.count();
+                        for (int i = 1; i <= columnCount; i++) {
+                            Object object = rs.getObject(i);
+                            if (object instanceof byte[]) {
+                                printer.print(Base64.getEncoder().encodeToString((byte[]) object));
```
shevek-google, line 419: I think I'd like to see this, and potentially the Clob path from Oracle, both pushed into the superclass. I think this is a reasonable good behaviour for the common code. What do other reviewers think?

shevek-google (changes requested):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsConnector.java
abogdanov93: I'd suggest to do it in next PR

shevek-google: If we know that a 'next' PR is going to happen on some sort of deterministic timeline, that's OK, but if this pushes the refactor off into the indefinite future, then I'd push us to do this now before the methods have a chance to diverge further.

abogdanov93: ok, it is done in this PR, please see a9a7a4be060a487ee2d299a0478dc569e18c1436

shevek-google (approved):
Note we typically don't do things like the inner-class to lambda change because it obscures the _actual_ semantic diff with formatting changes. Formatting changes are encouraged in formatting-only PRs, but not in semantic PRs.

PR #111, opened 2022-10-10
Pick QueryID from DBQLogTbl/QryLogV instead DBQLSQLTbl

frsv:
Left outer join is based on L alias, but QueryID is taken from ST, making logs, missed in "dbc.DBQLSQLTbl", indistinguishable from each other.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsConnector.java
@@ -95,7 +95,7 @@ public class TeradataLogsConnector extends AbstractTeradataConnector implements 
     // MUST match TeradataLogsDumpFormat.Header
     @VisibleForTesting
     static final String[] EXPRESSIONS = new String[]{
-        "ST.QueryID",
+        "L.QueryID",
         "ST.SQLRowNo",
         "ST.SQLTextInfo",
         "L.UserName",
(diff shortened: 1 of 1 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsConnector.java
```diff
@@ -116,7 +114,9 @@ public class TeradataLogsConnector extends AbstractTeradataConnector implements
         "L.SessionWDID",
         "L.Statements",
         "L.TotalIOCount",
-        "L.WarningOnly"
+        "L.WarningOnly",
+        "ST.SQLRowNo",
+        "ST.SQLTextInfo"
```
shevek-google, line 119: Why the order change? It seems like this should have been a 1 line change, not a 7 line change?

shevek-google (changes requested):
Can this change be done smaller, without changing orders, or is the reorder significant?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsConnector.java
frsv: Done. Reorder is not significant and, apparently, requires enum updates which I definitely didn't intend.

shevek-google (approved):
I'm still curious to know what jobs the OUTER misses on, but this is a good change now.

frsv:
@shevek-google short queries might not be present in DBQLSQLTbl, in this case their QueryText column in query logs table already contains full query text

PR #113, opened 2022-10-13
Extract the aborted column for RedShift queries

haselbach:
When the `--assessment` flag is set, extract the `aborted` column from `STL_QUERIES`.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftRawLogsConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftRawLogsConnector.java
@@ -17,6 +17,8 @@
 package com.google.edwmigration.dumper.application.dumper.connector.redshift;
 
 import com.google.auto.service.AutoService;
+import com.google.common.base.Joiner;
+import com.google.common.collect.ImmutableList;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftRawLogsConnector.java
```diff
@@ -97,14 +99,21 @@ public void addTasksTo(List<? super Task<?>> out, ConnectorArguments arguments)
         // 1. STL_QUERY has starttime, queryid, but text is 4000 char wich is useless
         // 2. STL_QUERY_TEXT has xid+squence+text which reconstructs query, but no starttime.
         // STL_QUERY is 1 row per query ; SQL_QUERY_TEXT is multi rows per query, using sequence and xid
-        String queryTemplateQuery
-                = "SELECT userid, xid, pid, query, trim(label) as label, starttime, endtime, sequence, text"
-                + " FROM STL_QUERY join STL_QUERYTEXT using (userid, xid, pid, query) WHERE ##";
+        List<String> queryTemplateColumns = new ArrayList<>(ImmutableList.<String>of(
```
bdw-g, line 102: `Lists.newArrayList(E .... elements)` is somewhat neater

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftRawLogsConnector.java
```diff
@@ -97,14 +99,21 @@ public void addTasksTo(List<? super Task<?>> out, ConnectorArguments arguments)
...
         if (arguments.isAssessment()) {
-            queryTemplateQuery += " ORDER BY starttime, query, sequence";
+            queryTemplateColumns.add("aborted");
+            queryTemplateOrderBy.addAll(ImmutableList.of("starttime", "query", "sequence"));
         }
 
+        String queryTemplateQuery = String.format(
+            "SELECT %s FROM STL_QUERY join STL_QUERYTEXT using (userid, xid, pid, query) WHERE ##%s",
+            Joiner.on(", ").join(queryTemplateColumns),
+            queryTemplateOrderBy.isEmpty() ? "" : (" ORDER BY " + Joiner.on(", ").join(queryTemplateOrderBy)));
```
bdw-g, line 115: Truth be told, why not just have two static strings, much simpler that way, much easier to debug, too.

bdw-g (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftRawLogsConnector.java
haselbach: Good point. I thought `newArrayList` was deprecated but that was just for the variant w/o arguments.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftRawLogsConnector.java
haselbach: I disagree because it would then require to keep them in sync and it would be easier for the parts that should be the same to be only changed in one of them. I'm not a big fan of constructing SQL statements out of strings. IMHO we should have a SQL builder, but that would be a bigger change. So, I'm just mimicking the important parts of that here.

I'd like to rather have something like this.

```java
SqlBuilder.select("STL_QUERY")
  .columns("userid", "xid", "pid")
  .columnAs("trim(label)", "trim")
  .orderBy("starttime")
```

bdw-g: True enough, but we don't have that gadget, nor any interest in building it and maintaining it :-)

bdw-g (approved):

frsv (approved):

PR #121, opened 2022-11-14
Remove partitions limit from Hive Thrift metadata connector

frsv:
Source code [indicates](http://go/gh/apache/hive/blob/master/standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift#L2655) that it's possible to return all the partitions for the table. 

With this I was able to get all 41K partitions for one of my tables. Tested with other smaller ones too.


bdw-g (approved):
neat

shevek-google (approved):

PR #122, opened 2022-11-14
Add owner info and create time to Hive functions task

frsv:
I decided to not put it under `--assessment` flag since this doesn't make much difference anyway – I assume we won't expect many UDFs


bdw-g (approved):

shevek-google:
Anything which doesn't have performance (or other) risk doesn't need to go under --assessment so this is a good call

PR #123, opened 2022-11-18
Test to confirm passing multiple JARs comma-separated to `--driver` argument allows loading classes from all of them.

paolomorandini:
This is needed for some JDBC drivers which provide optional features through additional JARs.


bdw-g (approved):
This looks okay but please fix the spotless error

shevek-google (approved):
This always worked because teradata required it, but it's nice to have a test.

PR #126, opened 2022-12-08
Bteq: support extraction of heredoc SQL fragments from KSH inputs

mjquinn-google:
See b/236838133

### DIFF
--- a/client/bqms_run/hooks.py
+++ b/client/bqms_run/hooks.py
@@ -16,10 +16,7 @@ import logging
 import re
 from typing import Mapping
 
-from bqms_run.ksh import (
-    KshExtractor,
-    ShellFragmentType
-)
+from bqms_run.ksh import KshExtractor
 from bqms_run.macros import (
     MacroExpanderRouter,
     PatternMacroExpander,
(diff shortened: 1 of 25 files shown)


client/examples/teradata/bteq/clean.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash -ex
```
mjquinn-google, line 1: Ideally we would symlink these and the next few shell scripts to a common directory shared among all examples, but the README advocates that examples be copied out of the repo and into a folder elsewhere on the user's filesystem, which would break the symlinks. So I've simply copied them verbatim from the existing teradata/sql example.

client/bqms_run/hooks.py
```diff
@@ -98,3 +103,25 @@ def un_expand(
...
+                 preprocessed.
+    :return: for non-KSH inputs, the input text is returned unmodified. For
+                 KSH inputs, all heredoc SQL fragments in the input are
+                 concatenated together and returned. If an input contains
+                 no heredoc SQL fragments, a comment saying as much is returned.
+    """
+    if not path.match("*.ksh"):
+        return text
+    logger.info("Extracting heredoc fragments from KSH input: %s", path)
+    fragments = KshExtractor("bteq").read_fragments(text)
```
mjquinn-google, line 123: In CompilerWorks, the command to use ("bteq" here) was configurable via Groovy DSL. I've left it as unconfigurable for now, unless and until we get a specific ask to do so, in which case we can add a config YAML option for it.

client/bqms_run/ksh.py
```diff
@@ -0,0 +1,228 @@
...
+    def get_text(self) -> str:
+        return self.text
+
+
+class ExtractorState(Enum):
+    SHELL = "SHELL"
+    BTEQ = "BTEQ"
+
+
+class KshExtractor:
```
bdw-g, line 74: It seems this is a direct port of the java version, correct? Comments and warnings and all :-)

mjquinn-google: Yes

client/bqms_run/hooks.py
mjquinn-google: (fyi)

client/examples/teradata/bteq/clean.sh
mjquinn-google: (fyi)

PR #130, opened 2023-01-09
Replace CollectTimeStamp with StartTime for Teradata query tables

frsv:
`CollectTimeStamp` doesn't relate to query – from the documentation it is "A date and time unique to each buffer cache, which changes for each new buffer", also see [community answer](https://support.teradata.com/community?id=community_question&sys_id=647632d01b9010d046fcb99e0a4bcb6b).

On the Assessment side we saw 2 year difference between `CollectTimeStamp` and `StartTime` columns. This PR changes SQL WHERE clauses for query tables to use `StartTime` instead.

It doesn't touch SQL log table on purpose – `CollectTimeStamp` there represents a time when row is written to DB ([link](https://docs.teradata.com/r/Teradata-Database-Administration/April-2018/Tracking-Query-Behavior-with-Database-Query-Logging-Operational-DBAs/SQL-Log-Table-DBQLSQLTbl).


bdw-g (approved):
Today we learn. Thanks!

mjquinn-google (approved):

PR #134, opened 2023-03-02
add timestamp to zip archive name

ishmum123:


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/Connector.java
```diff
@@ -36,10 +38,19 @@ public enum DefaultProperties implements ConnectorProperty {
     public String getName();
 
     @Nonnull
-    public String getDefaultFileName();
+    default String getDefaultFileName() {
```
frsv, line 41: I wonder if this change will break existing CW, translation and Assessment loaders.

E.g. for metadata connectors:

* Batch translation looks for just `.zip` file to locate metadata
* Assessment expects the file to end with a specific prefix, i.e. `redshift-metadata.zip`
* I'm not in the context of CW loaders

I'm not sure about query logs connectors (I believe there are only Assessment and CW loaders there).

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/Connector.java
```diff
@@ -36,10 +38,19 @@ public enum DefaultProperties implements ConnectorProperty {
...
+        SimpleDateFormat format = new SimpleDateFormat("yyyyMMdd'T'HHmmss");
+        Timestamp timestamp = new Timestamp(System.currentTimeMillis());
+        return "dwh-migration-" + getName() + "-" + type() + "-" + format.format(timestamp)
```
dawidxc, line 44: The `format` method accepts also the milliseconds, so the wrapping with `Timestamp` is not required.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/Connector.java
```diff
@@ -36,10 +38,19 @@ public enum DefaultProperties implements ConnectorProperty {
     public String getName();
 
     @Nonnull
-    public String getDefaultFileName();
+    default String getDefaultFileName() {
+        SimpleDateFormat format = new SimpleDateFormat("yyyyMMdd'T'HHmmss");
```
dawidxc, line 42: nit: I would recommend to change the type of the variable to `DateFormat`.

Reason: Item 64: Refer to objects by their interfaces, Effective Java (the `DateFormat` is an abstract class, not an interface, but I would use the same rationale here)

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/Connector.java
dawidxc: The change of the filenames needs to be applied only when the `--assessment` flag is used in order not to break other tools.

ishmum123: Added the `--assessment` flag check. Keeping in draft mode until the loaders are ready for this

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/Connector.java
ishmum123: Used the `Format` interface directly 😄

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/Connector.java
ishmum123: Ah thanks! Didn't know this before. Applied!

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/Connector.java
dawidxc: The alternative could be to use `ISO_LOCAL_DATE_TIME`.
https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_LOCAL_DATE_TIME

See example usage in existing code:
https://github.com/google/dwh-migration-tools/search?q=DateTimeFormatter

ishmum123:
Loaders should be ready now. Would merge it after @dawidxc approves

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/Connector.java
ishmum123: I wanted to get rid of all the `-`-s. Do let me know if you would prefer the  `ISO_LOCAL_DATE_TIME`

dawidxc: Makes sense. I didn't notice the `-` and `:` separators. Thank you for checking.

dawidxc (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/Connector.java
```diff
@@ -36,10 +38,20 @@ public enum DefaultProperties implements ConnectorProperty {
     public String getName();
 
     @Nonnull
-    public String getDefaultFileName();
+    default String getDefaultFileName(boolean isAssessment) {
+        Format format = new SimpleDateFormat("yyyyMMdd'T'HHmmss");
+        String timeSuffix = "-" + format.format(System.currentTimeMillis());
+        return String.format("dwh-migration-%s-%s%s.zip",
+            getName(), type(), isAssessment ? timeSuffix : ""
```
paolomorandini, line 45: Nitpick: use `StringUtils.EMPTY` instead of `""`

paolomorandini (approved):
Looks good, thank you. I left some small nitpicks.

PR #136, opened 2023-03-02
Fix b/271428453

mjquinn-google:
* Enable dependabot
* Update at least the subset of versions mentioned in b/271428453, plus a few others, excluding others that are not affected by security vulnerabilities


dumper/app/build.gradle
```diff
@@ -75,7 +75,7 @@ dependencies {
     testRuntimeOnly "org.datanucleus:datanucleus-rdbms:4.1.19"
 
     sources "org.slf4j:jcl-over-slf4j:1.7.14@sources"
-    sources "ch.qos.logback:logback-classic:1.2.3@sources"
+    sources "ch.qos.logback:logback-classic:1.4.5@sources"
```
shevek-google, line 78: These need to be in sync with the binaries, for legal reasons.

shevek-google: Question: Should we be specifying sources constraints, in the buildSrc where we have the known-version-numbers, rather than duplicating and syncing version numbers to here?

shevek-google (changes requested):

dumper/app/build.gradle
mjquinn-google: (Resolved in discussion; logback-classic binary has been updated to 1.4.5 in this PR)

shevek-google (approved):

paolomorandini (approved):

PR #138, opened 2023-03-08
add output-dir flag

ishmum123:


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -138,7 +138,8 @@ public class ConnectorArguments extends DefaultArguments {
     private final OptionSpec<String> optionOracleSID = parser.accepts(OPT_ORACLE_SID, "SID name for oracle").withRequiredArg().describedAs("orcl").ofType(String.class);
     private final OptionSpec<String> optionConfiguration = parser.accepts("config", "Configuration for DB connector").withRequiredArg().ofType(String.class).withValuesSeparatedBy(';').describedAs("key=val;key1=val1");
     // private final OptionSpec<String> optionDatabase = parser.accepts("database", "database (can be repeated; all if not specified)").withRequiredArg().describedAs("my_dbname").withValuesSeparatedBy(',');
-    private final OptionSpec<File> optionOutput = parser.accepts("output", "Output file").withRequiredArg().ofType(File.class).describedAs("cw-dump.zip");
+    private final OptionSpec<File> optionOutput = parser.accepts("output", "Output file").withOptionalArg().ofType(File.class).describedAs("cw-dump.zip");
```
dawidxc, line 141: 1. I think this option should keep the `withRequiredArg` config, because it's the argument of the option that is required (`my-file.zip` in the expression `--output my-file.zip`), not the option itself.
2. I would suggest to change the description "Output file" to "Output file (cannot be used together with --output-dir)".

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -138,7 +138,8 @@ public class ConnectorArguments extends DefaultArguments {
...
+    private final OptionSpec<String> optionOutputDir = parser.accepts("output-dir", "Output directory").withOptionalArg().ofType(String.class).defaultsTo("").describedAs("/home/user/...");
```
dawidxc, line 142: 1. I would recommend using `withRequiredArg` instead of `withOptionalArg` (similarly to the option above).
2. I would suggest to change the description from "Output directory" to "Output directory (cannot be used together with --output)".
3. nit: In my opinion, the description should not contain the ellipsis (`...`), but rather should be an example of a valid path, e.g. `/home/user/my-dump-dir`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
```diff
@@ -191,8 +191,14 @@ protected void run(@Nonnull Connector connector, @Nonnull ConnectorArguments arg
         // We had a customer request to base it on the database, but that isn't well-defined,
         // as there may be 0 or N databases in a single file.
         File outputFile = arguments.getOutputFile();
+        String outputDirectory = arguments.getOutputDirectory();
+
+        if ((outputDirectory == null) == (outputFile == null)) {
```
dawidxc, line 196: The condition `((outputDirectory == null) == (outputFile == null))` is difficult to understand. Consider extracting the subexpressions to appropriately named local variables and construct the condition using such variables to make the whole condition more readable.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
```diff
@@ -191,8 +191,14 @@ protected void run(@Nonnull Connector connector, @Nonnull ConnectorArguments arg
...
+            throw new IllegalStateException("Please use exactly one of --output & --output-dir flags");
```
dawidxc, line 197: I'm sorry, I wasn't very precise in my previous explanation "only one of the two should be allowed". What I missed to mention is that we also still want to accept no options.

We need to handle 3 possibilities:
1. Not `--output` nor `--output-dir` is used. This should be the default behavior as it worked before and it should not change. The file should be dumped in the current directory with the default filename.
2. Only `--output` is used. This should work the same as it worked before, i.e. if the argument is a filename ending with `.zip`, then dump to this file, otherwise dump to the directory or return an error. No change in the logic in this case.
3. Only `--output-dir` is used. This is the new logic being introduced in this PR. A default zip filename should be appended to the argument of this option and the result should be dumped to this zip file (i.e. a default zip file will be created in the directory provided by the `--output-dir` option).

If both options are used by mistake, then display an error and return immediately. Instead of throwing an exception, please, display an error in the user-friendly way (with the stars around), similarly to:
https://github.com/google/dwh-migration-tools/blob/ed75f929f475b27701bd71c3da120292f10b1127/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java#L308

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
dawidxc: I would suggest to use even more elaborate description, because "Output directory" seems confusing when compared with "Output file". For example:

"Output directory where the zip file with the default filename should be created (cannot be used together with --output option). If you want to specify the filename, use --output option instead.".

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
```diff
@@ -191,8 +191,22 @@ protected void run(@Nonnull Connector connector, @Nonnull ConnectorArguments arg
...
+                + "* Please use --help for more information.\n"
+                + "**********************************************************"
+            );
+            return;
+        }
+
         if (outputFile == null) {
-            outputFile = new File(connector.getDefaultFileName(arguments.isAssessment()));
+            String separator = "".equals(outputDirectory) ? "" : File.separator;
+            outputFile = new File(outputDirectory + separator + connector.getDefaultFileName(arguments.isAssessment()));
```
dawidxc, line 209: The two lines above can be simplified to:
```
outputFile = new File(outputDirectory, connector.getDefaultFileName(arguments.isAssessment()));
```

dawidxc (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -138,7 +138,8 @@ public class ConnectorArguments extends DefaultArguments {
...
     private final OptionSpec<String> optionConfiguration = parser.accepts("config", "Configuration for DB connector").withRequiredArg().ofType(String.class).withValuesSeparatedBy(';').describedAs("key=val;key1=val1");
     // private final OptionSpec<String> optionDatabase = parser.accepts("database", "database (can be repeated; all if not specified)").withRequiredArg().describedAs("my_dbname").withValuesSeparatedBy(',');
-    private final OptionSpec<File> optionOutput = parser.accepts("output", "Output file").withRequiredArg().ofType(File.class).describedAs("cw-dump.zip");
+    private final OptionSpec<File> optionOutput = parser.accepts("output", "Output file (cannot be used together with --output-dir)").withRequiredArg().ofType(File.class).describedAs("cw-dump.zip");
+    private final OptionSpec<String> optionOutputDir = parser.accepts("output-dir", "Output directory where the zip file with the default filename should be created (cannot be used together with --output option). If you want to specify the filename, use --output option instead.").withRequiredArg().ofType(String.class).defaultsTo("").describedAs("/home/user/my-dump-dir");
```
paolomorandini, line 142: Should this be `ofType(File.class)` like `--output` is?

ishmum123: We are assuming this is the folder name rather than a particular file. `String` seemed more appropriate in that case. Would love to know your thoughts on it

dawidxc: My opinion is that the documentation should not be misleading to the customer. `File` is better from the implementation perspective (because it can represent both a file and a directory). However, displaying such an implementation detail in the output of the `--help` command is in my opinion confusing.

The best solution would be to use `File` here, but then output either `Directory` or (better) nothing in the `--help` command output. I think we can prepare such an improvement in a separate PR.

paolomorandini (approved):
Works for me, please address this comment in this or a following PR:  https://github.com/google/dwh-migration-tools/pull/138/files#r1132466555

PR #142, opened 2023-03-10
Reformat code according to Google Style

dawidxc:
https://google.github.io/styleguide/intellij-java-google-style.xml


ishmum123:
Is there any way we can add a check for this? i.e. - check that each PR maintains this format?

haselbach (approved):

PR #157, opened 2023-03-15
merge `output` & `output-dir` flags

ishmum123:


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
```diff
@@ -229,21 +228,11 @@ protected void run(@Nonnull Connector connector, @Nonnull ConnectorArguments arg
...
-                + "* Please use --help for more information.\n"
-                + "**********************************************************"
-            );
-            return;
-        }
-
-        if (outputFile == null) {
-            outputFile = new File(outputDirectory, connector.getDefaultFileName(arguments.isAssessment()));
+        String defaultFileName = connector.getDefaultFileName(arguments.isAssessment());
+        if (outputFile == null || outputFile.getName().isEmpty()) {
```
dawidxc, line 232: I would suggest to avoid null, wherever possible. In this case, I would recommend changing the return type of `ConnectorArguments#getOutputFile()` method to `Optional<File>`. In addition to `null`, empty string can also be represented as empty optional, so that this condition is simplified.

See example:
https://github.com/google/dwh-migration-tools/blob/3161155bdbb796c30f8caee89d15276ad2032f74/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java#L441

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
```diff
@@ -229,21 +228,11 @@ protected void run(@Nonnull Connector connector, @Nonnull ConnectorArguments arg
...
+            outputFile = new File(defaultFileName);
+        } else if (!StringUtils.endsWithIgnoreCase(outputFile.getName(), ".zip")) {
+            outputFile = new File(outputFile.getName(), defaultFileName);
```
dawidxc, line 235: One more check is needed before this line. If the path passed in the `outputFile` exists and it is a regular file (not a directory nor a directory that does not exist yet), then the dumper should fail.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -160,7 +160,6 @@ public class ConnectorArguments extends DefaultArguments {
     private final OptionSpec<String> optionConfiguration = parser.accepts("config", "Configuration for DB connector").withRequiredArg().ofType(String.class).withValuesSeparatedBy(';').describedAs("key=val;key1=val1");
     // private final OptionSpec<String> optionDatabase = parser.accepts("database", "database (can be repeated; all if not specified)").withRequiredArg().describedAs("my_dbname").withValuesSeparatedBy(',');
     private final OptionSpec<File> optionOutput = parser.accepts("output", "Output file (cannot be used together with --output-dir)").withRequiredArg().ofType(File.class).describedAs("cw-dump.zip");
-    private final OptionSpec<String> optionOutputDir = parser.accepts("output-dir", "Output directory where the zip file with the default filename should be created (cannot be used together with --output option). If you want to specify the filename, use --output option instead.").withRequiredArg().ofType(String.class).defaultsTo("").describedAs("/home/user/my-dump-dir");
     private final OptionSpec<Void> optionOutputContinue = parser.accepts("continue", "Continues writing a previous output file.");
```
damian-compilerworks, line 163: Is this option still supported?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -160,7 +160,6 @@ public class ConnectorArguments extends DefaultArguments {
     private final OptionSpec<String> optionConfiguration = parser.accepts("config", "Configuration for DB connector").withRequiredArg().ofType(String.class).withValuesSeparatedBy(';').describedAs("key=val;key1=val1");
     // private final OptionSpec<String> optionDatabase = parser.accepts("database", "database (can be repeated; all if not specified)").withRequiredArg().describedAs("my_dbname").withValuesSeparatedBy(',');
     private final OptionSpec<File> optionOutput = parser.accepts("output", "Output file (cannot be used together with --output-dir)").withRequiredArg().ofType(File.class).describedAs("cw-dump.zip");
```
damian-compilerworks, line 162: Can we update the help message to match the current, extended semantics?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -160,7 +160,6 @@ public class ConnectorArguments extends DefaultArguments {
...
-    private final OptionSpec<String> optionOutputDir = parser.accepts("output-dir", "Output directory where the zip file with the default filename should be created (cannot be used together with --output option). If you want to specify the filename, use --output option instead.").withRequiredArg().ofType(String.class).defaultsTo("").describedAs("/home/user/my-dump-dir");
```
damian-compilerworks, line 163: Deleting the option is backwards incompatible change. Can we do that without the notice?
If we can, then I think it's fair to keep the old option, as a transitory measure / a grace period, but print a message and terminate the program if the option is supplied by the user. The message should cover the new ways of achieving the same thing (or explaining it's no longer supported).

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
ishmum123: no it's not

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
ishmum123: the flag was introduced in #138 2 days ago. I felt it's safe to remove for this reason.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
```diff
@@ -348,6 +308,30 @@ public <T> T runChildTask(Task<T> task) throws MetadataDumperUsageException {
...
+        .getOutputFile()
+        .map(
+            file -> {
+              String fileName = file.getName();
+              if (StringUtils.endsWithIgnoreCase(fileName, ".zip")) return file;
+              if (file.isFile()) {
+                throw new IllegalStateException(
+                    String.format(
+                        "A file already exists at %1$s. "
+                            + "If you wished to create a directory, please remove/rename the file first. "
```
dawidxc, line 326: I would suggest to not mix `wished` and `meant`. I think the intention is the same in both cases, so I would stick to only one verb.

dawidxc, line 326: I would suggest to change
"remove/rename the file first."
to
"provide the path to the directory."

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
```diff
@@ -348,6 +308,30 @@ public <T> T runChildTask(Task<T> task) throws MetadataDumperUsageException {
...
+                            + "If you meant to create %1$s.zip, please add the `.zip` extension manually",
```
dawidxc, line 327: nit: A dot is missing at the end.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
```diff
@@ -348,6 +308,30 @@ public <T> T runChildTask(Task<T> task) throws MetadataDumperUsageException {
...
+                        fileName));
+              }
+              return new File(fileName, defaultFileName);
```
dawidxc, line 330: The `fileName` String does not contain the full path. I would suggest to change to:
```
return new File(file, defaultFileName);
```

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -237,19 +237,12 @@ public class ConnectorArguments extends DefaultArguments {
   // specified)").withRequiredArg().describedAs("my_dbname").withValuesSeparatedBy(',');
   private final OptionSpec<File> optionOutput =
       parser
-          .accepts("output", "Output file (cannot be used together with --output-dir)")
+          .accepts(
+              "output",
+              "Output file or directory name. If the file name, along with the `.zip` extension, is not provided dumper will attempt to create the zip file with default file name in the directory")
```
dawidxc, line 242: nit 1: "default file name" => "the default file name"
nit 2: A dot is missing at the end.

dawidxc (approved):

dawidxc (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
damian-compilerworks: Not sure about the release process, but `output-dir` nevertheless was part of the https://github.com/google/dwh-migration-tools/releases/tag/v1.0.17

I'll allow that, since it's indeed a very fresh one, but I think in the future we must be careful in what changes we introduce (and revert!) to public API, especially if their logic is `isAssessment`-independent.

damian-compilerworks (approved):

PR #161, opened 2023-03-15
Add connector-specific max-rows options.

dawidxc:
Add a possibility to limit the number of rows extracted from DBC.DiskSpaceV table during teradata metadata extraction.

New connector-specific options:
- for TableSizeV: `-Dteradata.metadata.tablesizev.max-rows`
- for DiskSpaceV: `-Dteradata.metadata.diskspacev.max-rows`
- for users in DatabasesV: `-Dteradata.metadata.databasesv.users.max-rows`
- for databases in DatabasesV: `-Dteradata.metadata.databasesv.dbs.max-rows`

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
@@ -95,9 +95,6 @@ public class ConnectorArguments extends DefaultArguments {
   public static final String OPT_DATABASE = "database";
   public static final String OPT_SCHEMA = "schema";
   public static final String OPT_ASSESSMENT = "assessment";
-  public static final String OPT_TERADATA_MAX_TABLESIZEV_ROWS = "max-tablesizev-rows";
-  public static final String OPT_TERADATA_MAX_DATABASESV_USER_ROWS = "max-databasesv-user-rows";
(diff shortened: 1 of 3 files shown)


paolomorandini (commented):
Do we need all of these distinct `max-rows` arguments? Why not have a single one?

cyulysses-corp (approved):

shevek-google (commented):
I've seen a few CLs now which expose Teradata-specific parameters in the global --help namespace. Do we have an alternative to this which keeps the parameters more "local" to Teradata? I'm concerned that --help becomes less useful the more text it emits.

dawidxc:
I will prepare a separate PR to remove the Teradata-specific `--max-rows` option from the global `--help` output (this option is introduced in PR #171).

shevek-google:
> I will prepare a separate PR to remove the Teradata-specific `--max-rows` option from the global `--help` output (this option is introduced in PR #171).

I think I'm curious about our general strategy for new options. Do we have a potential way to make them connector-local? How many of these options are we likely to have? Should connectors have a generic -D-like parameter for passing arguments? Should arguments after a connector be passed to the connector? I'm hunting for a general usability strategy. Hiding flags is one of the options, but may not be the best one.

dawidxc:
I have merged PR #171 into this PR.

paolomorandini (approved):

PR #173, opened 2023-03-22
Update README.md

davidkhala:
The original git clone will hang in some environment. 

Use more safe one.


davidkhala:
Hi @ajwelch4,  I want to add one improvement to this again.

ajwelch4 (approved):

PR #174, opened 2023-03-22
Optimize Hive metadata connector partitions metadata performance by constructing partition name instead of requesting it

frsv:
Currently, metadata connector retrieves all partition metadata one by one to match it with partition names. This creates a lot of RPCs and degrades the performance of the connector drastically –  relatively small setup with ~20 tables, having 200K partitions in total, took more than 2 hours to dump, with the overall zip size being only 1.5 MB.

Internally, hive always keeps partition name in `partitionKey1=partitionValue1/partitionKey2=partitionValue2` format (see source: https://github.com/apache/hive/blob/rel/release-2.3.6/metastore/src/java/org/apache/hadoop/hive/metastore/Warehouse.java#L315). For managed tables (or others following Hive partition naming style), partition name is a part of the partition location. For external tables, not necessary following the Hive style, location might differ. For example, I managed to create a following pair:

* Location: `hdfs://bq-assessment-hive-m/user/hive/warehouse/rf.db/rf_many_partitions/test_10002`
* Partition name: `ws_sold_date_sk=10002`

This removes the request of the partition names as redundant and instead constructs it on the fly for a given partition using partition values list from the partition data and partition keys list from the table. The change reduces number of RPCs to only 1 per table and it speeds up the connector a lot - the same dump with the change took only 50 seconds.


dumper/lib-ext-hive-metastore/src/main/java/com/google/edwmigration/dumper/ext/hive/metastore/utils/PartitionNameGenerator.java
```diff
@@ -0,0 +1,92 @@
...
+                        entry.getKey()));
+                    return "";
+                  }
+                  return escapePartitionPart(entry.getKey()) + "=" + escapePartitionPart(
+                      entry.getValue());
+                })
+            .filter(part -> !part.isEmpty())
+            .collect(toImmutableList());
+
+    return Joiner.on('/').join(partitionParts);
```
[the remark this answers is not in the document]
frsv: Done

dumper/lib-ext-hive-metastore/src/main/java/com/google/edwmigration/dumper/ext/hive/metastore/HiveMetastoreThriftClient_v2_3_6.java
```diff
@@ -311,90 +323,94 @@ public String getComment() {
...
+                    @Nonnull
+                    @Override
+                    public String getPartitionName() {
+                      return partitionName;
+                    }
+
+                    @CheckForNull
+                    @Override
+                    public String getLocation() {
+                      return (partition.isSetSd() && partition.getSd().isSetLocation()
```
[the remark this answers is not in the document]
frsv: It's wrapped like this in different places as well. I'll remove it in a separate PR

shevek-google (commented):
Is it possible to ALTER TABLE ADD PARTITION with a different path to a managed table?

frsv:
> Is it possible to ALTER TABLE ADD PARTITION with a different path to a managed table?

Yes. I managed to add a partition which points to other directory:

```
ALTER TABLE rf_5k_partitions_1 ADD PARTITION (ws_sold_date_sk=6001) LOCATION 'some-random-path/directory/test/test_6001';
```

This doesn't load the data though (it's a different command), but I expect it to work. It's also possible to point to a full HDFS path. The internal partition name doesn't change though:

```
mysql> SELECT PART_NAME FROM PARTITIONS WHERE PART_ID>213305;
+----------------------+
| PART_NAME            |
+----------------------+
| ws_sold_date_sk=6000 |
| ws_sold_date_sk=6001 |
+----------------------+

mysql> SELECT LOCATION  FROM SDS WHERE SD_ID>213450;
+--------------------------------------------------------------------------------------------------------------------+
| LOCATION                                                                                                           |
+--------------------------------------------------------------------------------------------------------------------+
| hdfs://bq-assessment-hive-m/user/hive/warehouse/rf.db/rf_5k_partitions_1/test_6000                                 |
| hdfs://bq-assessment-hive-m/user/hive/warehouse/rf.db/rf_5k_partitions_1/some-random-path/directory/test/test_6001 |
+--------------------------------------------------------------------------------------------------------------------+
```

shevek-google:
> This doesn't load the data though (it's a different command), but I expect it to work. It's also possible to point to a full HDFS path.
I think this is typically used to tell Hive to read data which was preloaded into HDFS by an external job, and is often followed by an MSCK REPAIR command.

> The internal partition name doesn't change though:
This is because the internal partition "name" is actually the KV data. It gets parsed, it's not just a name.

frsv:
Yes, so parsing partition names from the internal state (keys and values from `get_table` and `get_partitions` responses are already available and sorted) is sufficient. The location will be also correct, as it is a part of the `get_partitions` response item.

dumper/lib-ext-hive-metastore/src/main/java/com/google/edwmigration/dumper/ext/hive/metastore/utils/PartitionNameGenerator.java
```diff
@@ -0,0 +1,77 @@
...
+  public static String makePartitionName(List<String> partitionKeys, List<String> partitionValues) {
+    return Streams.zip(
+            partitionKeys.stream(),
+            partitionValues.stream(),
+            PartitionNameGenerator::constructPartitionName)
+        .flatMap(Function.identity())
+        .collect(Collectors.joining("/"));
+  }
+
+  private static Stream<String> constructPartitionName(String partitionKey, String partitionValue) {
```
paolomorandini, line 60: Can't this return a `String` and we drop the `flatMap` above?

frsv: `Stream<String>` and `Optional<String>` could be used interchangeably in this context, but `Optional::stream` is since Java 9. I don't see it in this repo, but pretty much sure it was mentioned that Java 8 is required:

```diff
-        .flatMap(Function.identity())
+        .flatMap(Optional::stream)
```

As for returning normal String – I will need to filter empty ones. I would prefer going with Optionals if Java 8 is not a requirement anymore though:

```diff
-        .flatMap(Function.identity())
+        .filter(part -> !part.isEmpty())
```

WDYT?

.DS_Store
dawidxc, line 1: `.DS_Store` - it looks like a mistakenly commited file. If so, could you remove it, please?

dawidxc: Possibly also add to `.gitignore` to prevent it being included in the future.

dumper/lib-ext-hive-metastore/src/main/java/com/google/edwmigration/dumper/ext/hive/metastore/utils/PartitionNameGenerator.java
paolomorandini: Let's stay on Java 8 for the time being.
Since the dumper is probably using lots of null checking we could also have `return null` and `Objects::nonNull` as filter predicate.

paolomorandini (approved):

PR #176, opened 2023-03-24
create GitHub workflow

davidkhala:
Along with an adjustment to cover more general install cases

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -1,0 +1,40 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python
+
+name: Python package
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.8", "3.9", "3.10"]
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v3
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install flake8 pytest
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+    - name: Lint with flake8
+      run: |
+        # stop the build if there are Python syntax errors or undefined names
+        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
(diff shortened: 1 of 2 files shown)


.github/workflows/python-package.yml
```diff
@@ -0,0 +1,33 @@
...
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+  venv:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.8", "3.9", "3.10"]
```
shevek-google, line 18: We claim python 3.7.2 in deps, but don't test it here.

shevek-google (approved):

.github/workflows/python-package.yml
davidkhala: @shevek-google Good suggestion.

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,33 @@
...
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+  venv:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
```
ajwelch4, line 18: @davidkhala We already use `nox` to run tests under different Python versions:
- https://github.com/google/dwh-migration-tools/blob/main/client/noxfile.py#L52
- https://github.com/google/dwh-migration-tools/blob/main/client/CONTRIBUTING.md#test

Also, we have formatting, linting and type checking:
- https://github.com/google/dwh-migration-tools/blob/main/client/CONTRIBUTING.md#test

So far we have just been running everything manually. Is it possible to take advantage of that here? @paolomorandini does it matter whether these checks are run via GH actions or our existing kokoro scripts?

paolomorandini: I think we can use GH actions as long as we don't need secrets.

ajwelch4: Integration/load tests will need secrets.

client/README.md
```diff
@@ -65,7 +63,9 @@ export BQMS_GCS_BUCKET="<YOUR_GCS_BUCKET>"
 
 ## Installation
 
-Prerequisites: Python >= 3.7.2.
+Prerequisites:
+- Python >= 3.7.2.
+- package `pkg-config` and `libicu-dev`
```
ajwelch4, line 68: @davidkhala Can you create a separate PR for this as this should be merged asap? Thanks.

ajwelch4 (changes requested):

client/README.md
davidkhala: @ajwelch4 good point, will do.

.github/workflows/python-package.yml
davidkhala: @ajwelch4 agree, will take a look to see if we can reuse those existing test script into GH workflow.

davidkhala:
Hi maintainers, I think this PR has been stale and I might no pay attention to it in near future. I prefer to make it back as issue and let you decide how to implement. 

Anyway, thanks for you past time spending on this.

PR #177, opened 2023-03-24
Fix gradlew path in bin/dwh-migration-dumper

haselbach:
Gradle must be executed from the main directory of the project. Thus, the script bin/dwh-migration-dumper only works if executed from this path. To allow it being run from anywhere, use a subshell to cd to the directory and execute gradlew.


dawidxc (approved):
I have tested the script and it works correctly when executed from another directory (both ancestor and non-ancestor).

paolomorandini (approved):

PR #193, opened 2023-04-09
Update README.md

davidkhala:
splited from #176

PR #195, opened 2023-04-11
Various improvements to first-time user documentation in READMEs

arotenberg-google:
This PR is intended to help address several problems we've seen with first-time user experience using these tools.

- Users were sometimes confused about the relationship between the tools in this repo. This PR adds a `README.md` for the dumper, which previously did not have one, and changes some wording to help clarify that these are separate tools in one repository.
- A common issue with users trying to use the dumper is that they were trying to build it themselves from source when there is no reason to and they could just be using the binary release. This PR adds a line at the top of all three `README.md` files directing the user to download the zip from the project's Releases page.
- The documentation in `client/README.md` previously did not clearly indicate the full set of prerequisites necessary before executing the commands in the "Quickstart" section. (This PR is intended to supersede #193, which also addressed part of this.)
- The documentation in `client/README.md` previously placed the "Recommended usage" below the instructions for the not-recommended direct usage. This PR reorders the sections so they consistently go from simpler use cases to more complicated ones, and more clearly distinguishes the `run.sh` usage from the `bqms-run` usage.

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -12,10 +12,10 @@ this client][how to submit].
 
 - [Installation](#installation)
 - [Quickstart](#quickstart)
+- [config.yaml](#configyaml)
 - [Running Using `run.sh`](#running-using-runsh)
 - [Running Using `bqms-run` Directly](#running-using-bqms-run-directly)
   - [Environment Variables](#environment-variables)
-- [config.yaml](#configyaml)
 - [Metadata Lookup](#metadata-lookup)
   - [DDL and Metadata Dump](#ddl-and-metadata-dump)
   - [Unqualified References](#unqualified-references)
@@ -32,12 +32,11 @@ this client][how to submit].
 Preferred OS: Linux or macOS. Windows usage may be possible through PowerShell,
 but it is not officially supported.
 
-**Download and extract the latest release zip `dwh-migration-tools-vX.X.X.zip`
-from [the Releases page](https://github.com/google/dwh-migration-tools/releases).**
+**Download and extract [the latest release zip `dwh-migration-tools-vX.X.X.zip`](https://github.com/google/dwh-migration-tools/releases/latest).**
 
 Python &ge; 3.7.2 is required, as well as the additional local dependencies
 `pkg-config` and `libicu-dev`. Typical commands for installing these
-dependencies on Linux would be:
+dependencies on a Debian-based Linux distribution such as Ubuntu would be:
 
 ```shell
 sudo apt update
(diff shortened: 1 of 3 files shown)


dumper/README.md
```diff
@@ -0,0 +1,16 @@
+# BigQuery Migration Service Metadata and Log Dumper
+
+This directory contains the Metadata and Log Dumper, a command line tool for
+connecting to an existing database and generating an archive of DDL metadata or
+logs. This tool generates archives in a format suitable for consumption by the
+[BigQuery Migration Service's][BQMS] Assessment or Translation Service.
+
+The Dumper is a Java tool. **Download the latest cross-platform release zip
+`dwh-migration-tools-vX.X.X.zip` from
+[the Releases page](https://github.com/google/dwh-migration-tools/releases).**
```
paolomorandini, line 10: We can point to latest release: https://github.com/google/dwh-migration-tools/releases/latest

client/README.md
```diff
@@ -1,36 +1,64 @@
...
 - [Performance](#performance)
 
-## Quickstart
+## Installation
+
+Preferred OS: Linux or macOS. Windows usage may be possible through PowerShell,
+but it is not officially supported.
+
+**Download and extract the latest release zip `dwh-migration-tools-vX.X.X.zip`
+from [the Releases page](https://github.com/google/dwh-migration-tools/releases).**
```
xiashang0624, line 36: Why do we want to enforce users to download zip from the release for this python client?  Have we seen any versioning issues for python client before?

client/README.md
```diff
@@ -224,6 +172,89 @@ following shell scripts for convenience:
...
+A Docker container of `bqms-run` with all the necessary dependencies can be
+built by running:
+
+```shell
+sudo docker build -t bqms-run client
+```
+
+See also the instructions for [deploying to Cloud Run](#deploying-to-cloud-run).
+
+### Environment Variables
```
xiashang0624, line 192: The Env Vars and the config.yaml sections are also important to first time users. I think moving this section upper on this page might be easier for beginner to get started.

client/README.md
arotenberg-google: There are at least three reasons to want to point users at the releases rather than telling them to `git clone` the source repository:

1. We have had a nontrivial number of users contact our support because they were having issues getting `git clone` to work. Telling users to just download a release short-circuits that entire problem.
2. With the Java dumper, we have had many users contact our support because they were failing to build the source code themselves while apparently being unaware that a binary build was available because it was not clearly mentioned in the documentation. While we do not currently have any comparable build activities for the Python client that would motivate wanting users to download a release over using the GitHub source, directing users to download releases future-proofs these customers' usage workflows against any such build processes we introduce to the client release packaging in the future.
3. When communicating with users about changes to the client, or determining whether a user is using a recent-enough client version to have a certain feature, it is easier to refer to a release version number than to a GitHub hash or download date.

client/README.md
```diff
@@ -1,36 +1,64 @@
...
+
+Preferred OS: Linux or macOS. Windows usage may be possible through PowerShell,
+but it is not officially supported.
+
+**Download and extract the latest release zip `dwh-migration-tools-vX.X.X.zip`
+from [the Releases page](https://github.com/google/dwh-migration-tools/releases).**
+
+Python &ge; 3.7.2 is required, as well as the additional local dependencies
+`pkg-config` and `libicu-dev`. Typical commands for installing these
+dependencies on Linux would be:
```
mjquinn-google, line 40: Nit: "Debian-based Linux" would be more appropriate here given the use of `apt` below, but anyone who really needs that clarification is probably going to be lost anyway, so up to you.

mjquinn-google: (it seems like at least one customer is interested in RHEL, so this may need to be split out into separate sections eventually)

davidkhala: @mjquinn-google +1, it alway introduce debate on `what should be the preference linux distro`. I suggest we only provide the dependency package name and let reader to find their own way to get it install.

dumper/README.md
arotenberg-google: I'm ambivalent pointing users at a page that only shows the latest release, but I went ahead and changed all the links to do this since you can get the full list of releases from the breadcrumbs bar on that page anyway if you need it.

PR #196, opened 2023-04-11
Remove the bin/dwh-migration-client script, as the executable it referred to no longer exists

arotenberg-google:
[The docs on the GCP site](https://cloud.google.com/bigquery/docs/batch-sql-translator#submit_a_translation_job) and the `bin` scripts included in the release zip refer to an executable named `dwh-migration-client`, but that command doesn't exist anymore, there's only `run.sh` and `bqms-run`.

There will need to be a separate Google-internal change to update the GCP docs.

The current behavior of the client on a clean `pip install` can be confirmed using [this Dockerfile](https://gist.github.com/arotenberg-google/9286f6856d43b95465281e73da64a70a):

```shell
> whereis dwh-migration-client
dwh-migration-client:
> whereis bqms-run            
bqms-run: /usr/local/bin/bqms-run
```

### DIFF
--- a/bin/dwh-migration-client
+++ b/bin/dwh-migration-client
@@ -1,3 +1,0 @@
-#!/bin/sh -ex
-
-exec dwh-migration-client "$@"
--- a/build.gradle
+++ b/build.gradle
@@ -11,9 +11,6 @@ distributions {
                 exclude "**/__pycache__"
                 into "client"
             }
-            from("bin/dwh-migration-client") {
-                into "bin"
-            }
             project(":dumper:app").afterEvaluate {
                 from it.tasks.installPublishedDist
             }


Hackerpilot (approved):

shevek-google:
Should we update the code to match the docs, or the docs to match the code? dwh-migration-client is a better name than run.sh

ajwelch4:
Both `run.sh` and `bqms-run` are admittedly bad names. I am not good at naming ;-).

If we exchange `SCRIPT_DIR` ([defined here](https://github.com/google/dwh-migration-tools/blob/main/client/examples/teradata/sql/run.sh#L71)) for `$PWD` then we should be able to pull `run.sh` into a top-level `bin` directory as long as `$PWD` contains `input` and `config` dirs. And of course we can rename `run.sh` as well and update the docs. 

`clean.sh` and `deprovision.sh` are trivial and could just be deleted. `provision.sh` shows all of the `gcloud `commands you would need to run if you were starting from zero without a project and needed to provision the correct resources to run locally or on Cloud Run. It is truly meant as an example as no one really uses a bash script with `gcloud` commands as a devops tool. I created it simply to spin up demo environments from scratch and as executable documentation. Not sure where that belongs, if anywhere.

PR #206, opened 2023-05-16
Fix --database global flag help message.

dawidxc:
The `--database` command-line flag depends on the connector used. It can be a single database name or a comma-separated list of databases. Sometimes it is used for connecting to the database (e.g. Redshift connectors), sometimes it is used for extracting only specified databases (e.g. the `teradata` connector). Sometimes this flag is optional, sometimes it is required.


shevek-google (commented):
Would this be better if we started design with a table of:
* Use the database for connection information
* Use the database to filter what is dumped
* It actually makes a difference, i.e. connection info != dump info

Can we consider it always to be a filter database, and if it's optionally used as a connection database, that's a bonus?

PR #212, opened 2023-06-08
Split out a dump runner out of MetadataDumper

haselbach:
To prepare creating an alternative main class for the Cloud Extractor, split out the re-usable part into `DumpRunner` and call it from `MetadataDumper`.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/Main.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/Main.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2022-2023 Google LLC
+ * Copyright 2013-2021 CompilerWorks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.edwmigration.dumper.application.dumper;
+
+import javax.annotation.Nonnull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** @author miguel */
+public class Main {
+
+  private static final Logger LOG = LoggerFactory.getLogger(Main.class);
+
+  private final MetadataDumper metadataDumper;
+
+  public Main(MetadataDumper metadataDumper) {
+    this.metadataDumper = metadataDumper;
+  }
+
+  public void run(@Nonnull String... args) throws Exception {
+    ConnectorArguments arguments = new ConnectorArguments(args);
+    try {
+      metadataDumper.run(arguments);
+    } finally {
+      if (arguments.saveResponseFile()) {
+        JsonResponseFile.save(arguments);
+      }
+    }
+  }
+
+  public static void main(String... args) throws Exception {
+    try {
+      Main main = new Main(new MetadataDumper());
+      args = JsonResponseFile.addResponseFiles(args);
+      // LOG.debug("Arguments are: [" + String.join("] [", args) + "]");
+      // Without this, the dumper prints "Missing required arguments:[connector]"
+      if (args.length == 0) {
+        args = new String[] {"--help"};
+      }
+      main.run(args);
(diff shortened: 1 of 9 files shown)


shevek-google (dismissed):
The MetadataDumper object is already philosophically performing the task of the DumpRunner: The business logic object that does the dump. Would a less intrusive way to do this change be:
1. Overload MetadataDumper.run() to take String... and ConnectorArguments
2. Add a the exitOnError as a (quite possibly mutable) field to MetadataDumper
This would be the same, but just move less code around?

haselbach:
This is basically just splitting the MetadataDumper into two files. GitHub's diff unfortunately does not seem to be smart enough to detect that. Smarter diffs would show this as rather removing lines from the split. So, I'd say that this isn't really moving much code around. It's just the presentation that makes it look as such.

I could also just use the MetadataDumper as is but IMHO it is cleaner to have a common dump runner w/o a main method and then two classes with main methods that use it.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataMetadataConnectorTest.java
```diff
@@ -273,9 +275,11 @@ public Void doInConnection(Connection connection)
             // "--database", "db_0"
             );
 
-    MetadataDumper dumper = new MetadataDumper().withExitOnError(false);
+    MetadataDumper dumper =
+        new MetadataDumper(new DumpRunner.Builder().withExitOnError(false).build());
 
     dumper.run(args.toArray(ArrayUtils.EMPTY_STRING_ARRAY));
+    DumpRunner.getInstance().run(new ConnectorArguments());
```
dawidxc, line 282: The `DumpRunner#run` method is invoked inside the `MetadataDumper#run` method.

Did you mean to execute both `run` methods in here?

haselbach:
> The MetadataDumper object is already philosophically performing the task of the DumpRunner: The business logic object that does the dump. Would a less intrusive way to do this change be:
> 
> 1. Overload MetadataDumper.run() to take String... and ConnectorArguments
> 2. Add a the exitOnError as a (quite possibly mutable) field to MetadataDumper
>    This would be the same, but just move less code around?

As I said above, I don't believe this change to be intrusive. I'm basically renaming `MetadataDumper` to `DumperRunner` and just retaining the `main` method in `MetadataDumper`. To make this more obvious, I am splitting this PR into two. This is the first step where I copy the `MetadataDumper`. The next step is in https://github.com/google/dwh-migration-tools/pull/221

dawidxc (approved):
Approving, but only as a split that is continued in PR#221, as explained in other comments. I am not reviewing the code itself.

https://github.com/google/dwh-migration-tools/pull/221

haselbach:
After discussing with Shevek, we agreed to rather keep the `MetadataDumper` class and move the `main` method into a new class called `Main`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/Main.java
```diff
@@ -0,0 +1,60 @@
...
+  public void run(@Nonnull String... args) throws Exception {
+    ConnectorArguments arguments = new ConnectorArguments(args);
+    try {
+      metadataDumper.run(arguments);
+    } finally {
+      if (arguments.saveResponseFile()) {
+        JsonResponseFile.save(arguments);
+      }
+    }
+  }
```
shevek-google, line 43: Question: Why did this run() overload move out of the MetadataDumper?

shevek-google (approved):
Basically, yes. Much cleaner, thank you.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
```diff
@@ -69,7 +69,7 @@
 /** @author miguel */
 public class MetadataDumper {
 
-  private static final Logger LOG = LoggerFactory.getLogger(MetadataDumper.class);
+  private static final Logger LOG = LoggerFactory.getLogger(Main.class);
```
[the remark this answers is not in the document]
haselbach: Yes, thanks. Fixed.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/Main.java
haselbach: It is pretty specific to the main dumper invocation. IMHO it would also be misleading to have this has an overload as that would suggest it only converts `String ...` to `ConnectorArguments` but it also adds JSON response file handling. 

So, if we want to keep it in MetadataDumper despite only being needed in the main dumper, I would suggest to rename it to `runWithJsonResponseFileHandling`.

PR #220, opened 2023-06-23
dump results of SHOW WAREHOUSES for Snowflake

dominik-klimczak:
- it seems the only way to get the size of the warehouse


bdw-g (approved):
lgtm

PR #221, opened 2023-06-23
Split out a dump runner out of MetadataDumper

haselbach:
To prepare creating an alternative main class for the Cloud Extractor, plit out the re-usable part into `DumpRunner` and call it from `MetadataDumper`.


dawidxc (approved):

haselbach:
This is no longer needed as the change is now done completely in https://github.com/google/dwh-migration-tools/pull/212

PR #227, opened 2023-07-06
Implement the cloud dumper main class

haselbach:
The cloud dumper main class will be used to invoke the dumper from a GCE instance. It is intended to be invoked without arguments and get its configuratio from the instance metadata.

BUG=290102538

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/clouddumper/Main.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/clouddumper/Main.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2022-2023 Google LLC
+ * Copyright 2013-2021 CompilerWorks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.edwmigration.dumper.application.dumper.clouddumper;
+
+import com.google.edwmigration.dumper.application.dumper.ConnectorArguments;
+import com.google.edwmigration.dumper.application.dumper.MetadataDumper;
+import com.google.edwmigration.dumper.application.dumper.MetadataDumperUsageException;
+import com.google.gson.Gson;
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.hc.client5.http.impl.classic.HttpClients;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class Main {
+
+  private static final Logger LOG =
+      LoggerFactory.getLogger(com.google.edwmigration.dumper.application.dumper.Main.class);
+
+  private static final Gson GSON = new Gson();
+
+  private final MetadataDumper metadataDumper;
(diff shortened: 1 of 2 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/clouddumper/Main.java
```diff
@@ -0,0 +1,77 @@
...
+            .orElseThrow(
+                () ->
+                    new IllegalStateException(
+                        ("Attribute dwh_extractor_configuration is not defined.")));
+    ArrayList<String> args = new ArrayList<>();
+    args.add("--connector");
+    args.add(connectorName);
+    args.addAll(dumperConfig.args);
+    ConnectorArguments arguments = new ConnectorArguments(args.toArray(new String[args.size()]));
+    metadataDumper.run(arguments);
```
shevek-google, line 62: Could this be done as a main which just retrieves a String[] (DumperConfiguration) from MetadataRetriever, then invokes Main.main(String...)? Is there more complexity than is required here? What does the existing Main do that we can't make this do just by retrieving the String[] by HTTP rather than originating it from the CLI?

shevek-google (changes requested):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/clouddumper/Main.java
haselbach: I don't understand the criticism here. The metadataDumper wants a `ConnectorArguments` instance. So, that's what we provide here.

We need to massage the arguments a bit. For one, the configuration retrieved via the metadata might be encrypted. Decryption is added in an upcoming PR. Then there is the connector that needs to be added.

The connector name is handled separately as it is also needed to retrieve the right JDBC driver. The driver argument also needs to be added to arguments array.

haselbach: PR for decryption: https://github.com/google/dwh-migration-tools/pull/230

haselbach: If you insist on having a way to pass in args directly to `MetadataDumper` I suggest to use https://github.com/google/dwh-migration-tools/pull/231.

Another upside of keeping it as is: It uses `ConnectorArguments` as a domain class which would ultimately allow to set up a run programmatically through it instead of manipulating string arrays. This would require additional effort but would at least enable it.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/clouddumper/Main.java
```diff
@@ -0,0 +1,89 @@
...
+      args.addAll(connectorConfiguration.args);
+      ConnectorArguments arguments = new ConnectorArguments(args.toArray(new String[args.size()]));
+      metadataDumperSupplier.get().run(arguments);
+    }
+  }
+
+  public static void main(String... args) throws Exception {
+    try {
+      new Main(
+              () -> new MetadataDumper(),
```
shevek-google, line 71: MetadataDumper::new

shevek-google (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/clouddumper/Main.java
shevek-google: We discussed; I don't think CA ever becomes a domain class, because it has a purpose, which is to adapt the String[] to an API. There is potential for an interface, as anyone who constructs it has to decode _something_, I think any POJO that backs it as a domain object would probably be tied to some sort of serialization mechanism.

haselbach: Fair enough. I'll update https://github.com/google/dwh-migration-tools/pull/231 after pushing this.

PR #229, opened 2023-07-07
Add launch script for the cloud extractor

haselbach:
Add a shell script to be used as a launch script for GCE instances that
will download the dumper and then invoke the cloud extractor main class.

BUG=286327910

### DIFF
--- a/dumper/app/build.gradle
+++ b/dumper/app/build.gradle
@@ -151,9 +151,9 @@ distributions {
 		distributionBaseName = "dwh-migration-dumper"
 		contents {
 			from installDist
-            from (cloudExtractorStartScripts) {
-                into "bin"
-            }
+			from (cloudExtractorStartScripts) {
+				into "bin"
+			}
 			from(generateLicenseReport) {
 				into "docs/licenses"
 			}
--- a/dumper/app/build.gradle
+++ b/dumper/app/build.gradle
@@ -14,6 +14,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import org.apache.tools.ant.filters.ReplaceTokens
+
 buildscript {
     dependencies {
         classpath 'com.github.jk1:gradle-license-report:1.17'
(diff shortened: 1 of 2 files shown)


dumper/app/src/main/sh/cloud_extractor/gce_launcher.sh
```diff
@@ -0,0 +1,131 @@
...
+  dwh_tmp="${DWH_TMP:-$(mktemp -d "${TMPDIR}/dwh.XXX")}"
+  dwh_zip="${dwh_tmp}/dwh-migration-tools.zip"
+
+  dwh_url="$(get_attribute dwh_download_url)"
+
+  log INFO 'Downloading the dumper.'
+  download "${dwh_url:-${dwh_url_default}}" "${dwh_zip}"
+
+  unzip -o "${dwh_zip}" -d "${dwh_tmp}"
+  dwh_home="$(find "${dwh_tmp}" -maxdepth 1 -name dwh-migration-tools -o -name dwh-migration-tools-*)"
```
Hackerpilot, line 108: Shellcheck says `SC2061 (warning): Quote the parameter to -name so the shell won't interpret it.` but the thing that it's warning about might be the intended behavior.

Hackerpilot (approved):

dumper/app/build.gradle
```diff
@@ -135,6 +151,9 @@ distributions {
 		distributionBaseName = "dwh-migration-dumper"
 		contents {
 			from installDist
+            from (cloudExtractorStartScripts) {
```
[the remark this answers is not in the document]
haselbach: There is a bad mix of tabs and spaces here. I'm planning to do a follow up CL to make spotless create consistent tab vs spaces usage.

dumper/app/src/main/sh/cloud_extractor/gce_launcher.sh
haselbach: Interesting. The parameter is a literal (not`$`, no `*` in it) but adding single quotes certainly does not hurt.

Hackerpilot: The `*` is near the end of the line. `-migration-tools-*)"`.

haselbach: Ah, ok, I was looking at the other `--name` parameter. Indeed, it would be better to pass the `*` to find. In this specific case it does not matter but it is more brittle if we let the shell expand. I added the single quotes.

paolomorandini (approved):

PR #230, opened 2023-07-07
Decrypt cloud dumper arguments

haselbach:
When a KMS key is provided via the metadata attribute dwh_key,
decrypt the cloud extractor configuration provided via the
metadata attribute dwh_extractor_configuration using that key.

BUG=290102538


haselbach:
This was merged with https://github.com/google/dwh-migration-tools/pull/234

PR #231, opened 2023-07-10
Add a run override to the dumper to pass in arguments

haselbach:
Add a `run(String... args)` override to allow passing in arguments directly. Also add handling of JSON response files to it.

This means that the overloaded `run` method has additional behavior beyond converting argument types which is not optimal and a result of abstraction leakage in `ConnectorArguments`. However, if we want to allow passing in arguments directly, short of a larger refactor, this is a reasonable compromise.


shevek-google (approved):
Yes, this looks like a good simplification, thank you.

PR #235, opened 2023-07-13
add cleanup variable

stankiewicz:
run.sh after some time leaves quite a mess in a bucket. In most cases this can be safely removed as translated artifacts are downloaded. It's up to the user if they want to keep those files or remove them and that's why we should introduce variable


ajwelch4 (approved):
LGTM, thanks!

PR #237, opened 2023-07-19
Set thread pool size to 1 if --assessment flag is used.

dawidxc:
Using thread pool size bigger than 1 causes the dumper to fail for large databases. Reducing it to 1 if the `--assessment` flag is used should improve user experience.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
@@ -744,6 +744,10 @@ public class ConnectorArguments extends DefaultArguments {
 
   public int getThreadPoolSize() {
     if (isAssessment()) {
+      // Dumping in a single thread lowers the probability that the dumper will put too much
+      // pressure on the Redshift cluster, which could cause other issues. The dumper should use
+      // less memory with a single thread only, thus lowering the probability of OOM. Even though
+      // the dumping may take longer, the execution is more robust (less likely to fail).
       return 1;
     }
     return getOptions().valueOf(optionThreadPoolSize);
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
@@ -334,8 +334,9 @@ public class ConnectorArguments extends DefaultArguments {
       parser
           .accepts(
               OPT_THREAD_POOL_SIZE,
-              "Set thread pool size (affects connection pool size). Defaults to "
-                  + OPT_THREAD_POOL_SIZE_DEFAULT)
+              "Set thread pool size (affects connection pool size)."
+                  + " If the --assessment flag is enabled, this option is"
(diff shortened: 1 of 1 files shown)


bdw-g (approved):

shevek-google (commented):
Can we have some information about the failure? Do we have a b/?

shevek-google (changes requested):
I'm going to explicitly CR this for an explanation, comment, or reference, within the code, saying why this code is there, because it's too easy for someone to remove otherwise.

Also, I think this is going to cause a bifurcation in process: assessment needs the flag, translation may end up in a situation where we can't use it, for scale. So can we please discuss this briefly?

Thank you.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -334,8 +334,9 @@ public class ConnectorArguments extends DefaultArguments {
       parser
           .accepts(
               OPT_THREAD_POOL_SIZE,
-              "Set thread pool size (affects connection pool size). Defaults to "
-                  + OPT_THREAD_POOL_SIZE_DEFAULT)
+              "Set thread pool size (affects connection pool size)."
+                  + " If the --assessment flag is enabled, this option is"
+                  + " ignored and thread pool size is set to 1.")
```
Hackerpilot, line 339: Would it make sense to still include the default thread count in this message?

dawidxc: The default was duplicated before, since it is also included by the library. After the change it is still visible.

Before:
```
--thread-pool-size <Integer>                               Set thread pool size (affects connection pool size).     
                                                             Defaults to 32 (default: 32)                           
```

After:
```
--thread-pool-size <Integer>                               Set thread pool size (affects connection pool size). If  
                                                             the --assessment flag is enabled, this option is       
                                                             ignored and thread pool size is set to 1. (default: 32)
```

dawidxc:
Dumping in a single thread lowers the probability that the dumper will put too much pressure on the Redshift cluster, which could cause other issues. Also, I expect dumper to use less memory with a single thread only, thus lowering the probability of OOM. Even though the dumping may take longer, the execution is more robust (less likely to fail).

It would be best to make the default equal to `1` for assessment and `32` otherwise and still make it possible to change the default, but this might introduce unnecessary complexity. I didn't want to change the existing default, so I guarded the change behind the assessment flag.

dawidxc:
I've added the explanation comment in the code.

dawidxc:
> Also, I think this is going to cause a bifurcation in process: assessment needs the flag, translation may end up in a situation where we can't use it, for scale.

The change affects assessment only. Nothing changes for translation, so the default is still 32 and it can be changed to any value. The `--assessment` flag doesn't have to be used for translation. Could you describe the situation (the command-line options used) that could cause trouble?

shevek-google:
Based on conversations, it sounds like an option to set thread pool size may be merited, but I'm not so keen on forcing it to 1 for --assessment and 32 otherwise. It would be nice if the assessment and translation modes were as similar as possible, otherwise we lose the benefit of sharing the codebase.

PR #244, opened 2023-07-24
Add missing columns

ishmum123:


dominik-klimczak (approved):
it'd be better not to fail if there is an extra column, so listing columns instead of * is better
please address that in one of the next PRs

PR #247, opened 2023-07-27
Ignore unexpected columns for assessment

ishmum123:


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
```diff
@@ -82,7 +93,28 @@ protected CSVFormat newCsvFormat(@Nonnull ResultSet rs) throws SQLException {
...
+        HashSet<String> columnNames = new HashSet<>();
+
+        for (int i = 0; i < metaData.getColumnCount(); i++) {
+          columnNames.add(metaData.getColumnName(i));
+        }
+
+        String missingColumns =
+            Arrays.stream(headerClass.getEnumConstants())
+                .map(Enum::name)
+                .map(name -> CaseFormat.UPPER_CAMEL.to(CaseFormat.UPPER_UNDERSCORE, name))
```
cyulysses-corp, line 108: Can be done in the same .map()

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
```diff
@@ -82,7 +93,28 @@ protected CSVFormat newCsvFormat(@Nonnull ResultSet rs) throws SQLException {
...
+                .filter(columnName -> !columnNames.contains(columnName))
```
cyulysses-corp, line 109: optional: this could alternatively be done by using a https://docs.oracle.com/javase/8/docs/api/java/util/Set.html#removeAll

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnector.java
```diff
@@ -149,14 +149,14 @@ protected void addSqlTasksWithInfoSchemaFallback(
   }
 
   @ForOverride
-  protected void addSingleSqlTask(
+  protected void addSingleSqlTaskForAssessment(
```
cyulysses-corp, line 152: Why rename? It looks like it still works for both Assessment and non-Assessment.

ishmum123: This has the assessment flags set to `true`. I didn't want to add yet another parameter to the function.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
ishmum123: I didn't want to add everything to a set and then remove unwanted ones and then `stream` again for joining. wdyt?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
```diff
@@ -82,7 +93,28 @@ protected CSVFormat newCsvFormat(@Nonnull ResultSet rs) throws SQLException {
     Class<? extends Enum<?>> headerClass = getHeaderClass();
     if (headerClass != null) {
       format = format.withHeader(headerClass);
-      if (headerClass.getEnumConstants().length != rs.getMetaData().getColumnCount())
+      ResultSetMetaData metaData = rs.getMetaData();
+      if (isAssessment) {
+
+        HashSet<String> columnNames = new HashSet<>();
+
+        for (int i = 0; i < metaData.getColumnCount(); i++) {
```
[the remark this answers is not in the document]
ishmum123: Thanks a lot for catching this!

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
ishmum123: Seeing the method reference felt better. Are you strongly against this?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
```diff
@@ -61,8 +65,15 @@ public abstract class AbstractJdbcTask<T> extends AbstractTask<T> {
 
   @CheckForNull private Class<? extends Enum<?>> headerClass;
 
+  private final boolean isAssessment;
+
   public AbstractJdbcTask(@Nonnull String targetPath) {
+    this(targetPath, false);
+  }
+
+  public AbstractJdbcTask(@Nonnull String targetPath, boolean isAssessment) {
```
dominik-klimczak, line 74: I think this is actually a case for subclassing (doesn't have to be done in this PR). A factory that has knowledge about parameters could control which class is actually created.

dominik-klimczak (commented):
I think we also need to remove unnecessary columns from the Format enum. I think we can leave these (requires testing):
Name,
Type,
Size,
MinClusterCount,
MaxClusterCount,
AutoSuspend,
AutoResume,
CreatedOn,
ResumedOn,
UpdatedOn,
Owner,
Comment,
ScalingPolicy

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
cyulysses-corp: I don't see the need to join again, because the column difference is just informational, and you can always copy the set. Set operations is usually faster than cross-iterating through both iterators.

IIUC, in current implementation this check is executed for every table-interval, so for a 30d total duration it would be around 24/day-table*30d*1table + 1/day-table*30d*15table = 1170 operations and not increasing terribly fast, so it is not a key point of optimization, and this change is optional.

One definitely nice optimization is to only execute it once for every table, though.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnector.java
cyulysses-corp: Ah so this method was already only used for assessment tasks and not being overriden by any method. If this is the case then we can remove "ForOverride"

cyulysses-corp:
> 

I think they are valuable information for WLM analysis, no?

cyulysses-corp:
> I think we also need to remove unnecessary columns from the Format enum. I think we can leave these (requires testing):...


I think they are valuable information for WLM analysis, no?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
cyulysses-corp: It could simply have been:
.map(enumVal -> CaseFormat.UPPER_CAMEL.to(CaseFormat.UPPER_UNDERSCORE, enumVal.name()))

This way feels much better for me...  but it is optional.

cyulysses-corp (commented):
Please also do a manual run for each of the following cases:
snowflake-metadata, Assessment
snowflake-metadata, non-Assessment
snowflake-logs, Assessment
snowflake-logs, non-Assessment
&
snowflake-metadata, against our non-Enterprise-tier database.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
ishmum123: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
```diff
@@ -82,7 +93,28 @@ protected CSVFormat newCsvFormat(@Nonnull ResultSet rs) throws SQLException {
     Class<? extends Enum<?>> headerClass = getHeaderClass();
     if (headerClass != null) {
       format = format.withHeader(headerClass);
-      if (headerClass.getEnumConstants().length != rs.getMetaData().getColumnCount())
+      ResultSetMetaData metaData = rs.getMetaData();
+      if (isAssessment) {
+
+        HashSet<String> columnNames = new HashSet<>();
```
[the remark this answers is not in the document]
ishmum123: Moved the whole validation process to a new method.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
ishmum123: Adding this to the `TODO` list

ishmum123:
> I think we also need to remove unnecessary columns from the Format enum. I think we can leave these (requires testing): Name, Type, Size, MinClusterCount, MaxClusterCount, AutoSuspend, AutoResume, CreatedOn, ResumedOn, UpdatedOn, Owner, Comment, ScalingPolicy

Let me address this in a separate PR

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
```diff
@@ -76,28 +84,68 @@ public AbstractJdbcTask<T> withHeaderClass(@Nonnull Class<? extends Enum<?>> hea
...
+      desiredColumnNames.removeAll(resultColumnNames);
+
+      if (!desiredColumnNames.isEmpty()) {
+        throw new SQLException(
+            new MetadataDumperUsageException(
+                "Fatal Error. ResultSet does not have the expected columns: "
+                    + desiredColumnNames));
+      }
+    } else if (headerClass.getEnumConstants().length != metaData.getColumnCount())
+      // Can we avoid nesting exceptions here?
```
cyulysses-corp, line 138: nit, optional: directly wrap the error message in SQLException instead of in MetadataDumperUsageException (here and above), then delete this comment.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
```diff
@@ -76,28 +84,68 @@ public AbstractJdbcTask<T> withHeaderClass(@Nonnull Class<? extends Enum<?>> hea
...
+      throw new SQLException(
```
cyulysses-corp, line 139: Aren't we supposed to not throw exception in this case?

ishmum123: oops! my bad. should be okay now

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
ishmum123: Is it okay now?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
```diff
@@ -76,28 +83,65 @@ public AbstractJdbcTask<T> withHeaderClass(@Nonnull Class<? extends Enum<?>> hea
...
+      validateHeader(headerClass, metaData);
     } else {
       format = format.withHeader(rs);
     }
     return format;
   }
 
+  private void validateHeader(Class<? extends Enum<?>> headerClass, ResultSetMetaData metaData)
+      throws SQLException {
+    if (isAssessment) {
```
shevek-google, line 114: Can we please spend a few minutes talking about why this should be an assessment-specific behaviour? I'm keen to keep the dumper as a single product, particularly when we have to load the dumps into two downstream consumers: translation and assessment. Also, we have just received a report of invalid SQL generated by the dumper, causing task failure, and the bifurcation of the SQL generation is making it harder to debug and test.

shevek-google (changes requested):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
ishmum123: Changed the flag to `ignoreExtraColumns`

cyulysses-corp (approved):
LGTM. Please do the test runs with the new flag configuration then. You may need to first remove that header that caused the problem with Snowflake non-enterprise tier warehouses first, and test it against both enterprise and non-enterprise tier warehouses.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
```diff
@@ -120,17 +120,17 @@ private void validateHeader(Class<? extends Enum<?>> headerClass, ResultSetMetaD
         resultColumnNames.add(metaData.getColumnName(i + 1));
       }
 
-      Set<String> desiredColumnNames =
+      Set<String> remainingColumnNames =
           Arrays.stream(headerClass.getEnumConstants())
               .map(
-                  enumVal -> CaseFormat.UPPER_CAMEL.to(CaseFormat.UPPER_UNDERSCORE, enumVal.name()))
+                  enumVal -> CaseFormat.UPPER_CAMEL.to(CaseFormat.LOWER_UNDERSCORE, enumVal.name()))
```
cyulysses-corp, line 126: We can skip the conversion since we compare using equalsIgnoreCase

cyulysses-corp (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
ishmum123: the conversion is mostly for `CamelCase` to `_underscore_` case. The java `equalsIgnoreCase` would be bother about the `underscores`

cyulysses-corp: I see. I suspect it might not work for all database types, then. Some databases fields are not underscored. We should make sure that all database types' cases can be compared with the Upper-Camel enum names.

ishmum123:
@cyulysses-corp can you please if your concern is addressed with the new commit?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnector.java
```diff
@@ -48,6 +50,8 @@ public class SnowflakeMetadataConnector extends AbstractSnowflakeConnector
   @SuppressWarnings("UnusedVariable")
   private static final Logger LOG = LoggerFactory.getLogger(SnowflakeMetadataConnector.class);
 
+  private static final Set<Class<?>> validatedTables = new HashSet<>();
```
shevek-google, line 53: At first review, this looks like a static mutable variable, which makes the dumper itself not be reusable in a thread-safe scenario.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
```diff
@@ -76,28 +83,81 @@ public AbstractJdbcTask<T> withHeaderClass(@Nonnull Class<? extends Enum<?>> hea
...
+              + "If a custom query has been specified please confirm the selected columns match"
+              + " the following: "
+              + StringUtils.join(headerClass.getEnumConstants(), ", "));
+  }
+
+  static Set<String> calculateRemainingColumns(
+      Class<? extends Enum<?>> headerClass, Set<String> resultColumnNames) {
+    Set<String> availableCamelCaseNames =
+        resultColumnNames.stream()
+            .map(name -> CaseFormat.UPPER_CAMEL.to(CaseFormat.LOWER_UNDERSCORE, name))
```
shevek-google, line 143: The variable is called CamelCase but the values appear to be LOWER_UNDERSCORE.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
```diff
@@ -76,28 +83,81 @@ public AbstractJdbcTask<T> withHeaderClass(@Nonnull Class<? extends Enum<?>> hea
...
+            .collect(Collectors.toSet());
+    Set<String> availableUnderscoreCaseNames =
+        resultColumnNames.stream()
+            .map(name -> CaseFormat.UPPER_UNDERSCORE.to(CaseFormat.LOWER_UNDERSCORE, name))
+            .collect(Collectors.toSet());
+
+    Set<String> requiredNames =
+        Arrays.stream(headerClass.getEnumConstants())
+            .map(enumVal -> CaseFormat.UPPER_CAMEL.to(CaseFormat.LOWER_UNDERSCORE, enumVal.name()))
+            .collect(Collectors.toSet());
```
shevek-google, line 153: Why are we manipulating the names? Is there a mismatch between some TD versions? Does that manipulation extend to underscores? Or is this to do with our convention to have enums all have UPPER_UNDERSCORE, which doesn't match the underlying?

shevek-google (changes requested):
Could we please spend a few minutes in a meeting reviewing this? I'd like to talk over motivation and design.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnector.java
```diff
@@ -148,16 +152,18 @@ protected void addSqlTasksWithInfoSchemaFallback(
     }
   }
 
-  @ForOverride
-  protected void addSingleSqlTask(
+  protected void addSingleSqlTaskForAssessment(
       @Nonnull List<? super Task<?>> out,
       @Nonnull Class<? extends Enum<?>> header,
```
shevek-google, line 157: [New review] Per discussion: Suggest to make this CheckForNull, avoid passing it when possible, do **NOT** pass it in any query that looks like a 'select *' (e.g. show warehouses is really a select * from ???), and let's not add the additional complexity.

cyulysses-corp (changes requested):
Please also run test dumps for both snowflake and another dialect type (either Redshift or Teradata)

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnector.java
```diff
@@ -148,16 +155,30 @@ protected void addSqlTasksWithInfoSchemaFallback(
...
+  private String[] transformHeaderToCamelCase(ResultSet rs, CaseFormat baseFormat) {
+    try {
+      ResultSetMetaData metaData = rs.getMetaData();
+      int columnCount = metaData.getColumnCount();
+      String[] columns = new String[columnCount];
+      for (int i = 0; i < columnCount; i++) {
+        columns[i] = baseFormat.to(CaseFormat.UPPER_CAMEL, metaData.getColumnLabel(i + 1));
+      }
+      return columns;
+    } catch (SQLException e) {
```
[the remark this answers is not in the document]
ishmum123: `getColumnCount` and `getColumnLabel` throws the same exception as well.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnector.java
```diff
@@ -148,16 +155,30 @@ protected void addSqlTasksWithInfoSchemaFallback(
...
+  }
+
+  @CheckReturnValue
+  private String[] transformHeaderToCamelCase(ResultSet rs, CaseFormat baseFormat) {
+    try {
+      ResultSetMetaData metaData = rs.getMetaData();
+      int columnCount = metaData.getColumnCount();
+      String[] columns = new String[columnCount];
+      for (int i = 0; i < columnCount; i++) {
+        columns[i] = baseFormat.to(CaseFormat.UPPER_CAMEL, metaData.getColumnLabel(i + 1));
```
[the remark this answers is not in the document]
ishmum123: The conversion itself is just `String` manipulation which is very unlikely to fail. The `try-catch` here is for `SQLException`. In case that occurs, we will not able to return the original column names as well.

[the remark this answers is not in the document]
ishmum123: ref: https://github.com/apache/commons-csv/blob/9bb8da934eb3843947977b63bcaf716335179ad2/src/main/java/org/apache/commons/csv/CSVFormat.java#L487

ishmum123:
> Please also run test dumps for both snowflake and another dialect type (either Redshift or Teradata)

Done!

cyulysses-corp (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnector.java
shevek-google, line 179: I don't particularly like catching and discarding exceptions, so I went on an adventure:

HeaderTransformer is called in a place where SQLException may be thrown.
So what if we define a custom functional interface with `String[] apply(ResultSet) throws SQLException` and use that instead of `Function<ResultSet, String[]>`? Then this try/catch can be removed entirely, and this method can just throw.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnector.java
```diff
@@ -148,16 +155,30 @@ protected void addSqlTasksWithInfoSchemaFallback(
...
+      @Nonnull TaskVariant task,
+      @Nonnull Function<ResultSet, String[]> transformer) {
     out.add(
         new JdbcSelectTask(
                 task.zipEntryName, String.format(format, task.schemaName, task.whereClause))
-            .withHeaderClass(header));
+            .withHeaderTransformer(transformer));
+  }
+
+  @CheckReturnValue
```
shevek-google, line 169: Did this mean @CheckForNull? And did that go away once we make the exception/interface change?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
```diff
@@ -92,8 +101,10 @@ protected CSVFormat newCsvFormat(@Nonnull ResultSet rs) throws SQLException {
                     "If a custom query has been specified please confirm the selected columns match"
                         + " the following: ",
                     StringUtils.join(headerClass.getEnumConstants(), ", "))));
-    } else {
+    } else if (headerTransformer == null || headerTransformer.apply(rs) == null) {
       format = format.withHeader(rs);
```
shevek-google, line 105: This is interesting: If we make the assumption that we're going to be error tolerant in headerTransformer, we should have a try/catch here, in which case it may be that the failure-case for headerTransformer is not just to use withHeader(rs) but something else? Because what caused the failure of headerTransformer might be rs.getColumnName(), which will just throw again in withHEader(rs).

So how about:
```
else if (headerTransformer != null) {
   try { // Do we actually want this try, or not?
       format = format.withHeader(headerTransformer.apply(rs));
    } catch (SQLException e) { /// something? Or should we just let this propagate }
}
else {
    format = format.withHeader(rs);
}

shevek-google (approved):
I'm approving this with one requested minor change, and one observation.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnector.java
ishmum123: This feels much cleaner. Thanks!

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnector.java
ishmum123: Yup, it went away with the new design

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
ishmum123: I am voting for letting the `SQLException` just propagate especially because of the reason you outlined

PR #249, opened 2023-07-27
Add memoization to the driver retriever

haselbach:
The cloud extractor can do multiple dumps and will likely require the same driver for all runs (e.g. metadata and logs for the same DB). Thus, the driver for it only needs to be downloaded once.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/clouddumper/DriverRetriever.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/clouddumper/DriverRetriever.java
@@ -16,6 +16,8 @@
  */
 package com.google.edwmigration.dumper.application.dumper.clouddumper;
 
+import static com.google.edwmigration.dumper.application.dumper.utils.MemoizationUtil.createMemoizer;
+
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
@@ -25,11 +27,11 @@ import com.google.common.io.BaseEncoding;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.net.URI;
-import java.net.URISyntaxException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.Optional;
+import org.apache.commons.io.function.IOFunction;
 import org.apache.hc.client5.http.impl.classic.CloseableHttpClient;
 import org.apache.hc.core5.http.ClassicHttpRequest;
 import org.apache.hc.core5.http.HttpStatus;
@@ -48,6 +50,7 @@ public class DriverRetriever {
   private final CloseableHttpClient httpClient;
   private final Path driverPath;
   private final ImmutableMap<String, DriverInformation> driverInformationMap;
+  private final IOFunction<DriverInformation, Path> downloader = createMemoizer(this::download);
 
   @VisibleForTesting
   DriverRetriever(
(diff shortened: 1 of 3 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/clouddumper/DriverRetriever.java
```diff
@@ -144,15 +141,18 @@ public Optional<Path> getDriver(String name) throws IOException {
       LOG.info("Got no driver for connector '{}'.", name);
       return Optional.empty();
     }
-    DriverInformation driverInformation = driverInformationMap.get(name);
+    return this.downloader.apply(driverInformationMap.get(name));
+  }
+
+  private Optional<Path> download(DriverInformation driverInformation) throws IOException {
```
dawidxc, line 147: When does `download` return empty Optional?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/MemoizationUtil.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Function;
+import org.apache.commons.io.function.IOFunction;
+
+/** Utility class for memoization. */
+public class MemoizationUtil {
+
+  private static class OptionalMemoizer<T, K, R> implements IOFunction<T, Optional<R>> {
```
dawidxc, line 29: IIUC the `OptionalMemoizer` class is needed, so that retries are possible. It seems that these two concerns (caching and retrying) are mixed in this solution. I would suggest separating them by using a `LoadingCache` from guava, so that retrying could be performed independently (internally in the loading function).

https://github.com/google/guava/wiki/CachesExplained

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/clouddumper/DriverRetriever.java
haselbach: When there is no driver needed for the connector.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/MemoizationUtil.java
haselbach: No, retries are handled by the Apache Commons Http Client. If all retries fail, an IOException is raised.

This is just to say that the delegate should be checked again if the optional is empty. It is not strictly necessary here as the information does not change on the delegate but there is also no expensive IO happening.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/clouddumper/DriverRetriever.java
dawidxc: The `getDriver` method can return empty Optional, but I don't see how the `download` method could.

If I'm not mistaken, by looking at the code I think that the `download` method executes HTTP request always, so the method should always return non-empty Optional (or throw an exception). If this is the behavior, then can the return type be changed from `Optional<Path>` to `Path`?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/MemoizationUtil.java
dawidxc: The PR description is: "The cloud extractor can do multiple dumps and will likely require the same driver for all runs (e.g. metadata and logs for the same DB). Thus, the driver for it only needs to be downloaded once."

If the download is unsuccessful (IOException happens), then the process will stop.

If the download is successful, then the driver path will be saved in the map and there won't be a need to call the delegate the second time.

> (...) delegate should be checked again if the optional is empty.

Could you point to the code, where the second check would happen (or if the code is not ready yet, then describe such a use-case)?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/clouddumper/DriverRetriever.java
haselbach: You are right. I initially wanted to have the memoization at an earlier point. Here, it does not make sense to have the memoization wrapped in `Optional` as it will always be populated. I've adapted it accordingly.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/MemoizationUtil.java
```diff
@@ -0,0 +1,63 @@
...
+      }
+      R result = delegate.apply(t);
+      cache.put(key, result);
+      return result;
+    }
+  }
+
+  /**
+   * Creates a memoizer that stores results in a map using the key created by the key function. When
+   * called with a value v, the map is checked if there already is an entry for keyFuncion(v). If
```
dawidxc, line 53: Please, fix typo in `keyFunction`

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/MemoizationUtil.java
```diff
@@ -0,0 +1,63 @@
...
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Function;
+import org.apache.commons.io.function.IOFunction;
+
+/** Utility class for memoization. */
+public class MemoizationUtil {
+
+  private static class Memoizer<T, K, R> implements IOFunction<T, R> {
```
dawidxc, line 28: I would recommend using a standard caching mechanism, e.g. guava `LoadingCache` with mapping the key to the `DriverInformation` in the loading function instead of creating a custom solution (`Memoizer` class).

The aliases could be moved from `driverInformationMap` to a new `aliasesMap`.

dawidxc (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/MemoizationUtil.java
haselbach: I don't understand. As I said, the Optional is not necessary here and hence I removed it. What are we debating now?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/MemoizationUtil.java
haselbach: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/MemoizationUtil.java
haselbach: I considered using `LoadingCache` but it is not a good fit. It is not really a cache and hence I don't need any of the eviction features. Handling of `IOException`s gets more complicated with `LoadCache`.

I don't strictly need the key handling I implemented (since the `DriverInformation` instances are all the same object for all aliases) but it is still nice that we can be sure that we point to the same URI.

The heavy lifting here is really done by `HashMap`. In that sense, what I need is a hash map, not a cache and `LoadingCache` would just save me from doing line 41-47 in `MemoizationUtil`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/MemoizationUtil.java
haselbach: Are you wondering about why the retriever might be called a 2nd time? That is because there can be multiple connectors in the cloud extractor configuration. See https://github.com/google/dwh-migration-tools/blob/main/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/clouddumper/Main.java#L79

The use case is that you want to download metadata and query logs in one go.

dawidxc: 2023-07-28 10:07 GMT you wrote:
> This is just to say that the delegate should be checked again if the optional is empty.

2023-07-28 10:26 GMT I wrote:
> Could you point to the code, where the second check would happen (or if the code is not ready yet, then describe such a use-case)?

2023-07-28 11:38 GMT you force-pushed the removal of Optional from Memoizer:
https://github.com/google/dwh-migration-tools/compare/9246c978283a982341ef5622cc60f9d9b6991a4d..c64143fc325b7bc7f3e427b103678bd449971d38

2023-07-29 10:52 GMT you wrote:
I don't understand. As I said, the Optional is not necessary here and hence I removed it.

---

I just wanted to point out that IIUC the delegate will never be called a second time for the same driver. You removed the Optional from the Memoizer after my comment, so the case you described as "the delegate should be checked again if the optional is empty" won't happen, so it's all good now.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/MemoizationUtil.java
dawidxc: > It is not really a cache and hence I don't need any of the eviction features.

From Guava manual:
"you should consider using caches when a value is expensive to compute or retrieve, and you will need its value on a certain input more than once. (...) In some cases a `LoadingCache` can be useful even if it doesn't evict entries, due to its automatic cache loading."
https://github.com/google/guava/wiki/CachesExplained#applicability

> I don't strictly need the key handling I implemented (since the `DriverInformation` instances are all the same object for all aliases) but it is still nice that we can be sure that we point to the same URI.

Yes, I would slightly suggest to either map URIs to the Path of the downloaded driver, or use the whole `DriverInformation` as a key. In both approaches the `Memoizer` could be simplified to require 2 generic parameters instead of 3.

> `LoadingCache` would just save me from doing line 41-47 in `MemoizationUtil`

This is my main point - to reuse existing code that was written for this exact purpose (unless you think using `LoadingCache` is an overkill in this case).

haselbach: The issue with `IOException` remains. The handling of it in `LoadingCache` introduces more code than what `MemoizationUtil` does. Otherwise, I would use `LoadingCache`.

haselbach: To explore this a bit further I created https://github.com/google/dwh-migration-tools/compare/feature-cache-dependency-downloads...feature-cache-dependency-downloads-with-loading-cache?expand=1

As you can see, the exception handling adds more code. But not just that, imagine debugging an issue: You know have to understand `LoadingCache`. Most Java developers have a decent understanding of maps but will not know the details of Guava's caching. So, by using `LoadingCache` it will be harder to reason about what is happening.

dawidxc: The code:
```
public static <T, R> IOFunction<T, R> createMemoizer(IOFunction<T, R> delegate) {
  LoadingCache<T, R> cache =
      CacheBuilder.newBuilder()
          .build(
              new CacheLoader<T, R>() {
                public R load(T t) throws IOException {
                  return delegate.apply(t);
                }
              });
  return t -> {
    try {
      return cache.get(t);
    } catch (UncheckedExecutionException e) {
      if (e instanceof RuntimeException) {
        throw (RuntimeException) e.getCause();
      } else {
        throw new RuntimeException("Got unexpected UncheckedExecutionException.", e);
      }
    } catch (ExecutionException e) {
      Throwable cause = e.getCause();
      if (cause instanceof IOException) {
        throw (IOException) cause;
      } else {
        throw new RuntimeException("Got unexpected ExecutionException.", e);
      }
    }
  };
}
```

could be simplified to:

```
public static <T, R> Function<T, R> memoize(Function<T, R> delegate) {
  LoadingCache<T, R> cache = CacheBuilder.newBuilder().build(CacheLoader.from(delegate));
  return t -> {
    try {
      return cache.get(t);
    } catch (ExecutionException e) {
      throw new RuntimeException("Got unexpected ExecutionException.", e);
    }
  };
}
```
The unwrapping of some exceptions, yet keeping others wrapped, might not bring a lot of simplification compared to the shorter version. However, the shorter version would require handling of `IOException` either in the Memoizer or inside of delegate (e.g. converting it to `UncheckedIOException`). Exceptions and lambdas require choosing some tradeoff.

If you want to use `IOFunction` and `IOException`, then it is probably not possible to write a much simpler code using `LoadingCache` than the exploration you've prepared. In this case using `HashMap` makes more sense.

PR #254, opened 2023-07-31
Add license option

haselbach:
Allow users to display license information via the command line option `--licenses`.


shevek-google (changes requested):
Could you possibly point me to the term that says we need this, as opposed to just putting the licenses into the distribution? This definitely applies to Android apps where the running app is the user's only access to the package, but do we need this here, where the licenses are already packaged under `docs/licenses` in a more granular format?

haselbach:
I'm moving the license review internals discussion to b/293934867. Please see there.

shevek-google:
Is this PR still outstanding, or are we happy with the filesystem situation so we can close it?

PR #258, opened 2023-08-01
 Validate `-D` command-line options.

dawidxc:
The regular command-line options are validated by the joptsimple library, but the definition options (`-Dkey=value`) accepts any keys without validation, so typos are difficult to debug.

This change adds validation for such definition options.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
@@ -16,13 +16,21 @@
  */
 package com.google.edwmigration.dumper.application.dumper;
 
+import static autovalue.shaded.com.google.common.collect.ImmutableMap.toImmutableMap;
 import static com.google.common.collect.ImmutableList.toImmutableList;
+import static com.google.common.collect.Maps.immutableEntry;
+import static java.util.Arrays.stream;
 
+import com.google.auto.value.AutoValue;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.MoreObjects.ToStringHelper;
+import com.google.common.base.Preconditions;
 import com.google.common.base.Predicates;
 import com.google.common.base.Strings;
 import com.google.common.collect.ComparisonChain;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.ImmutableSetMultimap;
 import com.google.edwmigration.dumper.application.dumper.annotations.RespectsInput;
 import com.google.edwmigration.dumper.application.dumper.connector.Connector;
 import com.google.edwmigration.dumper.application.dumper.connector.ConnectorProperty;
@@ -372,7 +380,7 @@ public class ConnectorArguments extends DefaultArguments {
           .withRequiredArg()
           .ofType(String.class)
           .describedAs("define");
-  private Map<String, String> definitionMap = null;
+  private Map<String, String> definitionMap;
 
   // because of quoting of special characeters on command line... the -password is made optional,
   // and if not given, asked at the console.
@@ -380,6 +388,10 @@ public class ConnectorArguments extends DefaultArguments {
 
   public ConnectorArguments(@Nonnull String... args) throws IOException {
     super(args);
+  }
+
+  public void initialize() {
+    searchDefinitions();
   }
 
   private static class InputDescriptor implements Comparable<InputDescriptor> {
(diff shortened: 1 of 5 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -816,21 +825,67 @@ public String getDefinition(@Nonnull ConnectorProperty property) {
...
+                          () ->
+                              new IllegalArgumentException(
+                                  String.format(
+                                      "Command-line option '-D%s' requires a value. Use: -D%s=value",
+                                      propertyName, propertyName)));
+                })
+            .collect(toImmutableMap(Map.Entry::getKey, Map.Entry::getValue));
+  }
+
+  private static ImmutableSetMultimap<String, String> createPropertyNameToConnectorsMapping() {
```
shevek-google, line 878: Should we use CONNECTORS from MetadataDumper rather than re-invoking ServiceLoader?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -382,6 +387,10 @@ public ConnectorArguments(@Nonnull String... args) throws IOException {
     super(args);
   }
 
+  public void initialize() {
+    searchDefinitions();
+  }
```
shevek-google, line 392: This exposes additional contract from the object. Should this be done either in the constructor, or in parseArguments()?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -816,21 +825,67 @@ public String getDefinition(@Nonnull ConnectorProperty property) {
...
+                      .map(value -> immutableEntry(propertyName, value))
+                      .orElseThrow(
+                          () ->
+                              new IllegalArgumentException(
+                                  String.format(
+                                      "Command-line option '-D%s' requires a value. Use: -D%s=value",
+                                      propertyName, propertyName)));
+                })
+            .collect(toImmutableMap(Map.Entry::getKey, Map.Entry::getValue));
+  }
```
shevek-google, line 876: I'm a little concerned with quite how large this routine became. Would it be more readable as a loop with if() statements in it? We have a requirement that this code be readable and auditable by end users as open source.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -816,21 +825,67 @@ public String getDefinition(@Nonnull ConnectorProperty property) {
...
+                      connectorName,
+                      compatibleConnectors);
+                  return propertyValue
+                      .map(value -> immutableEntry(propertyName, value))
+                      .orElseThrow(
+                          () ->
+                              new IllegalArgumentException(
+                                  String.format(
+                                      "Command-line option '-D%s' requires a value. Use: -D%s=value",
+                                      propertyName, propertyName)));
```
shevek-google, line 873: Is this a deferred exception which is always thrown if the value is empty? In which case, could we have done something above like:
```
if (idx < 0)
    throw new IAE(...)
```

instead of storing the Optional in the ImmutableMap and then using orElseThrow as a functional form here?

shevek-google (changes requested):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
dawidxc: I need the new contract in order to fail fast if the command-line options are invalid. Doing work in constructor is an anti-pattern, so I would like to avoid this.

As for `parseArguments` - I haven't found such an identifier:
```
$ grep -r parseArguments | wc -l
0
```

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
dawidxc: I do not want `ConnectorArguments` to depend on `MetadataDumper`. However, reusing `CONNECTORS` makes sense. I've moved them to a new class `ConnectorsHolder`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
dawidxc: I've extracted the body of the loop to a separate `parse` method.

ishmum123 (approved):

PR #264, opened 2023-08-03
Use database predicate instead of schema predicate

frsv:
[Language manual](https://cwiki.apache.org/confluence/display/hive/languagemanual+ddl) explicitly states `The uses of SCHEMA and DATABASE are interchangeable – they mean the same thing`. Assessment uses "database" term. The current option potentially brings a confusion in the public documentation.

Schema option is only used in Hive and Oracle connectors (where it means, well, [a user](https://www.janbasktraining.com/community/sql-server/how-is-oracle-schema-different-from-database-or-user#:~:text=In%20Oracle%20schemas%20and%20users,that%20belong%20to%20that%20account.) ), while Database option is used in Teradata and Netezza connectors specifically to filter by database name, so the PR makes Hive connector the same.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnector.java
@@ -20,7 +20,7 @@ import com.google.auto.service.AutoService;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Preconditions;
 import com.google.edwmigration.dumper.application.dumper.ConnectorArguments;
-import com.google.edwmigration.dumper.application.dumper.annotations.RespectsArgumentSchemaPredicate;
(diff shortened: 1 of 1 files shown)


shevek-google (commented):
The reason for the difference is that Netezza and Teradata don't have schemas, but other databases do. Actually, NZ does, but it doesn't use them. Hive has them both interchangeably. PostgreSQL and Redshift schema, but not (usefully) database. So we should look not for consistency across all databases here, we should look for the closest match to any source database's own language, and the expectations of its users. Do we argue that Hive users **expect** to use the word 'schema' or 'database'?

frsv:
@shevek-google, thanks for the context. Yes, I would say we should use `database`:

1) I asked PSO and CE, according to them `database` is more common and natural
2) From my experience, I've almost never seen `schema` used over the internet
3) According to google search results and [trends](https://trends.google.com/trends/explore?date=all&q=hive%20schema,hive%20database&hl=en), `database` is more common

shevek-google (approved):

PR #265, opened 2023-08-03
Add LogDate to the JOIN condition

dawidxc:
LogDate is part of Partitioned Primary Index of both tables, so it needs to be part of the JOIN condition.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsJdbcTask.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsJdbcTask.java
@@ -190,7 +190,12 @@ public class TeradataLogsJdbcTask extends AbstractJdbcTask<Summary> {
       // "QueryID is a system-wide unique field; you can use QueryID
       // to join DBQL tables ... without needing ProcID as an additional join field."
       // (https://docs.teradata.com/reader/B7Lgdw6r3719WUyiCSJcgw/YIKoBz~QQgv2Aw5dF339kA)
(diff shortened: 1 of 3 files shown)


ishmum123 (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsJdbcTask.java
```diff
@@ -190,7 +190,12 @@ private String getSql(@Nonnull JdbcHandle handle) {
       // "QueryID is a system-wide unique field; you can use QueryID
       // to join DBQL tables ... without needing ProcID as an additional join field."
       // (https://docs.teradata.com/reader/B7Lgdw6r3719WUyiCSJcgw/YIKoBz~QQgv2Aw5dF339kA)
-      buf.append(" LEFT OUTER JOIN ").append(queryTable).append(" ST ON L.QueryID=ST.QueryID");
+      buf.append(" LEFT OUTER JOIN ").append(queryTable).append(" ST ON (L.QueryID=ST.QueryID");
+
+      if (logDateColumn != null) {
+        buf.append(" AND L.").append(logDateColumn).append("=ST.").append(logDateColumn);
```
paolomorandini, line 196: Discussed offline the interaction between this and the condition at line 216 which could make the join an inner join.

dawidxc: I've removed the condition on `ST.LogDate` column from the `WHERE` clause, but as suggested by another person, this will not make a difference, because it will still be treated as an inner join.

PR #266, opened 2023-08-03
Bring in runtime dependencies for redshift

haselbach:
Bring in explicit runtime dependencies for redshift needed for IAM credentials access. Without them, using `iam-accesskeyid` / `iam-secretaccesskey` instead of `password` fails with `ClassNotFoundException`s.

BUG=294435345

### DIFF
--- a/dumper/app/build.gradle
+++ b/dumper/app/build.gradle
@@ -70,6 +70,7 @@ dependencies {
     runtimeOnly "org.postgresql:postgresql"
     runtimeOnly "net.snowflake:snowflake-jdbc"
     runtimeOnly "com.amazon.redshift:redshift-jdbc42"
+    runtimeOnly "com.amazonaws:aws-java-sdk-redshift"
 
     testFixturesApi testFixtures(project(':dumper:lib-common'))
 
@@ -94,7 +95,7 @@ dependencies {
 }
 
 startScripts {
-    classpath = files( '$APP_HOME/lib/*' )
+    classpath = files('$APP_HOME/lib/*')
 }
 
 application {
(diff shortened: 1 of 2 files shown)


dumper/app/build.gradle
```diff
@@ -70,6 +70,8 @@ dependencies {
     runtimeOnly "org.postgresql:postgresql"
     runtimeOnly "net.snowflake:snowflake-jdbc"
     runtimeOnly "com.amazon.redshift:redshift-jdbc42"
+    runtimeOnly "com.amazonaws:aws-java-sdk-core:1.12.521"
+    runtimeOnly "com.amazonaws:aws-java-sdk-redshift:1.12.520"
```
shevek-google, line 74: 1. Do we need both of those, or does sdk-redshift bring in sdk-core transitively?
2. Can we specify the version in the conventions file, so the version is used consistently, globally, and refer only to the artefact name here?

dumper/app/build.gradle
```diff
@@ -113,12 +115,12 @@ task cloudExtractorStartScripts(type: CreateStartScripts) {
 licenseReport {
     // onlyIf { ! gradle.startParameter.offline }
     filters = [
-        new com.github.jk1.license.filter.LicenseBundleNormalizer(bundlePath: rootProject.file("gradle/license-bundle-normalizer.json"), createDefaultTransformationRules: true)
+            new com.github.jk1.license.filter.LicenseBundleNormalizer(bundlePath: rootProject.file("gradle/license-bundle-normalizer.json"), createDefaultTransformationRules: true)
```
shevek-google, line 118: It looks from the PR as if this whitespace change changed both the spaces-to-tabs (I assume) but also the indentation level, which is now apparently inconsistent (4 vs 8). Is that so?

shevek-google (approved):
Approved with minor change requests.

dumper/app/build.gradle
haselbach: 1. Core can be removed as sdk-redshift depends on it. I removed it.
2. Sure. Done.

shevek-google (approved):

dumper/app/build.gradle
haselbach: That was intellijs doing. As far as I can tell, intellij applies groovy rules and uses the continuation indentation rules for list elements. After setting this to 4, it formatted it as it was before. I also increased column width to 120 to avoid the line breaks it introduced.

PR #274, opened 2023-08-08
Call thrift directly instead of using the plugin

haselbach:
Remove the dependency on the gradle thrift plugin in favor of calling thrift directly.

Furthermore, check in the generated source code so that we can avoid generating the Java files from thrift as these rarely need to be regenerated and thrift is sometimes cumbersome to install.

This creates the task `generateThrift` which deletes the subfolder `src/main/gen-java` and then runs `thrift` for each thrift file. The generated code is written to `src/main/gen-java`.


dumper/lib-ext-hive-metastore/build.gradle
```diff
@@ -78,15 +74,30 @@ shadowJar {
...
+    delete project.projectDir.toPath().resolve('src/main/gen-java')
+}
+
+task generateThrift {
+    dependsOn cleanThrift
+}
+
+task thrift {
+    dependsOn cleanThrift, generateThrift
+}
```
shevek-google, line 87: Suggestion: Gradle really likes being idempotent, so the 'delete-then-rebuild' strategy we're used to from make doesn't really work in Gradle. The trick is to use unions and syncs.

So consider a pattern where each Thrift task writes into a separate src directory.

It might be simpler to write a single thrift task which generates all thrift files into a single directory.

To avoid the Delete/ task, generate the thrift java into a directory in project.buildDir and then Sync it into src/ - the sync is a way of doing a delete+union-copy while still being idempotent.

shevek-google (changes requested):
This is fundamentally good, but I'm a little nervous about the task setup, because there's no guarantee that gradle will run all the tasks in the chain. It's slightly better to structure things as idempotent, so temp-dir+sync, and to use the sync to merge the output files into a single dir.

If this isn't clear (I'm not sure I wrote this review very well) please grab a few minutes and we'll walk through it together.

dumper/lib-ext-hive-metastore/build.gradle
haselbach: I've done the changes as discussed.

dumper/lib-ext-hive-metastore/build.gradle
```diff
@@ -78,15 +75,25 @@ shadowJar {
...
+                workingDir project.projectDir.toPath().resolve('src/main')
+                commandLine 'thrift', '-r', '--gen', 'java', thriftPath
+            }
         }
     }
 }
 
 sourceSets.main.java {
-    srcDir file('build/generated-sources/thrift/main')
+    srcDir file('src/main/gen-java')
```
shevek-google, line 98: I think it might be possible (preferable) to say `srcDir generateThrift` here to say "Consume the output of that task, whatever it does."

dumper/lib-ext-hive-metastore/build.gradle
```diff
@@ -78,15 +75,25 @@ shadowJar {
 
 }
 
-generateThriftSource {
-    generators {
-        java {
-            out file('build/generated-sources/thrift/main')
-            // option 'java5'
+task generateThrift {
```
shevek-google, line 78: It's possible to say `def generateThrift = tasks.register('generateThrift') { ... }` instead of `task generateThrift { }` and the build will be slightly faster - only matters for large builds, because that does lazy configuration.

shevek-google (approved):
Approved, wonderful. I left two comments, but you're welcome to just ignore/resolve both as FYI/educational, as neither technically represents an issue.

dumper/lib-ext-hive-metastore/build.gradle
shevek-google: Beautiful.

PR #277, opened 2023-08-11
Update from gradle 6 to 7

haselbach:


shevek-google (approved):
If, after you merge this, you want to step up to 8, that wouldn't be a bad thing either.

PR #279, opened 2023-08-15
Update to gradle8

haselbach:


dumper/app/build.gradle
```diff
@@ -179,4 +179,8 @@ distributions {
             }
         }
     }
+}
+
+['publishedDistZip', 'publishedDistTar', 'installPublishedDist'].collect(tasks::named).each {
+    it.configure { dependsOn('checkLicense') }
```
shevek-google, line 185: This is very Googly code. 😝 

As a comparative luddite, I might have written:
```
['publishedDistZip', 'publishedDistTar', 'installPublishedDist'].each {
    tasks.named(it).configure { dependsOn('checkLicense') }
}
```

But I haven't tested that, as I'm the peanut gallery.

shevek-google (approved):

dumper/app/build.gradle
haselbach: Groovy probably creates a new list or array with `collect` (unlike streams) but with this small amount of elements it does not matter. I forgot why I chose that route. Something annoyed me with the other approach but probably something unimportant. Anyway, I don't think it matters one way or another.

PR #295, opened 2023-10-16
Log progress of dumper

ishmum123:


dawidxc (approved):

dumper/app/src/main/resources/logback.xml
```diff
@@ -7,6 +7,13 @@
 		</encoder>
 	</appender>
 
+	<appender name="FILE" class="ch.qos.logback.core.FileAppender">
+		<file>dumper.log</file>
```
dawidxc, line 11: I would suggest to include a timestamp in the filename of the progress log file, so that different executions don't overwrite the same file.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
```diff
@@ -279,7 +279,7 @@ public <T> T runChildTask(Task<T> task) throws MetadataDumperUsageException {
         TASK:
         for (int i = 0; i < tasks.size(); i++) {
           runTask(runContext, state, tasks.get(i));
-          log("Finished " + (i + 1) + " out of " + tasks.size() + " tasks");
+          LOG.info("Finished " + (i + 1) + " out of " + tasks.size() + " tasks");
```
dawidxc, line 282: I would suggest to use a separate slf4j logger to log progress, so that it does not get mixed up with other logs. Appenders are by default cumulative, but you can override this with `additivity` attribute.

https://logback.qos.ch/manual/configuration.html#overridingCumulativity

dumper/app/src/main/resources/logback.xml
ishmum123: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
ishmum123: Done

dawidxc (approved):

PR #299, opened 2023-10-18
Add Cloudera connector task for dumping services

kravikumar:
Adds a Cloudera connector task for dumping services and service yarn metrics. The two are combined into one task to avoid redundancy

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/ClouderaServicesTask.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/ClouderaServicesTask.java
@@ -16,41 +16,46 @@ import java.nio.charset.StandardCharsets;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 
-public class ClouderaServicesTask extends AbstractClouderaTask{
+public class ClouderaServicesTask extends AbstractClouderaTask {
 
+  public ClouderaServicesTask() {
+    super("services.json");
+  }
 
-    public ClouderaServicesTask() {
-        super("services.json");
-    }
+  // It would be more optimal to combine all metrics that fall under a common parent into a task, so
+  // we avoid repeated requests to the same API. For example, HDFS usage can be combined under this
+  // task since it pretty much calls the same APIs.
+  @CheckForNull
+  @Override
+  protected Void doRun(TaskRunContext context, @Nonnull ByteSink sink, @Nonnull Handle handle)
+      throws Exception {
+    ClouderaHandle h = (ClouderaHandle) handle;
+    ApiClusterList clusters = getClusters(h);
+    ServicesResourceApi api = new ServicesResourceApi(h.getClient());
 
-    // It would be more optimal to combine all metrics that fall under a common parent into a task, so we avoid repeated requests to the same API.
-    // For example HDFS usage can be combined under this task since it pretty much calls the same APIs.
-    @CheckForNull
-    @Override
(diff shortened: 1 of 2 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/ClouderaServicesTask.java
```diff
@@ -0,0 +1,61 @@
...
+    ImmutableList.Builder yarnMetricsBuilder = ImmutableList.builder();
+
+    // TODO: Accept startDate as an input. Arbitrarily setting it to today minus 7
+    String startDate = LocalDateTime.now().minusDays(7).format(DateTimeFormatter.ISO_DATE);
+    try (Writer writer = sink.asCharSink(StandardCharsets.UTF_8).openBufferedStream()) {
+      for (ApiCluster cluster : clusters.getItems()) {
+        ApiServiceList services = api.readServices(cluster.getName(), null);
+        for (ApiService service : services.getItems()) {
+          // Includes name and health of each service
+          servicesBuilder.add(service);
```
shevek-google, line 46: This adds the service (which is part of the cluster) to the overall list. Does the service cotnain the cluster name, or should we be wrapping the service itself in an outer bean which contains the cluster name?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/ClouderaServicesTask.java
```diff
@@ -0,0 +1,61 @@
...
+          ApiYarnUtilization yarnUtilization =
+              api.getYarnUtilization(
+                  cluster.getName(), service.getName(), null, null, startDate, null, null, null);
+          yarnMetricsBuilder.add(yarnUtilization);
+        }
+      }
+      ImmutableMap result =
+          ImmutableMap.of(
+              "services", servicesBuilder.build(), "yarnMetrics", yarnMetricsBuilder.build());
+      CoreMetadataDumpFormat.MAPPER.writeValue(writer, result);
```
shevek-google, line 57: Do we want to dump the whole thing as one big JSON object, or do we want to write this in JSONL format? The advantage of JSONL is that we can read it line-by-line, which lets us read a large dump line-wise, rather than having to allocate memory. I hit the same question in the HDFS usage task, where I didn't write the data, but I got as far as the writing line-by-line question.

shevek-google (approved):
Since this is a pull into my branch, and my branch isn't ready for master yet, I approve this for merge into my branch, but there are questions below which we should cover before we merge to master.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/ClouderaServicesTask.java
kravikumar: The service has a reference to the cluster via ApiClusterRef. I think one way could be to a task that reads clusters, services, reports, metrics etc and dumps them while preserving the hierarchy of clusters->services and clusters->hosts.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/ClouderaServicesTask.java
kravikumar: We could do JSONL if that's how we are dumping in the other connectors and if it makes for easy consumption. Let's discuss these offline before we merge to master.

PR #300, opened 2023-10-25
Add a shorter error at the end of the output

dawidxc:
In case of the exception that causes the dumper to fail, its stack trace is printed to the output at the end of the output before the dumper exits. Reading stack traces is easy for Java developers, but they contain too many details for the regular user, who is only interested in the error message.

I am extracting the error messages from the exception and its causal chain and printing them without the stack trace at the end of the output to make it easier for the user to see the error faster.

The original error message with stack trace is still preserved and it is printed before the final short error message.

Sample output:
```
* ERROR
* Failed to initialize pool: FATAL: password authentication failed for user "joe"
* Caused by: FATAL: password authentication failed for user "joe"
```

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/Main.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/Main.java
@@ -16,6 +16,10 @@
  */
 package com.google.edwmigration.dumper.application.dumper;
 
+import static com.google.common.collect.ImmutableList.toImmutableList;
+
+import com.google.common.base.Throwables;
+import com.google.common.collect.ImmutableList;
 import javax.annotation.Nonnull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/Main.java
```diff
@@ -46,7 +55,12 @@ public static void main(String... args) throws Exception {
       main.run(args);
     } catch (MetadataDumperUsageException e) {
       LOG.error(e.getMessage());
-      for (String msg : e.getMessages()) LOG.error(msg);
+      for (String msg : e.getMessages()) {
+        LOG.error(msg);
+      }
+    } catch (Exception e) {
+      e.printStackTrace();
+      System.out.println("\nERROR\n" + extractExceptionMessages(e));
```
ishmum123, line 63: Can we box this with `STARS` so that the user knows where to focus on?

ishmum123 (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/Main.java
dawidxc: Done. The new output looks like this:

```
* ERROR
* Failed to initialize pool: FATAL: password authentication failed for user "joe"
* Caused by: FATAL: password authentication failed for user "joe"
```

PR #303, opened 2023-10-31
Add `SqlPiiRemover`

ishmum123:
This will be used to remove `PII` from all dialects


arotenberg-google (changes requested):
Design discussion pending.

PR #310, opened 2023-11-21
Add alternative location for Teradata resource usage tables

dawidxc:
Add two command-line options to provide alternative location of Teradata resource usage tables: ResUsageScpu and ResUsageSpma. The options are:

```
-Dteradata-logs.res-usage-scpu-table
-Dteradata-logs.res-usage-spma-table
```

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsConnector.java
@@ -20,6 +20,7 @@ import static com.google.common.collect.ImmutableList.toImmutableList;
 
 import com.google.auto.service.AutoService;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.edwmigration.dumper.application.dumper.ConnectorArguments;
 import com.google.edwmigration.dumper.application.dumper.MetadataDumperUsageException;
 import com.google.edwmigration.dumper.application.dumper.annotations.RespectsArgumentAssessment;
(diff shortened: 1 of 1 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsConnector.java
```diff
@@ -109,16 +119,17 @@ public TeradataLogsConnector() {
...
-    return ImmutableList.of("ResUsageScpu", "ResUsageSpma").stream()
+  private ImmutableList<TeradataJdbcSelectTask> createTimeSeriesTasks(
+      ZonedInterval interval, @Nonnull ConnectorArguments arguments) {
+    return ImmutableList.of(RES_USAGE_SCPU_TABLE, RES_USAGE_SPMA_TABLE).stream()
         .map(
-            tableName ->
+            table ->
                 new TeradataJdbcSelectTask(
-                    createFilename("dbc." + tableName + "_", interval),
+                    createFilename(table.getDefaultValue() + "_", interval),
```
ishmum123, line 128: Should we call the variable `defaultValue`? Creating the file-name with `defaultValue` even after we allow change seems a bit counter-intuitive

ishmum123 (approved):
Do we still want to create

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsConnector.java
dawidxc: We want the filename to have the same prefix that signifies that type of data within the file. Either the data is in the `ResUsageScpu` format, in which case the filename will have prefix `dbc.ResUsageScpu_`, or it is in `ResUsageSpma` format, in which case it will be `dbc.ResUsageSpma_`. In a sense the prefix is always the default value, but the source table can be changed.

I think the current code expresses this idea. If you think the code is not clear enough, could you, please, suggest how to change it?

ishmum123 (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsConnector.java
ishmum123: LGTM

PR #311, opened 2023-11-22
Add text splitting of the Teradata TableTextV table

dawidxc:
The `RequestText` column of the `dbc.TableTextV` table may need splitting due to 64kB row size limit on Teradata 15. This change introduces a new command-line option that allows for splitting this column into multiple rows if the length is longer than the specified max length.

By default, if the option is not used the column is extracted as is, which may reach 32000 characters. With the option, the column is split into multiple rows. The max length can be any integer between 5000 and 30000 characters.

Example usage:
```
-Dteradata.metadata.max-text-length=7000
```


ishmum123 (approved):
ditto: https://github.com/google/dwh-migration-tools/pull/312#discussion_r1403229816

PR #312, opened 2023-11-23
Add splitting of long SQL queries in DBQLSqlTbl

dawidxc:
The `SqlTextInfo` column of the `dbc.DBQLSqlTbl` table may need splitting due to 64kB row size limit on Teradata 15. This change introduces a new command-line option that allows for splitting this column into multiple rows if the length is longer than the specified max length.

By default, if the option is not used the column is extracted as is, which may reach 31000 characters. With the option, the column is split into multiple rows. The max length can be any integer between 5000 and 30000 characters.

Example usage:

```
-Dteradata-logs.max-sql-length=7000
```

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsJdbcTask.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsJdbcTask.java
@@ -16,13 +16,17 @@
  */
 package com.google.edwmigration.dumper.application.dumper.connector.teradata;
 
+import static com.google.edwmigration.dumper.application.dumper.connector.teradata.TeradataLogsConnector.DBQLSQLTBL_SQLTEXTINFO_LENGTH;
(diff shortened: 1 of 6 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsJdbcTask.java
```diff
@@ -48,6 +52,8 @@
 
 public class TeradataLogsJdbcTask extends AbstractJdbcTask<Summary> {
 
+  private static final int DBQLSQLTBL_SQLTEXTINFO_LENGTH = 31000;
```
ishmum123, line 55: nit: I would prefer if this value is included in the range accepted from the user

ishmum123 (approved):
nit: I would prefer having the accepted range to be (5000 -

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsJdbcTask.java
dawidxc: Done.

PR #321, opened 2023-11-28
Use query builder in DiskSpaceV query generation

dawidxc:


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/MetadataQueryGenerator.java
```diff
@@ -110,21 +111,38 @@ static String createSelectForTableTextV(
...
+      String tableName,
+      Optional<Expression> condition,
+      String orderByColumnDesc) {
+    FromClauseStepBuilder builder =
+        rowCount.isPresent()
+            ? selectTop(rowCount.getAsLong(), projection(star())).from(tableName)
+            : selectAll().from(tableName);
+    AfterWhereStepBuilder builder2 = condition.map(builder::where).orElse(builder);
+    return rowCount.isPresent()
+        ? builder2.orderBy(OrderBySpec.create(identifier(orderByColumnDesc), Direction.DESC))
```
kpiotrowski, line 145: Why orderBy is used if rowCount is provided only? I guess order doesn't matter when we do not limit number of returned records, but this logic is misleading. rowCount is OptionalLong and orderByColumnDesc is a String so I would expect I can provide order column with no rowCount.

kpiotrowski (changes requested):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/MetadataQueryGenerator.java
kpiotrowski: or if it is an expected behavior can we add test case for it?

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/MetadataQueryGeneratorTest.java
```diff
@@ -253,4 +254,54 @@ public void createSelectForTableTextV_withTextMaxLengthAndCondition() {
             + "   WHERE DatabaseName = 'sample_db'",
         query);
   }
+
+  @Test
+  public void createSelectForDiskSpaceV_success() {
```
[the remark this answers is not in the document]
dawidxc: Done.

kpiotrowski (approved):

PR #329, opened 2023-12-21
Bump Snowflake to version 3.14.4 as it resolves Arrow-related bugs.

cyulysses-corp:
https://docs.snowflake.com/en/release-notes/clients-drivers/jdbc-2023

Using ./gradlew versionCatalogUpdate gives the same result for snowflake-jdbc, but we will update the rest (more than 20) dependencies after holiday season.

Verified that the desired snowflake-jdcb version was used in the build.


dawidxc (approved):

kpiotrowski (approved):

cyulysses-corp:
Tested using both `snowflake` and `snowflake-logs` connectors and they work now.

PR #330, opened 2023-12-22
Return exit status 1 if extraction failed

dawidxc:
If any exception was not caught in the main thread before reaching the `main` method, it is now caught there and processed, returning exit status 1. The `System.exit(1)` invocation is moved from `MetadataDumper` class to `Main` class.

The original behavior in terms of exit status for required/optional tasks is preserved:
If any of the required tasks fail, the process returns exit status 1. Otherwise, if no required task fails, then the process returns exit status 0 (no matter if any of the optional tasks failed or not).

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/Main.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/Main.java
@@ -35,8 +35,8 @@ public class Main {
     this.metadataDumper = metadataDumper;
   }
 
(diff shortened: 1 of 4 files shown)


cyulysses-corp (approved):

shevek-google (commented):
This suggests that if ANY task fails, the dumper exits with 1. I can see the point, but since we are free to throw tasks into the dumper and it can tolerate failures, are we risking having 'exit 1' be "just a normal occurrence"? I encourage the group to think outside the scope of the teradata log dumper for this question.

PR #332, opened 2023-12-28
Extend Redshift metadata connector with cluster properties.

kpiotrowski:
- Extend Redshift metadata connector with cluster properties (node type, nodes number, storage) fetched from the AWS API.
- Additional file `cluster_nodes.csv` is included in the Redshift metadata ZIP file with the following columns:
`cluster_identifier,host,nodes_num,node_type,total_storage`
- Use existing AWS flags to configure credentials. (AWS profile or static IAM credentials).

### DIFF
--- a/dumper/app/build.gradle
+++ b/dumper/app/build.gradle
@@ -60,6 +60,7 @@ dependencies {
     implementation libs.google.cloud.nio
     implementation libs.google.cloud.kms
     implementation libs.httpclient5
+    implementation libs.aws.java.sdk.redshift
 
     runtimeOnly libs.logback.classic
     runtimeOnly libs.jcl.over.slf4j
@@ -67,7 +68,6 @@ dependencies {
     runtimeOnly libs.postgresql
     runtimeOnly libs.snowflake.jdbc
     runtimeOnly libs.redshift.jdbc
-    runtimeOnly libs.aws.java.sdk.redshift
 
     testFixturesApi testFixtures(project(':dumper:lib-common'))
 
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/AbstractAwsApiTask.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/AbstractAwsApiTask.java
@@ -16,6 +16,8 @@
  */
 package com.google.edwmigration.dumper.application.dumper.connector.redshift;
 
+import static com.google.common.collect.ImmutableList.toImmutableList;
+
 import com.amazonaws.auth.AWSCredentialsProvider;
 import com.amazonaws.auth.AWSStaticCredentialsProvider;
 import com.amazonaws.auth.BasicAWSCredentials;
@@ -27,7 +29,6 @@ import com.google.edwmigration.dumper.application.dumper.ConnectorArguments;
 import com.google.edwmigration.dumper.application.dumper.task.AbstractTask;
 import com.google.edwmigration.dumper.plugin.ext.jdk.progress.RecordProgressMonitor;
 import java.io.IOException;
-import java.io.UncheckedIOException;
 import java.io.Writer;
 import java.nio.charset.StandardCharsets;
 import java.util.Optional;
@@ -70,15 +71,10 @@ public abstract class AbstractAwsApiTask extends AbstractTask<Void> {
         Writer writer = sink.asCharSink(StandardCharsets.UTF_8).openBufferedStream()) {
       CSVPrinter printer = format.print(writer);
 
-      records.forEach(
-          record -> {
-            try {
-              monitor.count();
-              printer.printRecord(record);
-            } catch (IOException e) {
-              throw new UncheckedIOException(e);
-            }
-          });
+      for (Object[] record : records.collect(toImmutableList())) {
+        monitor.count();
+        printer.printRecord(record);
+      }
     }
     return null;
   }
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/AbstractAwsApiTask.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/AbstractAwsApiTask.java
@@ -35,7 +35,7 @@ import org.apache.commons.csv.CSVPrinter;
 public abstract class AbstractAwsApiTask extends AbstractTask<Void> {
 
   Class<? extends Enum<?>> headerEnum;
-  Optional<AmazonRedshift> redshifClient;
+  Optional<AmazonRedshift> redshiftClient;
 
   public AbstractAwsApiTask(String zipEntryName, Class<? extends Enum<?>> headerEnum) {
     super(zipEntryName);
@@ -43,13 +43,13 @@ public abstract class AbstractAwsApiTask extends AbstractTask<Void> {
   }
 
   @Nonnull
-  public AbstractAwsApiTask withRedshiftApiClient(AmazonRedshift redshifClient) {
-    this.redshifClient = Optional.of(redshifClient);
+  public AbstractAwsApiTask withRedshiftApiClient(AmazonRedshift redshiftClient) {
+    this.redshiftClient = Optional.of(redshiftClient);
     return this;
   }
 
   public AmazonRedshift redshiftApiClient() {
-    return redshifClient.or(AmazonRedshiftClientBuilder.standard().build());
+    return redshiftClient.or(AmazonRedshiftClientBuilder.standard().build());
   }
 
   public Void writeRecordsCsv(ByteSink sink, Stream<Object[]> records) throws IOException {
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/AbstractAwsApiTask.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/AbstractAwsApiTask.java
@@ -16,9 +16,14 @@
  */
 package com.google.edwmigration.dumper.application.dumper.connector.redshift;
 
+import com.amazonaws.auth.AWSCredentialsProvider;
+import com.amazonaws.auth.AWSStaticCredentialsProvider;
+import com.amazonaws.auth.BasicAWSCredentials;
+import com.amazonaws.auth.profile.ProfileCredentialsProvider;
 import com.amazonaws.services.redshift.AmazonRedshift;
-import com.amazonaws.services.redshift.AmazonRedshiftClientBuilder;
+import com.amazonaws.services.redshift.AmazonRedshiftClient;
 import com.google.common.io.ByteSink;
+import com.google.edwmigration.dumper.application.dumper.ConnectorArguments;
 import com.google.edwmigration.dumper.application.dumper.task.AbstractTask;
 import com.google.edwmigration.dumper.plugin.ext.jdk.progress.RecordProgressMonitor;
 import java.io.IOException;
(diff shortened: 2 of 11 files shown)


dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/task/AbstractTaskTest.java
```diff
@@ -30,7 +30,7 @@ protected static class DummyHandle extends AbstractHandle {}
 
   protected static Handle HANDLE = new DummyHandle();
 
-  protected static class MemoryByteSink extends ByteSink {
+  public static class MemoryByteSink extends ByteSink {
```
cyulysses-corp, line 33: An alternative to this is to let RedshiftClusterNodesTaskTest extend AbstractTaskTest . WDYT?

dumper/app/build.gradle
```diff
@@ -60,14 +60,14 @@ dependencies {
     implementation libs.google.cloud.nio
     implementation libs.google.cloud.kms
     implementation libs.httpclient5
+    implementation libs.aws.java.sdk.redshift
```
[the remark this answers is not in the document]
kpiotrowski: It's because dependency declared with `implementation` is available in compile time and runtime. We now use this library in the dumper code so this needs to be available in the compile time.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/task/AbstractTaskTest.java
kpiotrowski: good idea, other test classes already extend AbstractTaskTest so this solution should be fine.

kpiotrowski: done, inner class has to stay public because RedshiftClusterNodesTaskTest is outside of task package, but this allows to get rid off MemoryByteSink import.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/AbstractAwsApiTask.java
```diff
@@ -0,0 +1,74 @@
...
+        Writer writer = sink.asCharSink(StandardCharsets.UTF_8).openBufferedStream()) {
+      CSVPrinter printer = format.print(writer);
+
+      records.forEach(
+          record -> {
+            try {
+              monitor.count();
+              printer.printRecord(record);
+            } catch (IOException e) {
+              throw new UncheckedIOException(e);
```
shevek-google, line 68: If we do this with a regular for-loop rather than a forEach/Closure, we don't need the try/catch, and this entire code block gets shorter, simpler, and more reliable, because we don't throw an unchecked exception.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterNodesTask.java
```diff
@@ -0,0 +1,56 @@
...
+    DescribeClustersResult result = client.describeClusters(request);
+
+    return writeRecordsCsv(
+        sink,
+        result.getClusters().stream()
+            .map(
+                cluster ->
+                    new Object[] {
+                      cluster.getClusterIdentifier(),
+                      cluster.getEndpoint().getAddress(),
```
shevek-google, line 50: How sure are we that endpoint is ALWAYS nonnull? I suggest 'not very'.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftMetadataConnector.java
```diff
@@ -64,6 +93,10 @@ public void addTasksTo(List<? super Task<?>> out, ConnectorArguments arguments)
     out.add(new DumpMetadataTask(arguments, FORMAT_NAME));
     out.add(new FormatTask(FORMAT_NAME));
     out.add(new RedshiftEnvironmentYamlTask());
+    // AWS API tasks
+    if (arguments.getDefinition(RedshiftMetadataConnectorProperties.AWS_API) != null) {
+      out.add(new RedshiftClusterNodesTask());
```
shevek-google, line 98: Why don't we do this by default, if we have the right credentials?

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/task/AbstractTaskTest.java
```diff
@@ -30,7 +30,7 @@ protected static class DummyHandle extends AbstractHandle {}
 
   protected static Handle HANDLE = new DummyHandle();
 
-  protected static class MemoryByteSink extends ByteSink {
+  public class MemoryByteSink extends ByteSink {
```
shevek-google, line 33: Why did we lose 'static'?

shevek-google (dismissed):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterNodesTask.java
kpiotrowski: good point, This is not documented that endpoint is nonnull, so better to check that.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftMetadataConnector.java
kpiotrowski: I changed the credentials flow based on your suggestions.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/AbstractAwsApiTask.java
```diff
@@ -0,0 +1,95 @@
...
+    this.redshiftClient = Optional.of(redshiftClient);
+    return this;
+  }
+
+  public AmazonRedshift redshiftApiClient() {
+    return redshiftClient.orElseGet(
+        () -> AmazonRedshiftClient.builder().withCredentials(credentialsProvider).build());
+  }
+
+  public Void writeRecordsCsv(@Nonnull ByteSink sink, Stream<Object[]> records) throws IOException {
```
dawidxc, line 68: I would suggest avoiding using `Stream` as a parameter. In this case, you could accept `ImmutableList`, since you already collect the stream to the list inside the implementation of this method.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/AbstractAwsApiTask.java
```diff
@@ -0,0 +1,95 @@
...
+        monitor.count();
+        printer.printRecord(record);
+      }
+    }
+    return null;
+  }
+
+  public static Optional<AWSCredentialsProvider> createCredentialsProvider(
+      ConnectorArguments arguments) {
+    if (arguments.getIAMProfile() != null) {
```
dawidxc, line 84: The calls to the methods in `arguments` are repeated. The code could be simplified to:
```
return Optional.ofNullable(arguments.getIAMProfile())
  .map(ProfileCredentialsProvider::new)
  .or(() -> 
    Optional.ofNullable(arguments.getIAMAccessKeyID())
      .flatMap(accessKeyId ->
        Optional.ofNullable(arguments.getIAMSecretAccessKey())
          .map(secretAccessKey ->
            new AWSStaticCredentialsProvider(
              new BasicAWSCredentials(accessKeyId, secretAccessKey)))));
```

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftMetadataConnector.java
```diff
@@ -65,6 +67,13 @@ public void addTasksTo(List<? super Task<?>> out, ConnectorArguments arguments)
     out.add(new FormatTask(FORMAT_NAME));
     out.add(new RedshiftEnvironmentYamlTask());
 
+    // AWS API tasks, enabled by default if IAM credentials are provided
+    Optional<AWSCredentialsProvider> awsCredentials =
+        AbstractAwsApiTask.createCredentialsProvider(arguments);
+    if (awsCredentials.isPresent()) {
```
dawidxc, line 73: I would recommend avoiding using `isPresent` followed by `get`. In this case, you can use `ifPresent`:

```
AbstractAwsApiTask.createCredentialsProvider(arguments)
  .ifPresent(awsCredentials ->
    out.add(new RedshiftClusterNodesTask(awsCredentials));
```

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/AbstractAwsApiTask.java
cyulysses-corp: A slight modification of dawidxc's suggestion - no need to immutable-qualify the input. List<Object[]> should suffice.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/AbstractAwsApiTask.java
kpiotrowski: What do you think about simply assigning arguments to local variables? You code looks nice but I think it's more clear with simple if statements. In the future we may need to extend it with additional Providers. Adding next if statement is easy, chaining Optionals may be confusing for the reader.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/AbstractAwsApiTask.java
```diff
@@ -0,0 +1,101 @@
...
+    this.redshiftClient = Optional.of(redshiftClient);
+    return this;
+  }
+
+  public AmazonRedshift redshiftApiClient() {
+    return redshiftClient.orElseGet(
+        () -> AmazonRedshiftClient.builder().withCredentials(credentialsProvider).build());
+  }
+
+  public Void writeRecordsCsv(@Nonnull ByteSink sink, List<Object[]> records) throws IOException {
```
dawidxc, line 69: This method could return `void`, because there is no point in passing its result in the `doRun` method. It's only the `doRun` method in `RedshiftClusterNodesTask` class that needs to adhere to the `AbstractTask#doRun` API.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/AbstractAwsApiTask.java
dawidxc: Using local variables with ifs also works. Possibly, the second part of the method could be further simplified:

```
return optionalIf(iamAccessKey != null && iamSecretAccessKey != null,
    () -> new AWSStaticCredentialsProvider(
        new BasicAWSCredentials(iamAccessKey, iamSecretAccessKey)));
```

dawidxc (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/AbstractAwsApiTask.java
kpiotrowski: I implemented similar idea inline because `optionalIf` method is defined in TeradataUtils class and we shouldn't import it. Later we can move it to some shared utils package.

cyulysses-corp (approved):

PR #333, opened 2023-12-29
Snowflake database flag fix

cyulysses-corp:
1. Un-require database flag for Snowflake because it is not a necessary parameter. Still allowing providing database by the user, and will do database access check in this case.
2. Actually use the provided database to check, instead of using the wrong "current_databaase()" method.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/AbstractSnowflakeConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/AbstractSnowflakeConnector.java
@@ -99,8 +99,8 @@ public abstract class AbstractSnowflakeConnector extends AbstractJdbcConnector {
   private void checkCurrentDatabaseExists(
       @Nonnull String databaseName, @Nonnull JdbcTemplate jdbcTemplate)
       throws MetadataDumperUsageException {
-    String currentDatabase = jdbcTemplate.queryForObject("USE DATABASE ?", String.class,
-        databaseName);
+    String currentDatabase = jdbcTemplate.queryForObject(
+        String.format("USE DATABASE %s", databaseName), String.class);
     if (currentDatabase == null) {
       List<String> dbNames =
           jdbcTemplate.query("SHOW DATABASES", (rs, rowNum) -> rs.getString("name"));
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/AbstractSnowflakeConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/AbstractSnowflakeConnector.java
@@ -65,6 +65,7 @@ public abstract class AbstractSnowflakeConnector extends AbstractJdbcConnector {
   }
 
   private static final int MAX_DATABASE_CHAR_LENGTH = 255;
+  private static final String DEFAULT_DATABASE = "SNOWFLAKE";
 
   @Nonnull
   @Override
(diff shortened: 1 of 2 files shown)


kpiotrowski (changes requested):
There are some format issues please run `./gradlew :dumper:app:spotlessApply` to fix it.

cyulysses-corp:
> There are some format issues please run `./gradlew :dumper:app:spotlessApply` to fix it.

Done.

kpiotrowski (approved):
LGTM 👍

shevek-google (dismissed):
This smells of a gateway to SQL injection. Are we sure we want to do this?

cyulysses-corp:
> This smells of a gateway to SQL injection. Are we sure we want to do this?

This app is meant to be used by the user who also has sufficient access to the database already, and this step does not bypass the credential validation. i.e. it is not the use cases where doing a direct SQL query to the database is harder than embedding it in a SQL injection attack.

If there are other use cases that you planned this app for but I do not know, and you think they are at risk, then I will happily put some SQL injection protection here.

kpiotrowski:
> > This smells of a gateway to SQL injection. Are we sure we want to do this?
> 
> This app is meant to be used by the user who also has sufficient access to the database already, and this step does not bypass the credential validation. i.e. it is not the use cases where doing a direct SQL query to the database is harder than embedding it in a SQL injection attack.
> 
> If there are other use cases that you planned this app for but I do not know, and you think they are at risk, then I will happily put some SQL injection protection here.

I tried to run Dumper with invalid database name and it fails at very beginning with credentials validation error. 
image

so the SQL injection is not possible here because of validation, but best practice would be avoid it by design. Can you provide databaseName as a parameter to the query? sth like:

```
String sql = "USE DATABASE ?";
return jdbcTemplate.queryForObject(sql, new Object[]{databaseName}, String.class);
```

cyulysses-corp:
> > > This smells of a gateway to SQL injection. Are we sure we want to do this?
> > 
> > 
> > This app is meant to be used by the user who also has sufficient access to the database already, and this step does not bypass the credential validation. i.e. it is not the use cases where doing a direct SQL query to the database is harder than embedding it in a SQL injection attack.
> > If there are other use cases that you planned this app for but I do not know, and you think they are at risk, then I will happily put some SQL injection protection here.
> 
> I tried to run Dumper with invalid database name and it fails at very beginning with credentials validation error. image
> 
> so the SQL injection is not possible here because of validation, but best practice would be avoid it by design. Can you provide databaseName as a parameter to the query? sth like:
> 
> ```
> String sql = "USE DATABASE ?";
> return jdbcTemplate.queryForObject(sql, new Object[]{databaseName}, String.class);
> ```

Thanks for weighing in. Actually I initially tried with the exact same way you suggested, but had to change to using explicit String.format for the sql query instead of jdbcTemplate(sql, requiredType, args) because args are not binded consistently. For example, this is the error message I got from that implementation:
```
* ERROR
* PreparedStatementCallback; bad SQL grammar [USE DATABASE ?]; nested exception is net.snowflake.client.jdbc.SnowflakeSQLException: SQL compilation error:
syntax error line 1 at position 13 unexpected '?'.
* Caused by: SQL compilation error:
syntax error line 1 at position 13 unexpected '?'.
```

I can implement custom SQL inspection to restrict the string inputs allowed here for sure, but I just was not sure about the necessity of it.

shevek-google:
Just because _we_ aren't (currently) deploying it in an injection-sensitive situation doesn't mean the customer isn't. But we should talk internally about the future.

However, given the proof that a ? isn't legal there, let's keep the quotes, sanitize the name to make sure it doesn't contain a quote, and then this is probably the best we can do. CharMatcher.anyMatch?

cyulysses-corp:
> Just because _we_ aren't (currently) deploying it in an injection-sensitive situation doesn't mean the customer isn't. But we should talk internally about the future.
> 
> However, given the proof that a ? isn't legal there, let's keep the quotes, sanitize the name to make sure it doesn't contain a quote, and then this is probably the best we can do. CharMatcher.anyMatch?

It turns out that a bad database name would already cause failure during the creation of the JdbcHandle, example:
```
    Jan 05, 2024 4:10:11 PM net.snowflake.client.jdbc.RestRequest execute
    SEVERE: Error response: HTTP Response code: 400, request: POST [https://compilerworks.snowflakecomputing.com:443/session/v1/login-request?databaseName=db12345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890&warehouse=testwh&roleName=tester&requestId=1ea7f4f8-bf3e-4999-b58d-db0c5e56d760](https://compilerworks.snowflakecomputing.com/session/v1/login-request?databaseName=db12345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890&warehouse=testwh&roleName=tester&requestId=1ea7f4f8-bf3e-4999-b58d-db0c5e56d760) HTTP/1.1
    Jan 05, 2024 4:10:11 PM net.snowflake.client.core.HttpUtil executeRequestInternal
    SEVERE: Error executing request: POST [https://compilerworks.snowflakecomputing.com:443/session/v1/login-request?databaseName=db12345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890&warehouse=testwh&roleName=tester&requestId=1ea7f4f8-bf3e-4999-b58d-db0c5e56d760](https://compilerworks.snowflakecomputing.com/session/v1/login-request?databaseName=db12345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890&warehouse=testwh&roleName=tester&requestId=1ea7f4f8-bf3e-4999-b58d-db0c5e56d760) HTTP/1.1
    Jan 05, 2024 4:10:11 PM net.snowflake.client.jdbc.SnowflakeUtil logResponseDetails
    SEVERE: Response status line reason: Bad Request
    Jan 05, 2024 4:10:11 PM net.snowflake.client.jdbc.SnowflakeUtil logResponseDetails
    SEVERE: Response content: Not a valid SQL Identifier.
```

Note that So the "USE DATABASE" will be redundant.

Snowflake driver also has intrinsic parsing that prevents trickery, but only because it parses it as a url.
```
    Jan 05, 2024 4:10:11 PM net.snowflake.client.jdbc.SnowflakeConnectString parse
    WARNING: Exception thrown while parsing Snowflake connect string. Illegal character in url.
```
It is actually kinda sad because no double quotes or unicode name is allowed.

I think the best way is for us to simply ignore --database flag for Snowflake connectors. i.e. we don't append "db=" with the database name even if the user specifies it.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/AbstractSnowflakeConnector.java
```diff
@@ -88,24 +88,28 @@ public Handle open(@Nonnull ConnectorArguments arguments) throws Exception {
         newDriver(arguments.getDriverPaths(), "net.snowflake.client.jdbc.SnowflakeDriver");
     DataSource dataSource =
         new SimpleDriverDataSource(driver, url, arguments.getUser(), arguments.getPassword());
-    return checkCurrentDatabaseExists(arguments, new JdbcHandle(dataSource));
+    JdbcHandle jdbcHandle = new JdbcHandle(dataSource);
+    if (!arguments.getDatabases().isEmpty()) {
+      checkCurrentDatabaseExists(arguments.getDatabases().get(0), jdbcHandle.getJdbcTemplate());
```
dawidxc, line 93: The method name `checkCurrentDatabaseExists` starts with `check`, even though it issues `USE DATABASE` that alters the JDBC session (i.e. it's not just a check). I would suggest to either rename the method or call 2 methods - one for checking that the database exists, another for setting the current database for the session.

dawidxc (approved):

cyulysses-corp:
> This smells of a gateway to SQL injection. Are we sure we want to do this?

PTAL. Thanks.

PR #334, opened 2024-01-03
Speed up unit tests

dawidxc:
The diagnostic query generation takes a long time, while it is not necessary for unit tests. The query was executed only because of failed connections to the non-existent database, which were attempted from the real connector. Replacing the real connector with the test connector eliminates the need to connect to the real database, so the tasks are no longer failing.

The execution time decreased from 31s to 7s.

Command used for time measurement:
```
time ./gradlew cleanTest :dumper:app:test --no-build-cache
```

### DIFF
--- a/dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/MetadataDumperTest.java
+++ b/dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/MetadataDumperTest.java
@@ -20,7 +20,7 @@ import static org.junit.Assert.assertTrue;
 
 import com.google.edwmigration.dumper.application.dumper.connector.Connector;
 import com.google.edwmigration.dumper.application.dumper.connector.bigquery.BigQueryLogsConnector;
-import com.google.edwmigration.dumper.application.dumper.connector.hive.HiveMetadataConnector;
+import com.google.edwmigration.dumper.application.dumper.connector.test.TestConnector;
 import java.io.File;
 import java.io.IOException;
 import org.apache.commons.io.FileUtils;
@@ -37,8 +37,8 @@ public class MetadataDumperTest {
   // TODO(ishmum): `testOverridesZipWithGivenName` with content check
   private File file;
   private Main dumper = new Main(new MetadataDumper());
-  private final Connector connector = new HiveMetadataConnector();
-  private final String defaultFileName = "dwh-migration-hiveql-metadata.zip";
+  private final Connector connector = new TestConnector();
+  private final String defaultFileName = "dwh-migration-test-metadata.zip";
 
   @After
   public void tearDown() throws IOException {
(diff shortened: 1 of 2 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/TasksRunner.java
```diff
@@ -152,21 +155,26 @@ private <T> T runTask(Task<T> task) throws MetadataDumperUsageException {
...
-      }
+      createExceptionFile(task, e);
     }
     return null;
   }
 
+  private void createExceptionFile(Task<?> task, Exception e) {
+    try {
+      ImmutableList.Builder<String> lines = ImmutableList.builder();
+      lines.add(task.toString()).add("******************************");
```
ishmum123, line 166: Do we need the stars if `runDiagnosticQueryError` is `false`?

dawidxc: We don't need the stars. However, I wanted to omit the diagnostic query only, leaving everything else as is, so as to affect the dumper execution as little as possible.

shevek-google (dismissed):
Can we please find out why it's slow before we make this change? This seems to add an extra code path to the codebase which isn't in service of the customer. Are we happy with not testing the path that the customer _will_ follow?

dawidxc:
@shevek-google 

> Can we please find out why it's slow before we make this change?

The expression `new DumperDiagnosticQuery(e).call()` takes a long time (1.5-2 seconds) and it is evaluated multiple times during unit test execution. It's likely slow because of the diagnostic queries underneath. Finding out why would require additional analysis and even then there would be a risk that the process needs to take this much time (i.e. it is possible that I would find out that I am not be able to speed it up).

> Are we happy with not testing the path that the customer will follow?

Good point. I've added a single test of the path that the customer will follow, so that the diagnostic query execution is covered. The test suite now runs in 14s instead of 7s, but it's still twice as fast compared to the original 31s.

As for the rest of the tests, I would suggest to keep them on the extra path to take advantage of the test execution speedup. The omitted code (i.e. creation of the exception file) in these tests is unrelated to what is being tested and it only causes the tests to take more time, while providing no value.

shevek-google:
@dawidxc Can you attach visualvm to the test suite, and enable hotspot capture?

Also, why do we have so many failed tasks in test?

shevek-google:
Could also do a microbenchmark on DumperDiagnosticQuery. Although be careful with cutting that down - every single class in there was written very deliberately, with a purpose, mostly debugging Hadoop. 😁

dawidxc:
@shevek-google 

> Also, why do we have so many failed tasks in test?
The tasks fail, because the dumper tries to connect to the non-existent database. Thank you for this question, because based on it I was able to prepare another solution - I've replaced the real connector with a test one. This allows the tests to execute the command-line options logic and create the necessary files based on the `--output` option without making any requests to the database. Since there are no failed tasks, the diagnostic query is not executed and the tests run fast (benchmarking is no longer necessary).

kpiotrowski (approved):

PR #335, opened 2024-01-08
Extend Redshift raw logs with CPU and Storage metrics from AWS CloudWatch API

kpiotrowski:
- Add `com.amazonaws:aws-java-sdk-cloudwatch` library.
- Extend Redshift raw logs with Cluster usage metrics (Average CPU, Average Disk space used)
- Create new files `cluster_usage_metrics_{DATE}.csv` in the `dwh-migration-redshift-raw-logs*` ZIP file.
- Automatically determine metric period (60 / 300 / 3600 sec) based on interval time.

This PR depends on: https://github.com/google/dwh-migration-tools/pull/332.


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterUsageMetricsTask.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    return metricsMap.keySet().stream()
+        .map(
+            metricDate -> {
+              return Stream.of(new Object[] {metricDate}, metricsMap.get(metricDate))
+                  .flatMap(Stream::of)
+                  .toArray();
+            })
+        .sorted((r1, r2) -> ((String) r1[0]).compareTo((String) r2[0]));
```
dawidxc, line 179: I would suggest making `metricsMap` into a `SortedMap`, because then there would be no need for this extra step to sort the rows.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftRawLogsConnector.java
```diff
@@ -233,4 +250,29 @@ private void makeTasks(
...
+  private void makeClusterMetricsTasks(
+      ConnectorArguments arguments,
+      ZonedIntervalIterable intervals,
+      String filePrefix,
+      Class<? extends Enum<?>> headerEnum,
+      List<MetricConfig> metrics,
+      List<? super Task<?>> out) {
+    Optional<AWSCredentialsProvider> awsCredentials =
+        AbstractAwsApiTask.createCredentialsProvider(arguments);
+    if (awsCredentials.isEmpty()) {
```
dawidxc, line 264: Optional: I would suggest avoiding `isEmpty/get` on `Optional`.

```
AbstractAwsApiTask.createCredentialsProvider(arguments).ifPresent(awsCredentials -> {
  ...
});
```

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterUsageMetricsTask.java
```diff
@@ -0,0 +1,196 @@
...
+   */
+  private int metricDataPeriod() {
+    ZonedDateTime start = interval.getStartUTC();
+    if (start.isAfter(currentTime.minusDays(14))) {
+      return 60;
+    }
+    if (start.isAfter(currentTime.minusDays(62))) {
+      return 300;
+    }
+    return 3600;
```
dawidxc, line 194: What happens if the start time is older than 455 days (e.g. 1000 days old)? Does the API return only the latest data or does it fail?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterUsageMetricsTask.java
```diff
@@ -0,0 +1,196 @@
...
+                  .toArray();
+            })
+        .sorted((r1, r2) -> ((String) r1[0]).compareTo((String) r2[0]));
+  }
+
+  /**
+   * Returns available metric period based on the interval time.
+   * https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_concepts.html
+   */
+  private int metricDataPeriod() {
```
dawidxc, line 186: I would suggest changing the return type to `Duration` and convert to seconds only later, when passing the value to the external API.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterUsageMetricsTask.java
```diff
@@ -0,0 +1,196 @@
...
+    ZonedDateTime start = interval.getStartUTC();
+    if (start.isAfter(currentTime.minusDays(14))) {
+      return 60;
```
dawidxc, line 189: I would suggest changing the `60` literal without a unit to `Duration.ofMinute(1)`. Similarly below.

dawidxc (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftRawLogsConnector.java
```diff
@@ -233,4 +250,29 @@ private void makeTasks(
...
+      return;
+    }
+
+    for (ZonedInterval interval : intervals) {
+      String file =
+          filePrefix
+              + DateTimeFormatter.ISO_OFFSET_DATE_TIME.format(interval.getStartUTC())
```
dawidxc, line 271: I would strongly suggest changing the format of the datetime to the one without colons, because they cause trouble during extraction from zip files in some environments.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterUsageMetricsTask.java
```diff
@@ -0,0 +1,196 @@
...
+                cluster -> {
+                  return zipMetricStreams(
+                          metrics.stream()
+                              .map(
+                                  metricConfig -> {
+                                    return getMetricDataPoints(
+                                        cluster.getClusterIdentifier(), metricConfig);
+                                  })
+                              .collect(toImmutableList()))
+                      .map(
```
cyulysses-corp, line 111: It looks the two mapping functions can be grouped together without being collected first?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterUsageMetricsTask.java
```diff
@@ -0,0 +1,196 @@
...
+      case Maximum:
+        return datapoint.getMaximum();
+      case Average:
+        return datapoint.getAverage();
+      default:
+        return null;
+    }
+  }
+
+  private Stream<Object[]> zipMetricStreams(List<Stream<Pair<Date, Double>>> metricStreams) {
```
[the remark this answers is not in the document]
cyulysses-corp: Pair<Date, Double> does not have the best readability as a function argument. Consider moving the operation on Pair<Date, Double> into the upstream operations, if possible.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterUsageMetricsTaskTest.java
```diff
@@ -0,0 +1,152 @@
...
+      ImmutableList.of(
+          new Cluster().withClusterIdentifier("clId1"),
+          new Cluster().withClusterIdentifier("clId2"));
+
+  private static final ZonedDateTime CURR_DATE_TIME =
+      ZonedDateTime.of(2024, 01, 01, 11, 0, 0, 0, ZoneId.of("UTC"));
+  private static final ZonedInterval TEST_INTERVAL =
+      new ZonedInterval(
+          ZonedDateTime.of(2024, 01, 01, 10, 0, 0, 0, ZoneId.of("UTC")),
+          ZonedDateTime.of(2024, 01, 01, 10, 10, 0, 0, ZoneId.of("UTC")));
```
cyulysses-corp, line 68: Prefer using different numbers for different timestamp components. e.g. 2024, 01, 02, 10, 11, 12 ...... for the test values.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterUsageMetricsTask.java
```diff
@@ -0,0 +1,196 @@
...
+import java.util.Map;
+import java.util.stream.Stream;
+import javax.annotation.Nonnull;
+import org.apache.commons.lang3.tuple.Pair;
+
+/** Extraction task to get Redshift time series metrics from AWS CloudWatch API. */
+public class RedshiftClusterUsageMetricsTask extends AbstractAwsApiTask {
+
+  protected static enum MetricName {
+    CPUUtilization,
```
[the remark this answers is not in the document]
kpiotrowski: "CPUUtilization" is the AWS metric name so it has to be like this. I could attach additional value to the Enum, but this seems redundant in this case.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftRawLogsConnector.java
```diff
@@ -233,4 +250,29 @@ private void makeTasks(
       out.addTask(new JdbcSelectIntervalTask(file, query, interval));
     }
   }
+
+  /** Creates tasks to get Redshift cluster metrics from AWS CloudWatch API. */
+  private void makeClusterMetricsTasks(
+      ConnectorArguments arguments,
+      ZonedIntervalIterable intervals,
+      String filePrefix,
```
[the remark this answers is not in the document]
kpiotrowski: There is no use case for it at this point. In theory we could use use it to generate separate files with different set of redshift metrics, but for now this seems redundant.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterUsageMetricsTask.java
```diff
@@ -0,0 +1,196 @@
...
+                              .collect(toImmutableList()))
+                      .map(
+                          metricData ->
+                              Stream.of(new Object[] {cluster.getClusterIdentifier()}, metricData)
+                                  .flatMap(Stream::of)
+                                  .toArray());
+                }));
+  }
+
+  private List<Cluster> listClusters() {
```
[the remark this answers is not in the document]
kpiotrowski: <List> type is retuned by AWS SDK, copying it to the Immutable it seems redundant here.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterUsageMetricsTask.java
kpiotrowski: I think we cannot merge because first `map` generates data for the single metric, then data is zipped and then second map applies additional column to the result.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterUsageMetricsTask.java
```diff
@@ -0,0 +1,196 @@
...
+
+    GetMetricStatisticsResult result = client.getMetricStatistics(request);
+    return result.getDatapoints().stream()
+        .map(
+            datapoint ->
+                Pair.of(datapoint.getTimestamp(), getDatapointValue(metricConfig, datapoint)));
+  }
+
+  private Double getDatapointValue(MetricConfig metricConfig, Datapoint datapoint) {
+    switch (metricConfig.type()) {
```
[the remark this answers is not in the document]
kpiotrowski: I suspect CPU max could be useful in the future, but we do not have a use case for it now. I'll remove `Maximum` metric for now, code can be easily extended in the future.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterUsageMetricsTask.java
kpiotrowski: image

It returns empty list when datapoints do not exist in the given time period. No error is returned.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftRawLogsConnector.java
kpiotrowski: I changed Date format to `yyyy-MM-dd HH/mm/ss`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterUsageMetricsTask.java
```diff
@@ -0,0 +1,196 @@
...
+      Integer metricIndex = Integer.valueOf(i);
+      metricStreams
+          .get(metricIndex)
+          .forEach(
+              dataPoint -> {
+                String dateString = DATE_FORMAT.format(dataPoint.getKey().toInstant());
+                if (!metricsMap.containsKey(dateString)) {
+                  metricsMap.put(dateString, new Double[metricStreams.size()]);
+                }
+                metricsMap.get(dateString)[metricIndex] = dataPoint.getValue();
```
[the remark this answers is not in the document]
kpiotrowski: I get rid of `Object/Object[]`, Streams and improved a few things in the code. I think it is much more readable now. Serialization to `Object[]` happens in `doRun()` method now that is writing it to CSV file. `zipMetrics` is still based on the index, but I think this is the easiest solution that we can have. Introducing additional maps would help, but I think it will overcomplicate it.

I thought about exporting each metric as a separate collection of CSV files (to avoid metric aggregation), but I think current solution allows us to easily extend data with additional metrics and keep the same ingestion process.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterUsageMetricsTask.java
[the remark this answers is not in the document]
kpiotrowski: I replaced Pair with `MetricDataPoint` AutoValue, removed Streams and replaced `Object` with move descriptive types.

dominik-klimczak (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterUsageMetricsTask.java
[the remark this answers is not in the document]
dawidxc: In case there is an external API that requires this exact value, I would suggest introducing a String constant. The enum is not required for `MetricName`, because there is no comparison of the enum constants in the code in this PR (like `if` or `switch` over the enum constants). However, enum also works, if the serialized form is passed to the constructor as String.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterUsageMetricsTask.java
```diff
@@ -0,0 +1,213 @@
...
+    writeRecordsCsv(
+        sink,
+        listClusters().stream()
+            .flatMap(
+                cluster ->
+                    getClusterMetrics(cluster.getClusterIdentifier()).entrySet().stream()
+                        .map(
+                            metricEntry ->
+                                Stream.of(
+                                        new Object[] {
```
dawidxc, line 123: I would suggest wrapping `Object[]` into a wrapper (e.g. `CsvRow`) and then adding the operation
`CsvRow prepend(Object... values)` that would encapsulate this expression `Stream.of(...)...toArray()`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterUsageMetricsTask.java
```diff
@@ -0,0 +1,213 @@
...
+    SortedMap<String, Double[]> metricsMap = new TreeMap<String, Double[]>();
+
+    metricsCollection.stream()
+        .forEach(
+            metricDatapoints -> {
+              int metricIndex = metricsCollection.indexOf(metricDatapoints);
+
+              metricDatapoints.forEach(
+                  dataPoint -> {
+                    String dateString = DATE_FORMAT.format(dataPoint.date().toInstant());
```
dawidxc, line 188: I would suggest changing the type of key in the map to `Instant` and moving `DATE_FORMAT.format` closer to the serialization to CSV row.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftRawLogsConnector.java
dawidxc: As discussed offline, this datetime format is about the timestamp that is inserted into the filename and it still needs changing.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterUsageMetricsTask.java
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  private ImmutableSortedMap<String, Double[]> zipMetrics(
+      ImmutableList<ImmutableList<MetricDataPoint>> metricsCollection) {
+    SortedMap<String, Double[]> metricsMap = new TreeMap<String, Double[]>();
+
+    metricsCollection.stream()
+        .forEach(
+            metricDatapoints -> {
+              int metricIndex = metricsCollection.indexOf(metricDatapoints);
```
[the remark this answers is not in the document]
kpiotrowski: I changed aggregation logic a little bit. `MetricDataPoint`is now extended with `MetricConfig` so I can print columns in the right order during CSV serialization. This simplified the whole aggregation part as I don't need to worry about keeping order and handling missing datapoints. WDYT?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterUsageMetricsTask.java
```diff
@@ -0,0 +1,207 @@
...
+    return null;
+  }
+
+  private Object[] serializeCsvRow(
+      String clusterId, Instant instant, List<MetricDataPoint> dataPoints) {
+    Map<MetricConfig, Double> values =
+        dataPoints.stream().collect(toMap(MetricDataPoint::metricConfig, MetricDataPoint::value));
+    return Stream.of(
+            new Object[] {clusterId, DATE_FORMAT.format(instant)},
+            metrics.stream().map(metric -> values.get(metric)).toArray())
```
dawidxc, line 141: The code currently uses 2 steps to sort the data points: first, it collects the points to the intermediate map `values` and then it extracts the data points from the map.

I would suggest replacing this with the comparator in the constructor and then performing the mapping in the single step in this method:

```
private final Comparator<MetricDataPoint> dataPointsByMetricsComparator;

in ctor:
dataPointsByMetricsComparator = comparing(MetricDataPoint::metricConfig, Ordering.explicit(metrics));

in this method:
Object[] orderedDataPoints =
  dataPoints.stream().sorted(dataPointsByMetricsComparator).map(MetricDataPoint::value).toArray();
```
https://guava.dev/releases/snapshot/api/docs/com/google/common/collect/Ordering.html#explicit(java.util.List)

dawidxc (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterUsageMetricsTask.java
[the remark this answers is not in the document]
dawidxc: I've added one more comment about simplifying the data points sorting, but overall looks good. Thank you for the changes.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftClusterUsageMetricsTask.java
dawidxc: My mistake. I didn't take into account that the input list might be incomplete.

In this case, the only optimization that I can see right now is reducing the two `toArray` invocations to one:
```
Stream.concat(Stream.of(clusterId, ...), metrics.stream().map(...)).toArray()
```

cyulysses-corp (approved):

PR #338, opened 2024-01-11
Add new flag to change Teradata Transaction mode

dawidxc:
The default transaction mode in the dumper is ANSI.

The new command-line flag `-Dteradata.tmode` allows setting this mode to one of ANSI, TERA, DEFAULT or NONE for the `teradata` and `teradata-logs` connectors.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/AbstractTeradataConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/AbstractTeradataConnector.java
@@ -16,8 +16,12 @@
  */
 package com.google.edwmigration.dumper.application.dumper.connector.teradata;
 
+import static com.google.edwmigration.dumper.application.dumper.connector.teradata.TeradataUtils.determineTransactionMode;
+
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
 import com.google.common.io.ByteSink;
 import com.google.edwmigration.dumper.application.dumper.ConnectorArguments;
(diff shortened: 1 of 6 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/AbstractTeradataConnector.java
```diff
@@ -87,6 +93,31 @@ public abstract class AbstractTeradataConnector extends AbstractJdbcConnector {
   @VisibleForTesting /* pp */ static final String DEF_LOG_TABLE = "dbc.DBQLogTbl";
   @VisibleForTesting /* pp */ static final String DEF_QUERY_TABLE = "dbc.DBQLSQLTbl";
 
+  protected enum CommonTeradataConnectorProperty implements ConnectorProperty {
+    TMODE(
+        "teradata.tmode",
+        "Transaction mode - one of ANSI, TERA, DEFAULT or NONE. If NONE is used, then no"
```
ishmum123, line 99: What's the difference between using `NONE` vs `DEFAULT`?

ishmum123 (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/AbstractTeradataConnector.java
ishmum123: ref: https://teradata-docs.s3.amazonaws.com/doc/connectivity/jdbc/reference/current/jdbcug_chapter_2.html#URL_TMODE

dawidxc: `NONE` does not specify the parameter in the JDBC connection string. `DEFAULT` passes the value `DEFAULT` to the Teradata server. It's possible that the effect is the same, but I wanted to keep both options for completeness, in case there is some difference.

PR #342, opened 2024-01-25
Add flag for Hive SASL QOP

ishmum123:


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -366,6 +368,21 @@ public class ConnectorArguments extends DefaultArguments {
           .ofType(String.class)
           .describedAs("principal/host");
 
+  private final OptionSpec<String> optionHiveSaslQop =
```
dawidxc, line 371: I would suggest using a different naming convention for Hive-specific flags: `-Dhiveql.hive-sasl-qop`. For an example, please see the `-Dteradata.metadata.tablesizev.max-rows=value` flag:
https://github.com/google/dwh-migration-tools/blob/9162ae5823a8e46468a5758d571a4976a204f180/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataMetadataConnector.java#L66-L67

As for other Hive-specific flags that do not follow this convention, we can migrate them later.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -366,6 +368,21 @@ public class ConnectorArguments extends DefaultArguments {
...
+      parser
+          .accepts(
+              OPT_HIVE_SASL_QOP,
+              "The QOP of the SASL connection between hadoop and the dumper. "
+                  + "Please check the 'hadoop.rpc.protection' configuration for your cluster. "
+                  + "Mapping between the 'hadoop.rpc.protection' value and SASL QOP:\n"
+                  + "authentication: auth\n"
+                  + "integrity: auth-int\n"
+                  + "privacy: auth-conf")
+          .withOptionalArg()
```
dawidxc, line 381: Why is the argument optional?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
dawidxc: Correction: `-Dhiveql.sasl-qop`. It doesn't make sense to repeat the `hive` word twice in the option name.

dumper/lib-ext-hive-metastore/src/main/java/com/google/edwmigration/dumper/ext/hive/metastore/HiveMetastoreThriftClient.java
```diff
@@ -159,7 +167,7 @@ private TTransport createTransport() throws SaslException, TTransportException {
 
       Map<String, String> saslProperties = new HashMap<>();
       saslProperties.put(Sasl.SERVER_AUTH, "true");
-      saslProperties.put(Sasl.QOP, "auth-conf");
+      saslProperties.put(Sasl.QOP, saslQop);
```
frsv, line 170: Current code implies that if nullable `kerberosUrl` is not null, the `saslQop` is also not-null. Consider adding a nullnes check  for `saslQop` too.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -366,6 +368,21 @@ public class ConnectorArguments extends DefaultArguments {
...
+  private final OptionSpec<String> optionHiveSaslQop =
+      parser
+          .accepts(
+              OPT_HIVE_SASL_QOP,
+              "The QOP of the SASL connection between hadoop and the dumper. "
+                  + "Please check the 'hadoop.rpc.protection' configuration for your cluster. "
+                  + "Mapping between the 'hadoop.rpc.protection' value and SASL QOP:\n"
+                  + "authentication: auth\n"
+                  + "integrity: auth-int\n"
+                  + "privacy: auth-conf")
```
frsv, line 380: Consider making a enum for internal values of `saslQop` and converting user input via `withValuesConvertedBy` to this enum to guard against unexpected values.

dumper/lib-ext-hive-metastore/src/main/java/com/google/edwmigration/dumper/ext/hive/metastore/HiveMetastoreThriftClient.java
ishmum123: I feel `saslQop` can never be null as it has a default value. Let me rather add `@NonNull` before `saslQop`

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
dawidxc: The user can either provide the option or not.
If they don't provide the option, the default is used (`auth-conf`).
If they do provide the option, it can either accept an argument or not. I think that in case of this option the argument should always be provided, i.e. it should be required, not optional.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/DefaultArguments.java
```diff
@@ -124,6 +124,47 @@ public String valuePattern() {
...
+      for (HadoopSaslQop qop : HadoopSaslQop.values()) {
+        if (qop.qopValue.equals(value) || qop.name().equalsIgnoreCase(value)) {
+          return qop.qopValue;
+        }
+      }
+      throw new ValueConversionException("Not a valid QOP: " + value);
+    }
+
+    @Override
+    public Class<? extends String> valueType() {
```
dawidxc, line 156: Methods `valueType` and `valuePattern` are unused. I would suggest removing them and making `HadoopSaslQopValueConverter` independent of joptsimple.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/DefaultArguments.java
```diff
@@ -124,6 +124,47 @@ public String valuePattern() {
...
+    private HadoopSaslQopValueConverter() {}
+
+    @Override
+    public String convert(String value) {
+      for (HadoopSaslQop qop : HadoopSaslQop.values()) {
+        if (qop.qopValue.equals(value) || qop.name().equalsIgnoreCase(value)) {
+          return qop.qopValue;
+        }
+      }
+      throw new ValueConversionException("Not a valid QOP: " + value);
```
dawidxc, line 152: I would suggest throwing `MetadataDumperUsageException` to make this converter independent of joptsimple library.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ConnectorPropertyWithDefault.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.edwmigration.dumper.application.dumper.connector;
+
+import javax.annotation.Nullable;
+
+public interface ConnectorPropertyWithDefault extends ConnectorProperty {
```
dawidxc, line 21: Optional: I would very slightly suggest avoiding the name containing the `With` word, unless there is no other option. In this case I don't have any good ideas, but I would rename the interface to `RichConnectorProperty`.

frsv (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ConnectorPropertyWithDefault.java
ishmum123: the prefix `Rich` seems a bit vague to me. I also don't like the current naming but could not find a better one

PR #344, opened 2024-01-30
Fix queries printed in the summary

dawidxc:
Queries from `teradata-logs` connector printed in the summary did not exclude the missing columns. This means that the actual query executed against the Teradata database did not contain the column in the SELECT clause, while the query printed in the task summary contained it. This is now fixed, so that the same query that is executed is printed in the task summary.

Additionally, I've perfomed cleanup to reduce duplicated code and unify how queries are printed in the task summary.


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsJdbcTask.java
```diff
@@ -299,7 +308,12 @@ private Boolean isValid(@Nonnull JdbcTemplate template, @Nonnull String expressi
   }
 
   @Override
-  public String toString() {
-    return getSql(Predicates.alwaysTrue());
+  public String describeSourceData() {
+    return sql.map(AbstractTask::createSourceDataDescriptionForQuery)
```
kpiotrowski, line 312: What is a downside of running `getSql` here? This would simplify `describeSourceData` function as we do not need to handle empty Optional.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsJdbcTask.java
```diff
@@ -171,17 +176,21 @@ private String getSql(@Nonnull JdbcHandle handle) {
    * @param predicate A predicate to compute whether a given expression is legal.
    * @return A SQL query containing every legal expression from EXPRESSIONS.
    */
-  @ForOverride
   @Nonnull
-  /* pp */ String getSql(@Nonnull Predicate<? super String> predicate) {
-    return getSql(predicate, EXPRESSIONS);
+  /* pp */ String getSql(
```
kpiotrowski, line 180: optional: `getSql` method modifies class property now, so this is no longer a simple getter. I would suggest renaming it to sth more appropriate (e.g. prepareSql ?). 
Additionally if you think calling `buildSql` function twice is fine (once to get SQL, second to print sql) you could remove private `sql` property and simplify this logic.

kpiotrowski (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/FormatTask.java
```diff
@@ -39,7 +39,7 @@ protected Void doRun(TaskRunContext context, ByteSink sink, Handle handle) throw
   }
 
   @Override
-  public String toString() {
-    return "Write " + getTargetPath() + " containing '" + format + "'.";
+  public String describeSourceData() {
+    return "containing '" + format + "'.";
```
cyulysses-corp, line 43: nit, optional: Consider using String.format() instead to avoid things like " " and "'" and "'."

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractTask.java
```diff
@@ -106,4 +106,23 @@ public T run(TaskRunContext context) throws Exception {
...
+  }
+
+  /**
+   * Returns the source data description for the SQL query.
+   *
+   * @param query SQL query
+   * @return the source data description
+   */
+  protected static String createSourceDataDescriptionForQuery(String query) {
+    return "from\n        " + query;
```
cyulysses-corp, line 121: nit: how about something similar to "from\n SQL query: " ? It's hard to count spaces.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/FormatTask.java
cyulysses-corp: (And similarly for other describeSourceData() overrides.)

cyulysses-corp (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsJdbcTask.java
dawidxc: The current implementation doesn't allow running `getSql` here easily, because we do not have the right `Predicate`. The `alwaysTrue` predicate was used before, which led to the confusion that I've described in the PR description - the printed query and the executed query were different, which is undesirable.

The idea of passing the `Predicate` into the `getSql` function itself is a bit convoluted. It would be better to prepare the SQL query first, save it in the `*Task` class and then execute the task. However, the check for the existing columns is being performed in the same stage as executing the query (right before the query is executed), so there is no separation of creating a query and then creating a task with this query.

Because such an SQL query is dynamically generated based on the schema of the database that is known only after connecting to the database, I rely on the fact that the query is executed first and printed only later in the summary at the end of dumper execution. Therefore saving it in the field in the `*Task` class and retrieving it here makes sense.

For the long-term: A better solution would require changing the whole structure of how tasks are prepared, so that the SQL query could be passed to a `*Task` class and be immutable there, but this would be a much bigger change.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/TeradataLogsJdbcTask.java
dawidxc: I've renamed the method to `getOrCreateSql`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/FormatTask.java
dawidxc: Done.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractTask.java
dawidxc: I have preserved the spaces to maintain the current pretty-printed output.

```
* Task Summary:
* Task SUCCEEDED (INFORMATIONAL) Write compilerworks-version.txt from product version information.
* Task SUCCEEDED (OPTIONAL) Write dbc.ResUsageSpma_2023-03-28T10:00:00Z.csv from
        SELECT * FROM dbc.ResUsageSpma WHERE TheTimestamp >= 1679997600 AND TheTimestamp < 1680001200
* Task SUCCEEDED (REQUIRED) Write utility_logs_2023-03-28T10:00:00Z.csv from
        SELECT AcctString, AcctStringDate, ...
* Task SUCCEEDED (REQUIRED) Write compilerworks-metadata.yaml containing dump metadata.
* Task SUCCEEDED (REQUIRED) Write compilerworks-format.txt containing 'teradata.logs.zip'.
* Task SUCCEEDED (OPTIONAL) Write dbc.ResUsageScpu_2023-03-28T10:00:00Z.csv from
        SELECT * FROM dbc.ResUsageScpu WHERE TheTimestamp >= 1679997600 AND TheTimestamp < 1680001200
* Task SUCCEEDED (INFORMATIONAL) Write compilerworks-arguments.txt from command-line arguments.
* Task SUCCEEDED (REQUIRED) Write query_history_2023-03-28T10:00:00Z.csv from
        SELECT L.QueryID, ST.SQLRowNo, ST.SQLTextInfo, ...
```

PR #345, opened 2024-01-31
Validate ConnectorProperty

ishmum123:


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorPropertyValidator.java
```diff
@@ -0,0 +1,49 @@
...
+        Arrays.stream(connector.getConnectorProperties().getEnumConstants())
+            .map(enumValue -> ((ConnectorProperty) enumValue).getName())
+            .collect(Collectors.toSet());
+    definitionMap
+        .keySet()
+        .forEach(
+            key -> {
+              if (!properties.contains(key)) {
+                throw new IllegalArgumentException(
+                    key + " is not a recognized option for " + connector.getName());
```
dawidxc, line 45: Such validation is already performed here:
https://github.com/google/dwh-migration-tools/blob/d0ba6ded9de651c5786660a305604701fcdcb8ef/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java#L888-L892

ishmum123: interesting. I missed it for some reason. Do you think making the validation more strict, i.e. - throwing an exception, makes sense in here?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -884,14 +884,12 @@ private static ImmutableMap<String, String> buildDefinitionMap(
         resultMap.put(name, value);
         if (connector != null && propertyNamesByConnector.get(connector).contains(name)) {
           LOG.info("Parsed property: name='{}', value='{}'", name, value);
-        } else if (allPropertyNames.contains(name)) {
```
dawidxc, line 887: I would suggest keeping the distinction between the unknown property and the property used in the wrong connector.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -884,14 +884,12 @@ private static ImmutableMap<String, String> buildDefinitionMap(
...
-              name,
-              value,
-              MoreObjects.firstNonNull(connector, "[not specified]"));
         } else {
-          LOG.warn("Skipping unknown property: name='{}', value='{}'", name, value);
+          String msg =
+              String.format(
+                  "Property: name='%s', value='%s' is not compatible with connector '%s'",
+                  name, value, MoreObjects.firstNonNull(connector, "[not specified]"));
+          throw new MetadataDumperUsageException(msg);
```
dawidxc, line 892: I would suggest inlining the `msg` variable.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
ishmum123: Here is my thought process - The person that's using the dumper (with very high probability) has only one dialect to think about, e.g. - Teradata OR Redshift OR .... For him/her, it doesn't matter if the flag is correct for some other dialect and the message might force him/her to think about that. It feels more user friendly to just say "this is not a valid flag you can use". wdyt?

dawidxc: > The person that's using the dumper (with very high probability) has only one dialect to think about
It's true, but even for one dialect we can have two connectors - for example, for Teradata we have a separate connector for metadata and for query logs. Some flags are compatible only with one of these connectors for the same dialect.

> It feels more user friendly to just say "this is not a valid flag you can use". 
The current implementation prints something different - "Property: name='%s', value='%s' is not compatible with connector '%s'". The user may think: "What are the implications? Is this property compatible with another connector? Or is it invalid altogether?"

I agree that for the first time user the message might not be that important, but for someone who runs the dumper a few times, it may be important, whether they've made a typo in the flag name or whether they provided a correct flag name, but used a wrong connector. Therefore I think that we should distinguish between "Unknown property" (or "Unrecognized property") and the known property, but associated with another connector.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/MetadataDumperTest.java
```diff
@@ -138,4 +140,62 @@ public void testDoesNotOverrideFileWithDirectory() throws IOException {
     // Assert
     assertTrue(exception.getMessage().startsWith("A file already exists at test"));
   }
+
+  @Test
+  public void testFailsOnUnrecognizedFlag() {
+    OptionException exception =
+        Assert.assertThrows(
+            "No exception thrown from " + dumper.getClass().getSimpleName(),
```
dawidxc, line 148: Optional: This is not very important, but I think that adding this message doesn't help much. I would suggest removing it (also from the existing test in this class).

The difference between the test failing with and without the message is:

with:
```
No exception thrown from Main: expected java.lang.IllegalStateException to be thrown, but nothing was thrown
```

without:
```
expected java.lang.IllegalStateException to be thrown, but nothing was thrown
```

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/test/TestConnector.java
```diff
@@ -35,10 +36,37 @@ public TestConnector() {
     super("test");
   }
 
+  public enum TestConnectorProperty implements ConnectorProperty {
+    TEST_PROPERTY("test.property", "This a a test property");
```
dawidxc, line 40: nit: I would suggest fixing the description: `This is a property`.

dawidxc: Correction: `This is a test property`.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/MetadataDumperTest.java
```diff
@@ -138,4 +140,62 @@ public void testDoesNotOverrideFileWithDirectory() throws IOException {
     // Assert
     assertTrue(exception.getMessage().startsWith("A file already exists at test"));
   }
+
+  @Test
+  public void testFailsOnUnrecognizedFlag() {
+    OptionException exception =
+        Assert.assertThrows(
```
dawidxc, line 147: I would suggest adding a static import for `assertThrows`.

dawidxc (approved):

PR #347, opened 2024-01-31
Fix rowSql null check

dawidxc:


kpiotrowski (approved):
LGTM 👍

shevek-google:
Given that this was also an assignment-of-wrong-value, do we have a test case that detects it now? Why did we not see a failure in CI?

dawidxc:
@shevek-google I've added a test case in #361.

PR #351, opened 2024-02-02
Solve performance issue in Oracle metadata task

kpiotrowski:
Oracle metadata extraction takes extremely lot of time for some tables. It processes 5-8 records per second resulting in many hours of extraction. This is caused by extracting columns with type `LONG`. 

This PR excludes following columns:

```
ALL_ARGUMENTS.DEFAULT_VALUE
ALL_CONSTRAINTS.SEARCH_CONDITION
ALL_TAB_COLUMNS.DATA_DEFAULT
ALL_TAB_PARTITIONS.HIGH_VALUE
ALL_VIEWS.TEXT
```

In addition it comments oracle XML tasks as all are currently broken.


kpiotrowski:
@shevek-google Do you know if similar issue happened before? Can we safely exclude these columns?

kpiotrowski:
Closing this PR as there is alternative solution that allows us to keep LONG columns: #352

PR #352, opened 2024-02-05
Set Oracle useFetchSizeWithLongColumn argument to true

kpiotrowski:
Oracle metadata extraction takes extremely lot of time for some tables. It processes 5-8 records per second resulting in many hours of extraction. This is caused by extracting columns with type LONG.

useFetchSizeWithLongColumn parameter enables prefetch of rows with a LONG or LONG RAW column. It significantly reduces extraction time. (for my test instance: 10h -> 5min).

This PR enables `useFetchSizeWithLongColumn` by default. User can disable it by setting `-Doracle.useFetchSizeWithLongColumn=false` command line argument or including this property in the connection URL.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnector.java
@@ -73,6 +73,10 @@ import org.springframework.jdbc.datasource.SimpleDriverDataSource;
 })
 public abstract class AbstractOracleConnector extends AbstractJdbcConnector {
 
+  /* This argument is required to improve performance of retrieving LONG columns */
+  private static final String USE_FETCH_SIZE_WITH_LONG_COLUMN_ARG =
+      "useFetchSizeWithLongColumn=true";
+
   public static final int OPT_PORT_DEFAULT = 1521;
 
   public AbstractOracleConnector(@Nonnull String name) {
@@ -106,7 +110,8 @@ public abstract class AbstractOracleConnector extends AbstractJdbcConnector {
         url = "jdbc:oracle:thin:@//" + host + ":" + port + "/" + arguments.getOracleServicename();
       }
     }
-    // LOG.info("URL IS " + url);
+    url = String.format("%s?%s", USE_FETCH_SIZE_WITH_LONG_COLUMN_ARG);
+
     Driver driver = newDriver(arguments.getDriverPaths(), "oracle.jdbc.OracleDriver");
     DataSource dataSource =
         new SimpleDriverDataSource(driver, url, arguments.getUser(), arguments.getPassword());
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnector.java
@@ -26,7 +26,7 @@ import com.google.edwmigration.dumper.application.dumper.annotations.RespectsArg
 import com.google.edwmigration.dumper.application.dumper.annotations.RespectsInput;
 import com.google.edwmigration.dumper.application.dumper.annotations.RespectsInputs;
 import com.google.edwmigration.dumper.application.dumper.connector.AbstractJdbcConnector;
-import com.google.edwmigration.dumper.application.dumper.connector.ConnectorProperty;
+import com.google.edwmigration.dumper.application.dumper.connector.ConnectorPropertyWithDefault;
 import com.google.edwmigration.dumper.application.dumper.handle.Handle;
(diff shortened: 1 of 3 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnector.java
```diff
@@ -106,7 +110,8 @@ public Handle open(ConnectorArguments arguments) throws Exception {
         url = "jdbc:oracle:thin:@//" + host + ":" + port + "/" + arguments.getOracleServicename();
       }
     }
-    // LOG.info("URL IS " + url);
+    url = String.format("%s?%s", url, USE_FETCH_SIZE_WITH_LONG_COLUMN_ARG);
```
dawidxc, line 113: 1. It seems that the question mark is available only in `Easy Connect Plus` Oracle JDBC URLs that work only since Oracle 19c.
https://docs.oracle.com/en/database/oracle/oracle-database/21/jjdbc/data-sources-and-URLs.html#GUID-C7D48803-09E7-42C5-A0F6-32B009D0C20C

    The older versions used `new OracleDataSource().setConnectionProperties(prop)`. Would it make sense to support all the options - 1. existing code for backwards-compatibility, 2. the Oracle 19c compatible code and 3. the `setConnectionProperties`?

2. If the URL is provided in the command-line, it can already contain the `?` character. The concatenation of a new property should take it into account and use an `&` character in such a case. Also, the property can already be included in the URL - setting the flag to either false or true. The property should not be included twice. It also probably should not change the value provided in the command-line.
https://docs.oracle.com/en/database/oracle/oracle-database/21/jjdbc/data-sources-and-URLs.html#GUID-843F6EC2-EDFD-4A88-98DE-B45E1CEADC0E

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnector.java
```diff
@@ -73,6 +73,10 @@
 })
 public abstract class AbstractOracleConnector extends AbstractJdbcConnector {
 
+  /* This argument is required to improve performance of retrieving LONG columns */
```
dawidxc, line 76: This flag can improve performance, but it may also cause issues, if the data is too big.
https://docs.oracle.com/en/database/oracle/oracle-database/21/jajdb/oracle/jdbc/OracleConnection.html#CONNECTION_PROPERTY_USE_FETCH_SIZE_WITH_LONG_COLUMN

I would suggest making it possible to turn off this parameter, so that the user can choose to wait longer and retrieve the data without errors if enabling the flag causes issues.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnector.java
kpiotrowski: I changed the approach to set the connection properties instead of modifying URL. I tested it with the URL that already includes this property and it has a higher priority.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnector.java
```diff
@@ -75,6 +79,46 @@ public abstract class AbstractOracleConnector extends AbstractJdbcConnector {
 
   public static final int OPT_PORT_DEFAULT = 1521;
 
+  protected enum CommonOracleConnectorProperty implements ConnectorProperty {
```
dawidxc, line 82: I would suggest implementing `ConnectorPropertyWithDefault`.
https://github.com/google/dwh-migration-tools/blob/b82b4e98f755a12c44c12ed33f3cede40865f41d/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ConnectorPropertyWithDefault.java#L21

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnector.java
```diff
@@ -75,6 +79,46 @@ public abstract class AbstractOracleConnector extends AbstractJdbcConnector {
...
+    USE_FETCH_SIZE_WITH_LONG_COLUMN(
+        "oracle.useFetchSizeWithLongColumn",
```
dawidxc, line 84: We do not have any flags that use camelCase. Existing examples:

```
-Dteradata-logs.query-logs-table
-Dteradata.metadata.tablesizev.max-rows
-Dsnowflake.metadata.databases.where
-Dteradata.tmode
-Dsnowflake.query_acceleration_history.query
```

There is no naming convention, but I would suggest following one of the existing case formats.

dawidxc (approved):

PR #353, opened 2024-02-08
Make Snowflake QUERY_ACCELERATION_HISTORY optional

dawidxc:
The extraction of Snowflake `QUERY_ACCELERATION_HISTORY` table is sometimes not possible, therefore I'm changing the TaskCategory from `REQUIRED` to `OPTIONAL`, so that the extraction doesn't fail in case this table is not available.

Additionally, cleanup the `TeradataJdbcSelectTask` class, because otherwise the `TeradataJdbcSelectTask` class would contain a duplicated code between it and the `JdbcSelectTask` class.


cyulysses-corp (approved):
LGTM to code. Could you also write in the description that this PR also aligns  AbstractTeradataConnector's category logic with that of JdbcSelectTask?

PR #355, opened 2024-02-09
Add OptionalUtils

dawidxc:
Move general-purpose methods related to `Optional` from `TeradataUtils` to `OptionalUtils`. Use the methods to simplify code.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
@@ -17,7 +17,6 @@
 package com.google.edwmigration.dumper.application.dumper;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
-import static com.google.edwmigration.dumper.application.dumper.utils.OptionalUtils.nonEmpty;
 import static com.google.edwmigration.dumper.application.dumper.utils.OptionalUtils.optionalIf;
 import static java.time.temporal.ChronoUnit.DAYS;
 import static java.time.temporal.ChronoUnit.HOURS;
@@ -975,7 +974,11 @@ public class ConnectorArguments extends DefaultArguments {
 
   @CheckForNull
   public String getDefinitionOrDefault(ConnectorPropertyWithDefault property) {
-    return nonEmpty(getDefinition(property)).orElseGet(property::getDefaultValue);
+    String stringValue = getDefinition(property);
+    if (StringUtils.isEmpty(stringValue)) {
+      return property.getDefaultValue();
+    }
+    return stringValue;
   }
 
   public static class ZonedParser implements ValueConverter<ZonedDateTime> {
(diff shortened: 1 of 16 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -976,11 +975,7 @@ public String toString() {
 
   @CheckForNull
   public String getDefinitionOrDefault(ConnectorPropertyWithDefault property) {
-    String stringValue = getDefinition(property);
-    if (StringUtils.isEmpty(stringValue)) {
-      return property.getDefaultValue();
-    }
-    return stringValue;
+    return nonEmpty(getDefinition(property)).orElseGet(property::getDefaultValue);
```
shevek-google, line 978: Are we sure this is easier to read?

dawidxc: It depends. This one is less important compared to other changes in this PR, so I've reverted it.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/OptionalUtils.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+package com.google.edwmigration.dumper.application.dumper.utils;
+
+import com.google.common.base.Strings;
+import java.util.Optional;
+import java.util.function.Supplier;
+
+public class OptionalUtils {
+
+  public static <T> Optional<T> optionalIf(boolean condition, Supplier<T> supplier) {
```
cyulysses-corp, line 25: nit: the meaning of the name is not clear to me. I would recommend something like:

optionallyNonEmptyIf

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/OptionalUtils.java
```diff
@@ -0,0 +1,34 @@
...
+    return condition ? Optional.of(supplier.get()) : Optional.empty();
+  }
+
+  public static Optional<String> nonEmpty(String value) {
```
cyulysses-corp, line 29: nit: optionallyNullableIfEmpty

cyulysses-corp (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/OptionalUtils.java
dawidxc: Renamed to `optionallyWhen` in #359 as discussed offline.

shevek-google:
My overall suggestion on PRs like this is that Optional does not improve the readability of the code, and readability is a critical requirement for code which is to be externally audited. I suspect the nullness analyzer in Checker Framework is better at doing the dataflow analysis than the (arguably technically equivalent, but nevertheless unimplemented) work to make sure isPresent() is always called before get(). We could respond "But we never call get, we always call orElse()" but that doesn't have the same assertion strength as Objects.requireNonNullElse, which also asserts on the second argument.

PR #356, opened 2024-02-09
Fix cleaning up files and directories in test

dawidxc:
The files and directories created in test were not always cleaned up due to the `file` field being either null or not assigned the correct file to delete. I'm adding `TopLevelTestFile` enum with all the possible files and directories created in tests in the `MetadataDumperTest` class and deleting them at the beginning and at the end of each test.

### DIFF
--- a/dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/MetadataDumperTest.java
+++ b/dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/MetadataDumperTest.java
@@ -28,6 +28,7 @@ import java.io.IOException;
 import joptsimple.OptionException;
 import org.apache.commons.io.FileUtils;
 import org.junit.After;
+import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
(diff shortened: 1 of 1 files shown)


dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/MetadataDumperTest.java
```diff
@@ -28,25 +28,54 @@
...
-  private final String defaultFileName = "dwh-migration-test-metadata.zip";
+
+  @Before
+  public void setUp() throws IOException {
+    deleteTopLevelTestFiles();
+  }
 
   @After
   public void tearDown() throws IOException {
-    if (file == null || !file.exists()) return;
```
cyulysses-corp, line 47: Out of curiosity, which cases would cause the file to be considered null or non-existing, while it actually exists?

cyulysses-corp (approved):

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/MetadataDumperTest.java
dawidxc: The `testCreatesDefaultOutputZipInProvidedDirectory` method creates the `dir` directory, but the `tearDown` method doesn't remove it, because the `file` field is associated with a different path.

The `testFailsOnUnrecognizedDialect` method creates the default filename `dwh-migration-test-metadata.zip`, but the `tearDown` method doesn't remove it, because the `file` field is not assigned at all.

shevek-google: Typically `file` would be null if setup failed, in which case it's nice to have that exception reported, and not masked by a subsequent failure in tearDown(). In detail, I'm not sure that's the exact case here, but that's how I would think about this as a defensive programming style.

PR #357, opened 2024-02-12
Extract connector properties from ConnectorArguments

dawidxc:
The `ConnectorArguments` class is long and it contains among others the `definitionMap` with connector properties and methods to build and access this map. I'm extracting it to a separate class to simplify the original class.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorProperties.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorProperties.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2022-2023 Google LLC
+ * Copyright 2013-2021 CompilerWorks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorProperties.java
```diff
@@ -0,0 +1,105 @@
...
+import com.google.edwmigration.dumper.application.dumper.connector.ConnectorPropertyWithDefault;
+import java.util.List;
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ConnectorProperties {
```
kpiotrowski, line 34: Should we move the `printConnectorHelp` function to this class as well? I think it would make sense as this function is used to print connector properties only.

kpiotrowski (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorProperties.java
dawidxc: The `printConnectorHelp` method prints both the connector properties and the other command-line flags. I've extracted only part of the method to a new method in `ConnectorProperties` class.

PR #328, opened 2023-12-20
[Dumper] [Hive] Extract hive table SerDe details

yogeshtewari:
Update Hive connector to extract Hive table StorageDescriptor details for SerDe and Input/Output formats. This will help in importing hive table metadata into DPMS.

PR #364, opened 2024-03-11
Wrap ``--add-opens=java.base/java.nio=ALL-UNNAMED` as a default envir…

cyulysses-corp:
…onment variable for the .sh script.

This is required to deal with Snowflake JDBC error https://community.snowflake.com/s/article/JDBC-Driver-Compatibility-Issue-With-JDK-16-and-Later , which recently re-emerged and is likely to re-emerge again.

### DIFF
--- a/bin/dwh-migration-dumper
+++ b/bin/dwh-migration-dumper
@@ -8,6 +8,6 @@ if [ ! -x "$BIN" ] ; then
 fi
 
 # "ALL-UNNAMED" option is required as a workaround for Snowflake JDBC Driver error.
-_JAVA_OPTIONS="--add-opens=java.base/java.nio=ALL-UNNAMED"
+_JAVA_OPTIONS="$_JAVA_OPTIONS --add-opens=java.base/java.nio=ALL-UNNAMED"
 
 exec env _JAVA_OPTIONS="$_JAVA_OPTIONS" "$BIN" "$@"
--- a/bin/dwh-migration-dumper
+++ b/bin/dwh-migration-dumper
@@ -7,4 +7,7 @@ if [ ! -x "$BIN" ] ; then
   (cd "${BASE_DIR}" && ./gradlew --parallel :dumper:app:installDist)
 fi
 
-exec "$BIN" "$@"
+# "ALL-UNNAMED" option is required as a workaround for Snowflake JDBC Driver error.
+_JAVA_OPTIONS="--add-opens=java.base/java.nio=ALL-UNNAMED"
+
+exec env _JAVA_OPTIONS="$_JAVA_OPTIONS" "$BIN" "$@"
--- a/bin/dwh-migration-dumper
+++ b/bin/dwh-migration-dumper
@@ -1,7 +1,9 @@
 #!/bin/sh -ex
 
 BASE_DIR=$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" 2>&1 > /dev/null && cd .. && pwd)
-BIN="${BASE_DIR}/dumper/app/build/install/app/bin/dwh-migration-dumper"
+. "${BASE_DIR}/dumper/app/src/main/sh/java_versioning.sh"
+BASE_BIN_DIR="${BASE_DIR}/dumper/app/build/install/app/bin"
+BIN=$(is_java_greater_than_8 && echo "${BASE_BIN_DIR}/dwh-migration-dumper" || echo "${BASE_BIN_DIR}/dwh-migration-dumper-java8")
 
 if [ ! -x "$BIN" ] ; then
   (cd "${BASE_DIR}" && ./gradlew --parallel :dumper:app:installDist)
--- a/bin/dwh-migration-dumper
+++ b/bin/dwh-migration-dumper
@@ -7,7 +7,4 @@ if [ ! -x "$BIN" ] ; then
   (cd "${BASE_DIR}" && ./gradlew --parallel :dumper:app:installDist)
 fi
 
-# "ALL-UNNAMED" option is required as a workaround for Snowflake JDBC Driver error.
-_JAVA_OPTIONS="$_JAVA_OPTIONS --add-opens=java.base/java.nio=ALL-UNNAMED"
-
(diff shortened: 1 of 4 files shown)


bin/dwh-migration-dumper
```diff
@@ -7,4 +7,7 @@ if [ ! -x "$BIN" ] ; then
   (cd "${BASE_DIR}" && ./gradlew --parallel :dumper:app:installDist)
 fi
 
-exec "$BIN" "$@"
+# "ALL-UNNAMED" option is required as a workaround for Snowflake JDBC Driver error.
+_JAVA_OPTIONS="--add-opens=java.base/java.nio=ALL-UNNAMED"
```
dawidxc, line 11: 1. What if the `_JAVA_OPTIONS` variable already contains some value? Would it make sense to preserve it `_JAVA_OPTIONS="$_JAVA_OPTIONS --add-opens..."`?

2. It seems from the linked manual that the parameter depends on the Java version. This workaround is specific for JDK 17 and 18, while for JDK 16 another workaround is suggested: `-Djdk.module.illegalAccess=permit`.

    The `_JAVA_OPTIONS` variable seems to be specific to a particular JVM implementation. The `JAVA_TOOL_OPTIONS` variable is suggested as a more general solution: https://stackoverflow.com/q/28327620

    Another possibility is to use `JAVA_OPTS` or `DWH_MIGRATION_DUMPER_OPTS`, although I'm not sure if they would work the same as `_JAVA_OPTIONS` or `JAVA_TOOL_OPTIONS`: https://github.com/gradle/gradle/blob/bda658e496d0f001bbad8c59d7cd8b6f0ad85067/platforms/jvm/plugins-application/src/main/resources/org/gradle/api/internal/plugins/unixStartScript.txt#L244

bin/dwh-migration-dumper
```diff
@@ -7,4 +7,7 @@ if [ ! -x "$BIN" ] ; then
...
+
+exec env _JAVA_OPTIONS="$_JAVA_OPTIONS" "$BIN" "$@"
```
dawidxc, line 13: Do you know, if the script would also work without changing this line?

For example,
```
export _JAVA_OPTIONS="--add-opens=java.base/java.nio=ALL-UNNAMED"
exec "$BIN" "$@"
```

cyulysses-corp: I tried and it works.

However, the `env` method is recommended by the official Arrow site: https://arrow.apache.org/docs/java/install.html#java-compatibility

bin/dwh-migration-dumper
cyulysses-corp: 1. Nice idea. Change applied.
2.a. I tried using Java 16 with this fix, and it still works.
2.b. I think it is desirable that _JAVA_OPTIONS "wins over" command-line arguments and _JAVA_TOOL_OPTIONS, because we do not want other changes to accidentally overwrite this break-preventing option. Also, it is recommended by the official Arrow site: https://arrow.apache.org/docs/java/install.html#java-compatibility

cyulysses-corp:
The spotless check error is weird - it is reporting format failures completely unrelated to this change. I am ignoring it.

dawidxc:
> The spotless check error is weird - it is reporting format failures completely unrelated to this change. I am ignoring it.

PR #328 was recently merged and that introduced spotless violations. I've prepared a fix in the PR #365.

bin/dwh-migration-dumper
dawidxc: After reading a bit more about `env` and `export`, I think they are unrelated to Arrow. The `export` command is a bash built-in and `env` is a command. They differ, but for the purpose of this script they both do the same thing (unless this script is not executed in Bash - due to `/bin/sh` in the first line, which is not necessarily Bash). If it was guaranteed that the script is executed in Bash (we should change the shebang to `/bin/bash`), I would prefer `export` so as 1. not to mix too many things on a single line and 2. keep the code short.

```
export _JAVA_OPTIONS="$_JAVA_OPTIONS --add-opens=java.base/java.nio=ALL-UNNAMED"
exec "$BIN" "$@"
```

However, if we leave `/bin/sh`, then `env` is probably the most portable solution.

https://askubuntu.com/questions/205688/whats-the-difference-between-set-export-and-env-and-when-should-i-use-each

https://unix.stackexchange.com/questions/368944/what-is-the-difference-between-env-setenv-export-and-when-to-use

dawidxc (approved):

dawidxc:
I've merged #365, so to fix spotless issues, it is enough to rebase this branch on `main`.

shevek-google (changes requested):
Is there any way to merge this into the CreateStartScripts task? There's no guarantee that the wrapper in bin/ is used - it ought to go into the generated script that ends up in the binary build,

Does this make us incompatible with some Java verions? Which ones?

cyulysses-corp:
> Is there any way to merge this into the CreateStartScripts task? There's no guarantee that the wrapper in bin/ is used - it ought to go into the generated script that ends up in the binary build,
> 
> Does this make us incompatible with some Java verions? Which ones?

Good point. I put the argument in the "application" plugin in gradle.build instead, because it is the most succinct.

Regarding Java versions, this change enables us to be compatible with Java >= 16 without breaking backward compatibility.

dumper/app/build.gradle
```diff
@@ -100,6 +100,9 @@ application {
     // Define the main class for the application.
     mainClass = 'com.google.edwmigration.dumper.application.dumper.Main'
     applicationName = 'dwh-migration-dumper'
+
+    // "ALL-UNNAMED" option is required as a workaround for Snowflake JDBC Driver error.
+    applicationDefaultJvmArgs = ["--add-opens=java.base/java.nio=ALL-UNNAMED"]
```
shevek-google, line 105: If we build this application with this argument, does it still run in JDK 8?

shevek-google (dismissed):

cyulysses-corp:
@shevek-google Friendly reminder.

cyulysses-corp:
> @shevek-google Friendly reminder.

@shevek-google Friendly ping.

PR #372, opened 2024-03-29
Add a precommit script

dawidxc:
Add a script to run before the commit to make sure that the correct formatting is applied, the unit tests are successful and the dist is building successfully.


cyulysses-corp (commented):
The presubmit is currently failing:
```
[07:16:28 PDT][ERROR] Failed to get build config
```

PR #382, opened 2024-05-08
[b/337839892] Create an empty oracle-stats connector

misolt:
- create a connector that will be used to collect `oracle-stats` data
- add an enum to represent the connector type (`logs`, `metadata` or `stats`) and use it in Oracle
- add tests verifying Oracle connector type information in output filenames and help text

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/LogsConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/LogsConnector.java
@@ -28,9 +28,9 @@ public interface LogsConnector extends Connector {
   default String getDefaultFileName(boolean isAssessment) {
     if (isAssessment) {
       Clock clock = Clock.systemDefaultZone();
-      return ArchiveNameUtil.getFileNameWithTimestamp(getName() + "-logs", clock);
+      return ArchiveNameUtil.getFileNameWithTimestamp(getName(), "logs", clock);
     } else {
-      return ArchiveNameUtil.getFileName(getName() + "-logs");
+      return ArchiveNameUtil.getFileName(getName(), "logs");
     }
   }
 }
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/LogsConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/LogsConnector.java
@@ -28,9 +28,9 @@ public interface LogsConnector extends Connector {
   default String getDefaultFileName(boolean isAssessment) {
     if (isAssessment) {
       Clock clock = Clock.systemDefaultZone();
-      return ArchiveNameUtil.getFileNameWithTime(getName(), "logs", clock);
+      return ArchiveNameUtil.getFileNameWithTimestamp(getName(), "logs", clock);
     } else {
-      return ArchiveNameUtil.getBasicFileName(getName(), "logs");
+      return ArchiveNameUtil.getFileName(getName(), "logs");
     }
   }
 }
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/LogsConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/LogsConnector.java
@@ -19,16 +19,20 @@ package com.google.edwmigration.dumper.application.dumper.connector;
 import java.text.Format;
 import java.text.SimpleDateFormat;
 import javax.annotation.Nonnull;
-import org.apache.commons.lang3.StringUtils;
 
 /** @author shevek */
 public interface LogsConnector extends Connector {
 
   @Nonnull
+  @Override
   default String getDefaultFileName(boolean isAssessment) {
+    String timeSuffix = isAssessment ? getTimeSuffix() : "";
+    return String.format("dwh-migration-%s-logs%s.zip", getName(), timeSuffix);
+  }
+
+  @Nonnull
+  static String getTimeSuffix() {
     Format format = new SimpleDateFormat("yyyyMMdd'T'HHmmss");
-    String timeSuffix = "-" + format.format(System.currentTimeMillis());
-    return String.format(
-        "dwh-migration-%s-logs%s.zip", getName(), isAssessment ? timeSuffix : StringUtils.EMPTY);
+    return "-" + format.format(System.currentTimeMillis());
   }
 }
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/LogsConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/LogsConnector.java
@@ -28,9 +28,9 @@ public interface LogsConnector extends Connector {
   default String getDefaultFileName(boolean isAssessment) {
     if (isAssessment) {
       Clock clock = Clock.systemDefaultZone();
-      return ArchiveNameUtil.getFileNameWithTimestamp(getName(), "logs", clock);
+      return ArchiveNameUtil.getFileNameWithTimestamp(getName() + "-logs", clock);
     } else {
-      return ArchiveNameUtil.getFileName(getName(), "logs");
+      return ArchiveNameUtil.getFileName(getName() + "-logs");
     }
   }
 }
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/LogsConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/LogsConnector.java
@@ -16,8 +16,8 @@
  */
 package com.google.edwmigration.dumper.application.dumper.connector;
 
-import java.text.Format;
-import java.text.SimpleDateFormat;
+import com.google.edwmigration.dumper.application.dumper.utils.ArchiveNameUtil;
+import java.time.Clock;
 import javax.annotation.Nonnull;
 
 /** @author shevek */
@@ -26,13 +26,11 @@ public interface LogsConnector extends Connector {
   @Nonnull
   @Override
   default String getDefaultFileName(boolean isAssessment) {
-    String timeSuffix = isAssessment ? getTimeSuffix() : "";
-    return String.format("dwh-migration-%s-logs%s.zip", getName(), timeSuffix);
-  }
-
-  @Nonnull
-  static String getTimeSuffix() {
-    Format format = new SimpleDateFormat("yyyyMMdd'T'HHmmss");
-    return "-" + format.format(System.currentTimeMillis());
+    if (isAssessment) {
+      Clock clock = Clock.systemDefaultZone();
+      return ArchiveNameUtil.getFileNameWithTime(getName(), "logs", clock);
+    } else {
+      return ArchiveNameUtil.getBasicFileName(getName(), "logs");
+    }
   }
 }
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
@@ -27,12 +27,12 @@ enum OracleConnectorScope {
   STATS("oracle-stats", "oracle.stats.zip", "stats");
 
   private final String connectorName;
-  private final String format;
+  private final String formatName;
   private final String resultType;
 
-  OracleConnectorScope(String displayName, String format, String resultType) {
+  OracleConnectorScope(String displayName, String formatName, String resultType) {
     this.connectorName = displayName;
-    this.format = format;
+    this.formatName = formatName;
     this.resultType = resultType;
   }
 
@@ -53,6 +53,6 @@ enum OracleConnectorScope {
   }
 
   String formatName() {
-    return format;
+    return formatName;
   }
 }
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
@@ -26,18 +26,18 @@ enum OracleConnectorScope {
   METADATA("oracle", OracleMetadataDumpFormat.FORMAT_NAME, "metadata"),
   STATS("oracle-stats", "oracle.stats.zip", "stats");
 
-  private final String displayName;
+  private final String connectorName;
   private final String format;
   private final String resultType;
 
   OracleConnectorScope(String displayName, String format, String resultType) {
-    this.displayName = displayName;
+    this.connectorName = displayName;
     this.format = format;
     this.resultType = resultType;
   }
 
-  String displayName() {
-    return displayName;
+  String connectorName() {
+    return connectorName;
   }
 
   String toFileName(boolean isAssessment) {
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
@@ -43,13 +43,13 @@ enum OracleConnectorScope {
   String toFileName(boolean isAssessment, Clock clock) {
     if (this == LOGS) {
       // add "-logs" twice for consistency with previous versions
-      String name = String.format("oracle-%s-%s", resultType, resultType);
+      String suffix = String.format("%s-%s", resultType, resultType);
       if (isAssessment) {
-        return ArchiveNameUtil.getFileNameWithTimestamp(name, clock);
+        return ArchiveNameUtil.getFileNameWithTimestamp("oracle", suffix, clock);
       }
-      return ArchiveNameUtil.getFileName(name);
+      return ArchiveNameUtil.getFileName("oracle", suffix);
     }
-    return ArchiveNameUtil.getFileName("oracle-" + resultType);
+    return ArchiveNameUtil.getFileName("oracle", resultType);
   }
 
   String formatName() {
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
@@ -43,9 +43,9 @@ enum OracleConnectorScope {
   String toFileName(boolean isAssessment) {
     if (this == LOGS && isAssessment) {
       Clock systemClock = Clock.systemDefaultZone();
-      return ArchiveNameUtil.getFileNameWithTimestamp("oracle", resultType, systemClock);
+      return ArchiveNameUtil.getFileNameWithTimestamp("oracle-" + resultType, systemClock);
     } else {
-      return ArchiveNameUtil.getFileName("oracle", resultType);
+      return ArchiveNameUtil.getFileName("oracle-" + resultType);
     }
   }
 
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
@@ -26,20 +26,16 @@ enum OracleConnectorScope {
   STATS;
 
   String toDisplayName() {
-    switch (this) {
-      case LOGS:
-        return "oracle-logs";
-      case METADATA:
-        return "oracle";
-      case STATS:
-        return "oracle-stats";
+    if (this == METADATA) {
+      return "oracle";
+    } else {
+      return "oracle-" + getResultType();
     }
-    throw new AssertionError();
   }
 
   String toFileName(boolean isAssessment) {
     String timeSuffix = getTimeSuffix(isAssessment);
-    return String.format("dwh-migration-%s%s.zip", toDisplayName(), timeSuffix);
+    return String.format("dwh-migration-oracle-%s%s.zip", getResultType(), timeSuffix);
   }
 
   String toFormat() {
@@ -50,6 +46,18 @@ enum OracleConnectorScope {
         return OracleMetadataDumpFormat.FORMAT_NAME;
       case STATS:
         return "oracle.stats.zip";
+    }
+    throw new AssertionError();
+  }
+
+  private String getResultType() {
+    switch (this) {
+      case LOGS:
+        return "logs";
+      case METADATA:
+        return "metadata";
+      case STATS:
+        return "stats";
     }
     throw new AssertionError();
   }
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
@@ -40,10 +40,14 @@ enum OracleConnectorScope {
     return connectorName;
   }
 
-  String toFileName(boolean isAssessment) {
-    if (this == LOGS && isAssessment) {
-      Clock systemClock = Clock.systemDefaultZone();
-      return ArchiveNameUtil.getFileNameWithTimestamp("oracle-" + resultType, systemClock);
+  String toFileName(boolean isAssessment, Clock clock) {
+    if (this == LOGS) {
+      // add "-logs" twice for consistency with previous versions
+      String name = String.format("oracle-%s-%s", resultType, resultType);
+      if (isAssessment) {
+        return ArchiveNameUtil.getFileNameWithTimestamp(name, clock);
+      }
+      return ArchiveNameUtil.getFileName(name);
     }
     return ArchiveNameUtil.getFileName("oracle-" + resultType);
   }
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
@@ -16,9 +16,10 @@
  */
 package com.google.edwmigration.dumper.application.dumper.connector.oracle;
 
-import com.google.edwmigration.dumper.application.dumper.connector.LogsConnector;
+import com.google.edwmigration.dumper.application.dumper.utils.ArchiveNameUtil;
 import com.google.edwmigration.dumper.plugin.lib.dumper.spi.OracleLogsDumpFormat;
 import com.google.edwmigration.dumper.plugin.lib.dumper.spi.OracleMetadataDumpFormat;
+import java.time.Clock;
 
 enum OracleConnectorScope {
   LOGS("oracle-logs", OracleLogsDumpFormat.FORMAT_NAME, "logs"),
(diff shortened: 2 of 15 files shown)


mjquinn-google (changes requested):
I'm curious to know the reason for making stats a separate connector, as opposed to one or more tasks within the metadata connector?

If it's expensive to dump stats, or if it requires a bunch of separate configuration that is not shared with metadata, I could see this being worthwhile.

But if it's not too expensive, or if there's no additiional configuration to do it, we should probably just do it as a task within the existing metadata connector, for at least two reasons: 1) we don't have to write new readers/ingesters/etc for the new format, 2) customers won't have to provide yet another zip as input, we can just tell them to dump metadata.

mjquinn-google (approved):
Dawid sent me more context on the previous discussion around this, plus Oracle Lite data extraction document, so seems like this is the agreed direction

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  String toFormat() {
+    switch (this) {
+      case LOGS:
+        return OracleLogsDumpFormat.FORMAT_NAME;
+      case METADATA:
+        return OracleMetadataDumpFormat.FORMAT_NAME;
+      case STATS:
+        return "oracle.stats.zip";
```
mjquinn-google, line 48: This looks like it belongs in `OracleLiteDumpFormat` and checked similar to the other cases above?

I'm referencing OracleLiteDumpFormat from the "Oracle Lite data extraction" Google doc.

mjquinn-google (changes requested):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
```diff
@@ -0,0 +1,72 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.edwmigration.dumper.application.dumper.connector.oracle;
+
+import com.google.edwmigration.dumper.application.dumper.connector.LogsConnector;
+import com.google.edwmigration.dumper.plugin.lib.dumper.spi.OracleLogsDumpFormat;
+import com.google.edwmigration.dumper.plugin.lib.dumper.spi.OracleMetadataDumpFormat;
+
+enum OracleConnectorScope {
```
dawidxc, line 23: Optional: I don't have a strong opinion on this, but as already discussed offline, I would follow the existing convention of implementing connectors and avoid any changes in the `OracleMetadataConnector` and `OracleLogsConnector` classes in this PR. For the stats one, it should be enough to implement `Connector` with the `String getDefaultFileName(boolean isAssessment);` method and add new format as a `String` constant in the `OracleStatsConnector` class.

Later, a similar code as in this enum can be introduced as a refactoring across all the connectors.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
```diff
@@ -0,0 +1,72 @@
...
+
+  String toDisplayName() {
+    if (this == METADATA) {
+      return "oracle";
+    } else {
+      return "oracle-" + getResultType();
+    }
+  }
+
+  String toFileName(boolean isAssessment) {
```
dawidxc, line 36: This is duplicated code (comparing with `LogsConnector#getDefaultFileName`). I would suggest introducing a new utils class, moving the method there and calling it from here and from `LogsConnector`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
```diff
@@ -0,0 +1,72 @@
...
+    String timeSuffix = getTimeSuffix(isAssessment);
+    return String.format("dwh-migration-oracle-%s%s.zip", getResultType(), timeSuffix);
+  }
+
+  String toFormat() {
+    switch (this) {
```
dawidxc, line 42: Instead of switching over enum, I would recommend adding a field and initializing it via the constructor. This method would then become a simple getter and no `AssertionError` would be required.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
```diff
@@ -0,0 +1,72 @@
...
+      case METADATA:
+        return OracleMetadataDumpFormat.FORMAT_NAME;
+      case STATS:
+        return "oracle.stats.zip";
+    }
+    throw new AssertionError();
+  }
+
+  private String getResultType() {
+    switch (this) {
```
dawidxc, line 54: Similarly as the previous comment - `resultType` can be a field in the enum.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
```diff
@@ -0,0 +1,72 @@
...
+import com.google.edwmigration.dumper.application.dumper.connector.LogsConnector;
+import com.google.edwmigration.dumper.plugin.lib.dumper.spi.OracleLogsDumpFormat;
+import com.google.edwmigration.dumper.plugin.lib.dumper.spi.OracleMetadataDumpFormat;
+
+enum OracleConnectorScope {
+  LOGS,
+  METADATA,
+  STATS;
+
+  String toDisplayName() {
```
dawidxc, line 28: This method returns the connector name. I would suggest renaming it to `connectorName` or `name`. Also, instead of using `if`, I would suggest adding a field to the enum. Even if there is some redundancy in the values passed as constructor arguments, the overall readability of this class should be improved.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
```diff
@@ -0,0 +1,72 @@
...
+        return "logs";
+      case METADATA:
+        return "metadata";
+      case STATS:
+        return "stats";
+    }
+    throw new AssertionError();
+  }
+
+  private String getTimeSuffix(boolean isAssessment) {
```
dawidxc, line 65: Optional: The method is so simple that you might consider inlining.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsConnector.java
```diff
@@ -0,0 +1,42 @@
...
+import com.google.edwmigration.dumper.plugin.ext.jdk.annotation.Description;
+import java.util.List;
+import javax.annotation.ParametersAreNonnullByDefault;
+
+// TODO: add @AutoService when implementation is completed
+@Description("Dumps aggregated statistics from Oracle")
+@ParametersAreNonnullByDefault
+public class OracleStatsConnector extends AbstractOracleConnector {
+
+  private final StatsTaskListGenerator taskListGenerator = new StatsTaskListGenerator();
```
dawidxc, line 31: It's hard to tell now, when the implementation is not yet ready, but it seems to me that the lifetime of the generator should be bound by the `addTasksTo` method. Unless you already have code that uses it outside of this method, I would suggest moving it into a local variable. It could always be moved to a field, if the scope is broadened.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
```diff
@@ -34,6 +34,39 @@ public class AbstractOracleConnectorTest {
   private final ConnectorArguments arguments = mock(ConnectorArguments.class);
   private final AbstractOracleConnector connector = new OracleMetadataConnector();
 
+  @Test
+  public void getDefaultFileName_notForAssessment_success() {
+    OracleLogsConnector logs = new OracleLogsConnector();
```
dawidxc, line 39: I would strongly suggest to put the test for `OracleLogsConnector` in the `OracleLogsConnectorTest` class. Similarly for other connectors and also other test methods that group different connectors in one test.

However, (in the context of this specific test method) if the method for generating the default filename is extracted to a utils class, then the situation would be simplified, because it would be enough to test this utils class only.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
```diff
@@ -34,6 +34,39 @@ public class AbstractOracleConnectorTest {
   private final ConnectorArguments arguments = mock(ConnectorArguments.class);
   private final AbstractOracleConnector connector = new OracleMetadataConnector();
 
+  @Test
+  public void getDefaultFileName_notForAssessment_success() {
```
dawidxc, line 38: I would suggest adding a test case for assessment also.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
```diff
@@ -34,6 +34,39 @@ public class AbstractOracleConnectorTest {
...
+    OracleLogsConnector logs = new OracleLogsConnector();
+    OracleMetadataConnector metadata = new OracleMetadataConnector();
+    OracleStatsConnector stats = new OracleStatsConnector();
+
+    assertEquals("dwh-migration-oracle-logs.zip", logs.getDefaultFileName(false));
```
dawidxc, line 43: I would suggest adding a parameter name for the `false` argument.

```
logs.getDefaultFileName(/* isAssessment= */ false)
```

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
misolt: You're right about the discrepancy with the doc, thanks for spotting it. Some later offline discussions aren't reflected in the doc. I’ll fill in the missing information in the design and let you know so that we can continue the discussion.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
misolt: I agree, but I would prefer to change this separately, because the for-assessment case occasionally uses system time and needs a refactor before a fake clock can be used.

dawidxc: I'm not sure, how much work that would be, but I would advise to restructure the code in such a way, so that mocking the Clock would be easy right from the start. If this is too much work though, then consider introducing this change in a separate PR.

misolt: The relevant line is from existing code:

https://github.com/google/dwh-migration-tools/blob/main/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/LogsConnector.java#L30

I added a second entry point to this logic, but I consider it existing code. The separate change would be switching from implementing this interface to calling a util method, which imo would be a good time to change the method signature as well.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
misolt: Done, I added the comments.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
dawidxc: > I added a second entry point to this logic, but I consider it existing code.

That is correct. If the code remained untested, this would not be an issue. However, my view on this is that when adding the test for the code dependent on the current time, it needs to be abstracted away, so that the current time can be mocked in tests.

I can see the following options:
1. First new feature, then mocking - i.e. avoid the issue by not writing tests in this PR. Prepare a separate PR with tests and with introducing `Clock` that allows to mock the time.
2. First mocking, then new feature - i.e. prepare a PR that would test the existing logic and base this PR on it.
3. Mix - apply all the changes in this PR.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
misolt: Done, I've split the tests into testing on the enum and testing if connectors have the right enum. They're placed in the enum's and the connector subclasses' test classes respectively.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
misolt: I removed the test that was using this method. The other tests are unrelated to the Clock issue, so they won't be affected by the future refactor.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsConnector.java
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
misolt: I prefer to use this enum for the Oracle connectors. I think it is an improvement compared to implementing the `LogsConnector` and `MetadataConnector` interfaces.

Implementing `Connector` for stats needs to be done through `AbstractOracleConnector`, because methods that we need are encapsulated in various classes of this hierarchy. And since we are making changes to the abstract oracle connector, we're able to use these changes in other Oracle connectors without much effort.

I agree that it would be better to have consistency across all connectors, but I would prefer to make use of this change in code that is currently being changed the most and propagate it to the old connectors later on.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
misolt: The doc and discussion were updated by me and @kpiotrowski, `OracleLiteDumpFormat` is now optional.

The main purpose of adding OracleLiteDumpFormat in Dumper SPI would be to provide the Translation team with a format that they can use to read Dumper outputs. Outputs of this connector won’t be used by Translation, so we can omit this format in the shared definitions.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/ArchiveNameUtil.java
```diff
@@ -0,0 +1,57 @@
...
+
+  /**
+   * Generate the archive file name, but do not include creation time. This is the most common case,
+   * dumping logs for assessment being a known exception.
+   *
+   * @param dwhType The type of the input data source, e.g. "teradata" or "oracle".
+   * @param resultType The type of the resulting dump, e.g. "metadata" or "stats".
+   * @return Full name for the .zip archive.
+   */
+  public static String getBasicFileName(String dwhType, String resultType) {
```
dawidxc, line 48: I would suggest dropping the `basic` word. There would be some symmetry for the existing two methods: `getFileName` and `getFileNameWithTimestamp`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
mjquinn-google: Ok. I still think it would be a good idea to have the format class for consistency, if nothing else. But I leave this one up to you.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/LogsConnector.java
```diff
@@ -16,19 +16,21 @@
...
-    Format format = new SimpleDateFormat("yyyyMMdd'T'HHmmss");
-    String timeSuffix = "-" + format.format(System.currentTimeMillis());
-    return String.format(
-        "dwh-migration-%s-logs%s.zip", getName(), isAssessment ? timeSuffix : StringUtils.EMPTY);
+    if (isAssessment) {
+      Clock clock = Clock.systemDefaultZone();
+      return ArchiveNameUtil.getFileNameWithTimestamp(getName(), "logs", clock);
+    } else {
+      return ArchiveNameUtil.getFileName(getName(), "logs");
+    }
```
kpiotrowski, line 34:
```suggestion
    }
    return ArchiveNameUtil.getFileName(getName(), "logs");
```

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/ArchiveNameUtil.java
```diff
@@ -0,0 +1,57 @@
...
+   * Generate the archive file name that includes creation time. Typically this is used with a logs
+   * connector when dumping for Assessment.
+   *
+   * @param dwhType The type of the input data source, e.g. "teradata" or "snowflake".
+   * @param resultType The type of the resulting dump, e.g. "logs", "account-usage-logs" or
+   *     "raw-logs".
+   * @param clock The Clock instance to provide the date.
+   * @return Full name for the .zip archive.
+   */
+  public static String getFileNameWithTime(String dwhType, String resultType, Clock clock) {
```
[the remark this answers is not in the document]
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
```diff
@@ -0,0 +1,55 @@
...
+  private final String format;
+  private final String resultType;
+
+  OracleConnectorScope(String displayName, String format, String resultType) {
+    this.displayName = displayName;
+    this.format = format;
+    this.resultType = resultType;
+  }
+
+  String toDisplayName() {
```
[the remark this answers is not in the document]
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/ArchiveNameUtil.java
```diff
@@ -0,0 +1,57 @@
...
+   * @param dwhType The type of the input data source, e.g. "teradata" or "oracle".
+   * @param resultType The type of the resulting dump, e.g. "metadata" or "stats".
+   * @return Full name for the .zip archive.
+   */
+  public static String getBasicFileName(String dwhType, String resultType) {
+    return getFileName(dwhType, resultType);
+  }
+
+  private static String getFileName(String dwhType, String details) {
+    return String.format("dwh-migration-%s-%s.zip", dwhType, details);
```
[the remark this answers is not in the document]
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/LogsConnector.java
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
```diff
@@ -0,0 +1,55 @@
...
+  String toFileName(boolean isAssessment) {
+    if (this == LOGS && isAssessment) {
+      Clock systemClock = Clock.systemDefaultZone();
+      return ArchiveNameUtil.getFileNameWithTime("oracle", resultType, systemClock);
+    } else {
+      return ArchiveNameUtil.getBasicFileName("oracle", resultType);
+    }
+  }
+
+  String toFormat() {
```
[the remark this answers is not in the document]
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/ArchiveNameUtil.java
```diff
@@ -0,0 +1,57 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.edwmigration.dumper.application.dumper.utils;
+
+import java.text.SimpleDateFormat;
+import java.time.Clock;
+
+public class ArchiveNameUtil {
```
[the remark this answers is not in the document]
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
```diff
@@ -0,0 +1,58 @@
...
+      if (isAssessment) {
+        return ArchiveNameUtil.getFileNameWithTimestamp("oracle", suffix, clock);
+      }
+      return ArchiveNameUtil.getFileName("oracle", suffix);
+    }
+    return ArchiveNameUtil.getFileName("oracle", resultType);
+  }
+
+  String formatName() {
+    return format;
```
dawidxc, line 56: I would suggest unifying the name for the field and the method - `format` vs. `formatName`.

dawidxc (approved):
Thank you applying all the comments from all the previous reviews!

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
misolt: Done

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/utils/ArchiveNameUtilTest.java
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  @Test
+  public void getFileNameWithTimestamp_success() {
+    Instant instant = Instant.ofEpochMilli(1715346130945L);
+    Clock clock = Clock.fixed(instant, ZoneId.systemDefault());
+    String name = "snowflake-information-schema-logs";
+
+    assertEquals(
+        "dwh-migration-snowflake-information-schema-logs20240510T130210.zip",
```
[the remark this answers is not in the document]
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
misolt: I'd prefer not to add this. I agree that adding it would be more consistent with other connectors, but it would also add clutter to the SPI which is used externally.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/LogsConnector.java
kpiotrowski: looks like `else` statement is still there. It is recommended to skip it alter the `return` statement. Could you please remove it?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  String toFileName(boolean isAssessment, Clock clock) {
+    if (this == LOGS) {
+      // add "-logs" twice for consistency with previous versions
+      String suffix = String.format("%s-%s", resultType, resultType);
+      if (isAssessment) {
+        return ArchiveNameUtil.getFileNameWithTimestamp("oracle", suffix, clock);
+      }
+      return ArchiveNameUtil.getFileName("oracle", suffix);
```
kpiotrowski, line 50: nit: For this kind of simple conditional return statements you could use a ternary operator:

```suggestion
      return isAssessment ? 
        ArchiveNameUtil.getFileNameWithTimestamp("oracle", suffix, clock) :
        ArchiveNameUtil.getFileName("oracle", suffix);
```

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  String connectorName() {
+    return connectorName;
+  }
+
+  String toFileName(boolean isAssessment, Clock clock) {
+    if (this == LOGS) {
+      // add "-logs" twice for consistency with previous versions
+      String suffix = String.format("%s-%s", resultType, resultType);
```
kpiotrowski, line 46: This actually may be redundant because:
1. Assessment classifies ZIP files based on the metadata YAML file and connector name so the matching ZIP file name is not required. (This change was introduced a few month ago)
2. Assessment is the only consumer of query logs ZIP files and it does not support Oracle logs yet.

kpiotrowski (commented):
I left some minor comment, but in general it looks good 👍

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
misolt: Acknowledged, I think in this case neither option is significantly better than the other

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorScope.java
misolt: This might be ok to change, but the change should be made separately. Someone reading the PR description won't expect the change from `oracle-logs-logs` to `oracle-logs` to be included in this one.

kpiotrowski (approved):

mjquinn-google (approved):

PR #385, opened 2024-05-14
[b/337839365] Implement StatsTaskListGenerator, add an example SQL

misolt:
b/337839365

Provide a single example of the new connector being used with a statspack DMA SQL.
- copy the DMA script
- adjust the SQL to work with JDBC
- omit the aggregation with 'avg()', provide basic field values along with DMA deltas
- add custom column names, apply small naming changes
- inline 'vcmdtypes'


misolt:
I think it's good to comment separately on the SQL file added by this PR.
It integrates some of the DB dumping mechanism created by fellow Googlers in [GCP DB Assessment](https://github.com/GoogleCloudPlatform/database-assessment)

The original script can be found [here](https://github.com/GoogleCloudPlatform/database-assessment/blob/99f772fe8a4ba619a3645ae5d119aaa4fa2c5b22/scripts/collector/oracle/sql/extracts/statspack/awrhistcmdtypes.sql).

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/StatsTaskListGenerator.java
```diff
@@ -16,17 +16,64 @@
...
+      this.value = value;
+    }
+  }
+
+  @AutoValue
+  abstract static class StatsQuery {
+
+    abstract String name();
+
+    abstract String tool();
```
dawidxc, line 64: I would suggest changing the type from `String` to the `Tool` enum. This is so that the type-safety is preserved for as long as possible - i.e. until the latest moment, where the conversion to `String` is necessary.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/StatsTaskListGenerator.java
```diff
@@ -16,17 +16,64 @@
...
 @ParametersAreNonnullByDefault
 class StatsTaskListGenerator {
 
   @Nonnull
-  ImmutableList<Task<?>> createTasks(ConnectorArguments arguments) {
-    throw new UnsupportedOperationException();
+  ImmutableList<Task<?>> createTasks(ConnectorArguments arguments) throws IOException {
+    ImmutableList.Builder<Task<?>> builder = ImmutableList.<Task<?>>builder();
+
+    ImmutableList<StatsQuery> queries =
```
dawidxc, line 38: I would suggest extracting this list to a constant.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/StatsTaskListGenerator.java
```diff
@@ -16,17 +16,64 @@
...
+    static StatsQuery create(String name, Tool tool) {
+      String toolName = tool.value;
+      return new AutoValue_StatsTaskListGenerator_StatsQuery(name, toolName);
+    }
+
+    @Nonnull
+    Task<?> toTask() throws IOException {
+      String path = String.format("oracle-stats/%s/%s", tool(), name());
+      URL queryUrl = Resources.getResource(path + ".sql");
+      String query = Resources.toString(queryUrl, Charset.forName("UTF-8"));
```
dawidxc, line 75: I would suggest replacing the expression `Charset.forName("UTF-8")` with the constant

```suggestion
      String query = Resources.toString(queryUrl, UTF_8);
```

https://docs.oracle.com/javase/8/docs/api/java/nio/charset/StandardCharsets.html#UTF_8

dawidxc (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/StatsTaskListGenerator.java
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/StatsTaskListGenerator.java
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/StatsTaskListGenerator.java
```diff
@@ -16,17 +16,64 @@
...
+
+    abstract Tool tool();
+
+    static StatsQuery create(String name, Tool tool) {
+      return new AutoValue_StatsTaskListGenerator_StatsQuery(name, tool);
+    }
+
+    @Nonnull
+    Task<?> toTask() throws IOException {
+      String path = String.format("oracle-stats/%s/%s", tool().value, name());
```
kpiotrowski, line 73: There is also a third group of SQL files - metadata that are neither AWS or STATSPACK data. Those files should be stored in the ZIP root directory. Feel free to address it later.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/StatsTaskListGenerator.java
```diff
@@ -16,17 +16,64 @@
...
+    ImmutableList.Builder<Task<?>> builder = ImmutableList.<Task<?>>builder();
+
+    for (StatsQuery item : QUERIES) {
+      builder.add(item.toTask());
+    }
+    return builder.build();
+  }
+
+  enum Tool {
+    AWR("awr"),
```
kpiotrowski, line 50: User should have an option to choose between AWR and STATSPACK queries depending on what tool is used by the customer. This can be ofc addressed later.

kpiotrowski (approved):
LGTM 👍  I left 2 comment but feel free to address them in a separate PR.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/StatsTaskListGenerator.java
misolt: Acknowledged, it shouldn't be a problem to add a value later, since the Tool enum works just as a wrapper for the strings.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/StatsTaskListGenerator.java
misolt: Thanks, I'll add it later.

I think if it's going to be at most one value for `Tool` per execution, then we can use this information and change things slightly.
The directory could be dropped from the output paths and instead we could use the metadata YAML to inform about which tool was used.

kpiotrowski: > The statspack directory could be dropped from the output paths and instead we could use the metadata YAML to inform about which tool was used

Thanks, I think this would be a great solution. Metadata YAML already includes information about dumper parameters so we can use it.

PR #387, opened 2024-05-17
[b/341297017] Integrate the PDBs version of app_schemas.sql from the DMA tool

misolt:
#### Java changes
Minor fixes discussed earlier:
- rename `Tool` to `Source` to better describe the relationship between Dumper and e.g. AWR
- rename 'other' to 'metadata', since we expect non-AWR, non-Statspack queries to only deal with metadata

Add an item to the tasks list in `StatsTaskListGenerator`.

#### SQL changes
Add a new SQL file representing the PDBs + CDB version of DMA's app_schemas.sql

DMA's app_schemas is an inner SQL script, which not a complete query by itself. It's more similar to a `select_list` term (details in [Oracle SELECT syntax](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/SELECT.html)).
This works in sqlplus, but needed to be changed for JDBC, replacing it with a complete query.

This script was used in two places and #387 represents one of them - inside pdbsinfo.sql. The other usage is simpler, here the `select_list` contains several subqueries, all of which have the same predicate of:

```
outer_query_table.con_id = subquery_table.con_id
```

with the intention to have a NULL value if there are no matches. The easiest way to copy this is as a join (LEFT, since the outer table, which we want to keep, is on the left side).


misolt:
Note that `#` being displayed as a comment character is a mistake on Github UI's part.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/StatsTaskListGenerator.java
```diff
@@ -46,13 +48,15 @@ ImmutableList<Task<?>> createTasks(ConnectorArguments arguments) throws IOExcept
     return builder.build();
   }
 
-  enum Tool {
+  /** The source of performance statistics. */
+  enum Source {
```
dawidxc, line 52: Optional: The `Source` word is very generic and thus easily confused with other concepts. I would suggest a two-word identifier to make it easier for future readers, e.g. `StatsSource`.

dumper/app/src/main/resources/oracle-stats/metadata/app-schemas-pdbs.sql
```diff
@@ -0,0 +1,116 @@
...
+    B.dbid,
+    B.con_id# pdb_id,
+    C.name pdb_name,
+    to_char(B.status) st,
+    to_char(B.flags) logging,
+    B.con_id# con_id,
+    B.con_uid
+  FROM sys.container$ B
+  JOIN sys.obj$ C
+  ON B.obj# = C.obj# AND B.con_id# = 1
```
dawidxc, line 55: Optional: Would quoting help with the rendering issues?

```suggestion
  ON "B.obj#" = "C.obj#" AND "B.con_id#" = 1
```

dawidxc (approved):

dumper/app/src/main/resources/oracle-stats/metadata/app-schemas-pdbs.sql
```diff
@@ -0,0 +1,116 @@
...
+    B.dbid,
+    B.con_id# pdb_id,
+    C.name pdb_name,
+    to_char(B.status) st,
+    to_char(B.flags) logging,
+    B.con_id# con_id,
+    B.con_uid
+  FROM sys.container$ B
+  JOIN sys.obj$ C
+  ON "B.obj#" = "C.obj#" AND "B.con_id#" = 1
```
kpiotrowski, line 55: I tried executing this script with SQL Plus (just to test the results) but it resulted in `"B.con_id#": invalid identifier` exception. It's because double quotes used in `ON` statement. Replacing it with single quotes solves the problem. I recommend changing it to have scripts that can be easily tested with SQL Plus.


```suggestion
  ON 'B.obj#' = 'C.obj#' AND 'B.con_id#' = 1
```

misolt: I posted about a different approach moments ago in here: https://github.com/google/dwh-migration-tools/pull/387#discussion_r1606702137

In the version suggested here, `'B.obj#'` will be a string literal, so the left equals will always be false.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/StatsTaskListGenerator.java
misolt: Done, I renamed it to StatsSource

PR #391, opened 2024-05-22
Expand and improve tests for teradata-logs

misolt:


dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/DriverClasspathTest.java
```diff
@@ -128,7 +129,7 @@ public DummyConnector(@Nonnull String name) {
 
     @Nonnull
     @Override
-    public String getDefaultFileName(boolean ignored) {
+    public String getDefaultFileName(boolean unused, Clock unused2) {
```
dawidxc, line 132: Optional: I would prefer keeping the parameter names from the interface, even if they are unused. Also, in the other places - for example, `MetadataConnector`.

dawidxc (approved):

PR #394, opened 2024-05-24
Cleanup of redundant generics for String

dawidxc:
- String is final, so `? extends String` is not needed.
- Predicates work directly on Strings, so `? super String` is also not needed.
- small fix of logging placeholder and checking string emptiness in `PartitionNameGenerator`

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/JsonResponseFile.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/JsonResponseFile.java
@@ -72,24 +72,24 @@ public class JsonResponseFile {
   }
 
   @Nonnull
-  private static List<? extends String> to_arguments(@Nonnull File file) throws IOException {
+  private static List<String> to_arguments(@Nonnull File file) throws IOException {
     return convert(newObjectMapper().readTree(file));
   }
 
   @VisibleForTesting
-  /* pp */ static List<? extends String> to_arguments(@Nonnull String text)
+  /* pp */ static List<String> to_arguments(@Nonnull String text)
       throws IOException, JsonProcessingException {
     return convert(newObjectMapper().readTree(text));
   }
 
   @Nonnull
-  private static List<? extends String> convert(@Nonnull JsonNode node) {
+  private static List<String> convert(@Nonnull JsonNode node) {
     List<String> out = new ArrayList<>();
     convert(node, out);
     return out;
   }
 
-  private static void convert(@Nonnull JsonNode node, @Nonnull List<? super String> ret) {
+  private static void convert(@Nonnull JsonNode node, @Nonnull List<String> ret) {
     Preconditions.checkArgument(node.isObject(), "Object expected as root node of JSON file.");
     node.fields()
         .forEachRemaining(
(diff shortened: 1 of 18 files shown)


misolt (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/JsonResponseFile.java
```diff
@@ -72,24 +72,24 @@ private static ObjectMapper newObjectMapper() {
   }
 
   @Nonnull
-  private static List<? extends String> to_arguments(@Nonnull File file) throws IOException {
```
paolomorandini, line 75: I would suggest to rewrite `List<? extends T>` to `ImmutableList<T>`  to maintain the guarantee that the original code did provide (i.e. you can't `add` to the list).

misolt: Though adding to such a list is tricky, I think there's no _guarantee_ that it won't happen. For example,
```
<T> void fun(List<T> x) {
  x.add(x.get(0));
}

...

fun(returned list);
```
successfully modifies a (nonempty) `List<? extends String>` object

misolt: > I would suggest to rewrite `List<? extends T>` to `ImmutableList<T>` to maintain the guarantee that the original code did provide (i.e. you can't `add` to the list).

Though adding to such a list is tricky, I think there's no _guarantee_ that it won't happen. For example,

```
<T> void fun(List<T> x) {
  x.add(x.get(0));
}

...

fun(returned list);
```
successfully modifies a (nonempty) `List<? extends String>` object

dawidxc: @paolomorandini I didn't notice this guarantee in the original code. However, the return type should be an immutable collection, whenever possible. If this was the intention behind how the code was written before, then I agree with you that rewriting to `ImmutableList` would maintain this intention (either of immutability or of partial immutability, such as that `add` cannot be called (even though @misolt already provided a counter-example)).

I've applied the change.

dumper/lib-ext-hive-metastore/src/main/java/com/google/edwmigration/dumper/ext/hive/metastore/utils/PartitionNameGenerator.java
```diff
@@ -54,15 +56,14 @@ public static String makePartitionName(List<String> partitionKeys, List<String>
...
             PartitionNameGenerator::constructPartitionName)
         .filter(Objects::nonNull)
-        .collect(Collectors.joining("/"));
+        .collect(joining("/"));
   }
 
   @CheckForNull
   private static String constructPartitionName(String partitionKey, String partitionValue) {
-    if (partitionValue == null || partitionValue.length() == 0) {
+    if (StringUtils.isEmpty(partitionValue)) {
```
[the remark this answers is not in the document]
dawidxc: Done.

shevek-google (commented):
I'm not convinced by the argument for this CL.

This isn't about the finality of String, it's about the compile-time immutability of the resulting collections. List<? extends String>.add(...) will fail at compile time, whereas ImmutableList will fail at runtime.

Also, Predicate<? super String> allows the use of things like Objects::notNull whereas Predicate<String> does not.

PR #398, opened 2024-05-27
[b/343037983] Extract catalogs from HMS

dawidxc:
Catalogs are saved to the JSONL file, obtained by the conversion of each of the catalogs received from Thrift message directly to JSON.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnector.java
@@ -306,7 +306,7 @@ public class HiveMetadataConnector extends AbstractHiveConnector
     }
   }
 
-  private static class CatalogsTask extends AbstractHiveMetadataTask implements FunctionsFormat {
+  private static class CatalogsTask extends AbstractHiveMetadataTask {
 
     private CatalogsTask() {
       super("catalogs.jsonl", unused -> true);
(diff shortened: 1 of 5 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnector.java
```diff
@@ -305,6 +306,36 @@ protected String toCallDescription() {
     }
   }
 
+  private static class CatalogsTask extends AbstractHiveMetadataTask implements FunctionsFormat {
```
misolt, line 309: Will this work using an import or static import instead of the `implements` clause?

The contents of the interface looked like it doesn't need to be implemented here. There could be other reasons, e.g. `instanceof` or `AutoService`, but if this is not the case I would suggest to use an import instead.

dumper/lib-ext-hive-metastore/src/main/java/com/google/edwmigration/dumper/ext/hive/metastore/HiveMetastoreThriftClient_Superset.java
```diff
@@ -498,4 +500,28 @@ public void close() throws IOException {
       throw new IOException("Unable to shutdown Thrift client.", e);
     }
   }
+
+  @Override
+  public ImmutableList<String> getCatalogsAsJsonl() throws TException {
+    TSerializer serializer = new TSerializer(new TSimpleJSONProtocol.Factory());
+    com.google.edwmigration.dumper.ext.hive.metastore.thrift.api.superset.GetCatalogsResponse
+        catalogs = client.get_catalogs();
```
misolt, line 508: I recommend adding an import and dropping the package prefix here as well as in lines 521 and 524.

misolt (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnector.java
dawidxc: It was a leftover from copying a different task. Thanks for catching that. Removed.

dumper/lib-ext-hive-metastore/src/main/java/com/google/edwmigration/dumper/ext/hive/metastore/HiveMetastoreThriftClient_Superset.java
dawidxc: Done.

PR #399, opened 2024-05-27
[b/343047614] Extract databases from HMS

dawidxc:
Extract all the fields from the databases and save the result in the new `databases.jsonl` file that contains a JSON serialization of the Thrift response.

### DIFF
--- a/dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnectorTest.java
+++ b/dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnectorTest.java
@@ -51,8 +51,11 @@ import org.apache.commons.lang3.SystemUtils;
 import org.junit.Assert;
 import org.junit.Assume;
 import org.junit.Test;
+import org.junit.experimental.theories.DataPoints;
+import org.junit.experimental.theories.FromDataPoints;
+import org.junit.experimental.theories.Theories;
+import org.junit.experimental.theories.Theory;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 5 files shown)


kpiotrowski (approved):
LGTM, but please check commends in the other PRs introducing similar changes.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnectorTest.java
```diff
@@ -51,16 +51,19 @@
...
+import org.junit.experimental.theories.DataPoints;
+import org.junit.experimental.theories.FromDataPoints;
+import org.junit.experimental.theories.Theories;
+import org.junit.experimental.theories.Theory;
 import org.junit.runner.RunWith;
-import org.junit.runners.JUnit4;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
```
misolt, line 62: Optional: This comment can be removed. It doesn't document the class, it only tells the reader how to use gradle.

misolt (approved):

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnectorTest.java
dawidxc: I find it helpful, because it provides the full command for using Java 8, so I would prefer to keep it. However, since the constraint about Java version applies to a single test only, I've moved the comment to this single test method.

PR #400, opened 2024-05-27
[b/343047614] Hide migration metadata extraction behind a flag

dawidxc:
Do not extract migration metadata by default, so that the discovery for assessment can be used without additional metadata required by migration.

For migration case require the `-Dhiveql.migration.metadata=true` flag.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnector.java
@@ -46,6 +46,7 @@ import java.util.List;
 import java.util.function.Predicate;
 import javax.annotation.Nonnull;
 import org.apache.commons.csv.CSVPrinter;
+import org.apache.commons.lang3.BooleanUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -384,8 +385,11 @@ public class HiveMetadataConnector extends AbstractHiveConnector
     out.add(new SchemataTask(databasePredicate));
     out.add(new TablesJsonTask(databasePredicate, shouldDumpPartitions));
     out.add(new FunctionsTask(databasePredicate));
-    out.add(new CatalogsTask());
-    out.add(new DatabasesJsonlTask());
+    if (BooleanUtils.toBoolean(
+        arguments.getDefinitionOrDefault(HiveConnectorProperty.MIGRATION_METADATA))) {
+      out.add(new CatalogsTask());
+      out.add(new DatabasesJsonlTask());
+    }
 
     if (arguments.isAssessment()) {
       out.add(new DatabasesTask(databasePredicate));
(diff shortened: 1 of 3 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnector.java
```diff
@@ -384,8 +385,11 @@ public void addTasksTo(List<? super Task<?>> out, @Nonnull ConnectorArguments ar
     out.add(new SchemataTask(databasePredicate));
     out.add(new TablesJsonTask(databasePredicate, shouldDumpPartitions));
     out.add(new FunctionsTask(databasePredicate));
-    out.add(new CatalogsTask());
-    out.add(new DatabasesJsonlTask());
+    if (BooleanUtils.toBoolean(
```
kpiotrowski, line 388: What is a reason for omitting metadata extraction by default? Does it cause any issues or is time consuming?

This change is not backward compatible because it changes the default tool behavior. I'd recommend avoiding such changes to the existing connectors, but this may be fine if we are sure this is not used by the customers.

kpiotrowski (approved):

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnectorTest.java
```diff
@@ -76,14 +76,43 @@ public void testConnector() throws Exception {
...
-  @DataPoints("expectedTaskNames")
+  @DataPoints("commonTaskNames")
   public static final ImmutableList<String> EXPECTED_TASK_NAMES =
-      ImmutableList.of("catalogs.jsonl", "databases.jsonl");
+      ImmutableList.of(
+          "compilerworks-metadata.yaml",
+          "compilerworks-format.txt",
+          "schemata.csv",
+          "tables.jsonl",
+          "functions.csv");
```
misolt, line 86: Do these names represent all tasks that are added by the Hive connector? Or is there a task which is added by the connector, but doesn't have its name listed here?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnector.java
dawidxc: The extraction may take a long time, especially for tables and partitions it doesn't make sense to extract them twice. This is not used by customers yet, so for now I'm introducing a boolean flag for testing purposes. Later we can change this.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnectorTest.java
dawidxc: The list of all the tasks depends on the value of the flag `-Dhiveql.migration.metadata=`. This list represents the names of the tasks that are always extracted no matter what the value of the flag is. The second list (i.e. `migrationMetadataTaskNames`) has all the tasks with the flag enabled.

shevek-google (commented):
My read of this is that it makes the dumper not work for migration by default. Can we please discuss this as a matter of urgency?

PR #401, opened 2024-05-27
[b/343058608] Extract master keys from HMS

dawidxc:
Extract master keys and save them in the `master-keys.csv` file.


kpiotrowski (approved):
LGTM, but this is the next PR that is duplicating the similar Task definition logic. I'd recommend to introduce more generic class. This can be done as a separate task after merging all PRs.

misolt:
I see that `migration-metadata-flag` is the destination for this merge. Are there any consequences for `main`?
AFAIK this shouldn't require a PR, the merge could be done locally and pushed later.

dawidxc:
@kpiotrowski To be honest I don't see a lot of potential to reduce code duplication. I will analyze the duplication after merging the PR chain and try to come up with some solution if possible.

@misolt I've created a PR chain with a few PRs dependent on each other, so that they are easier to review. After merging the first one to the `main` branch, the destination of the second is automatically changing to `main`. Then after merging the second one, the same happens for the third etc.

PR #404, opened 2024-05-27
[b/343068084] Extract resource plans from HMS

dawidxc:
Extract resource plans and save them in the `resource-plans.jsonl` file.


kpiotrowski (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnector.java
```diff
@@ -135,6 +138,39 @@ protected String toCallDescription() {
     }
   }
 
+  private static class DatabasesJsonlTask extends AbstractHiveMetadataTask {
+
+    private DatabasesJsonlTask() {
+      super("databases.jsonl", unused -> true);
```
misolt, line 144: The 2nd parameter equivalent to `Predicates.alwaysTrue` shouldn't be necessary here.
My suggestion is to either add an alternative `AbstractHiveMetadataTask` constructor which doesn't set `predicate` or, preferably, add a trivial `isIncludedDatabase` implementation to `AbstractHiveTask` and inherit from it directly where possible.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnector.java
```diff
@@ -477,6 +477,39 @@ protected String toCallDescription() {
     }
   }
 
+  private static class ResourcePlansJsonlTask extends AbstractHiveMetadataTask {
```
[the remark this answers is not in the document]
dawidxc: I've added `HiveJsonlTask` and I've used it in all the `*JsonlTask` classes.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnector.java
dawidxc: I've added a second constructor. Filtering by database makes sense only in the context of metadata.

However, currently there are no other classes extending `AbstractHiveTask` besides `AbstractHiveMetadataTask`, so possibly these two could be merged.

PR #405, opened 2024-05-27
[b/343069385] Extract tables from HMS

dawidxc:
Extract tables in the new format to the `tables-raw.jsonl` file.

The file contains the JSON serialization of the raw response from Hive Metastore API, each table on a separate line.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnector.java
@@ -291,6 +291,75 @@ public class HiveMetadataConnector extends AbstractHiveConnector
     }
   }
 
+  private static class TablesRawJsonlTask extends AbstractHiveMetadataTask {
+
+    private TablesRawJsonlTask(Predicate<String> databasePredicate) {
+      super("tables-raw.jsonl", databasePredicate);
+    }
+
+    @Override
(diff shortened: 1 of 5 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnector.java
```diff
@@ -500,6 +570,9 @@ public void addTasksTo(List<? super Task<?>> out, @Nonnull ConnectorArguments ar
       out.add(new DelegationTokensTask());
       out.add(new FunctionsJsonlTask());
       out.add(new ResourcePlansJsonlTask());
+      out.add(new TablesRawJsonlTask(databasePredicate));
```
kpiotrowski, line 573: Is there any reason for not including both `TablesJsonTask` and `TablesRawJsonlTask`? My concern is that ZIP extracted with `MIGRATION_METADATA` property won't include `tables.jsonl` file that is required for Hive assessment. User will have to run Dumper twice to get this data.

Do you plan to adjust assessment loader to load data from both files?

dawidxc: The flag is temporary. After the development is finished and the new input tested I plan to remove the flag and extract everything that is hidden behind the flag now.

Long-term we can adjust the loader and drop the old `TablesJsonTask`, but this is lower priority at the moment.

kpiotrowski (approved):

PR #418, opened 2024-06-06
Use BooleanUtils for parsing boolean flag

dawidxc:


dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/BooleanValueConverterTest.java
```diff
@@ -0,0 +1,59 @@
...
+import org.junit.runner.RunWith;
+
+@RunWith(Theories.class)
+public class BooleanValueConverterTest {
+
+  @DataPoints("testCases")
+  public static ImmutableList<TestCase> testCases =
+      ImmutableList.of(
+          TestCase.create("", false),
+          TestCase.create("&*%^$*&@#*^**$", false),
```
shevek-google, line 37: Is one of the effects of this CL to change the parsing of an invalid value to be a silent false, rather than an error?

misolt:
From `toBoolean(String)` documentation:
```
'true', 'on', 'y', 't' or 'yes' (case insensitive) will return true. Otherwise, false is returned.
```
This means `1` used to be valid and true, but now it will be valid and false

misolt:
Is there a context in the project where it's necessary to parse a boolean flag?

At first glance it seems like we should be able to use `OptionSpec<Void>` to replace any `OptionSpec<Boolean>` that has a default value.

dawidxc:
@misolt 

> This means "1" used to be valid and true, but now it will be valid and false

"1" parses as `true`. I've added a test case. Maybe you've checked the documentation for a different version of the method.

> Is there a context in the project where it's necessary to parse a boolean flag?

I've used `--hive-metastore-dump-partition-metadata` recently. It seems it's the only flag that uses `BooleanValueConverter`. I haven't checked if there aren't other boolean flags.

> At first glance it seems like we should be able to use OptionSpec<Void> to replace any OptionSpec<Boolean> that has a default value.

Yes, it might be a good idea to replace the boolean flag and remove the `BooleanValueConverter` class.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/DefaultArguments.java
```diff
@@ -76,12 +62,7 @@ public Class<? extends Boolean> valueType() {
 
     @Override
     public String valuePattern() {
-      StringBuilder buf = new StringBuilder();
-      Joiner joiner = Joiner.on('/');
-      joiner.appendTo(buf, V_TRUE);
-      buf.append('/');
-      joiner.appendTo(buf, V_FALSE);
-      return buf.toString();
+      return "true/t/yes/y/1/false/f/no/n/0";
```
paolomorandini, line 65: If possible get true and false values from the utility class itself, if that's not provided by the utility consider keeping the `V_TRUE` and `V_FALSE` arrays and having a test to confirm every value in `V_TRUE` parses as `true` (same for `V_FALSE`).

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/BooleanValueConverterTest.java
dawidxc: Yes, but after thinking about it again, I think I should preserve the existing behavior. Fixed.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/DefaultArguments.java
dawidxc: You're right. It turns out that changing the parsing to the library call causes more trouble than the benefit of shortening the code, which was my original intent. It made some sense to keep the sample supported values in the `toString` method, but introducing the same values as the library supports and keeping them in sync promises to be a much bigger task. The existing code will be easier to maintain than my proposed change.

I'm closing the PR as it now makes little sense to me to continue working on this.

PR #421, opened 2024-06-06
Use Jackson to generate the JSON objects and arrays.

dawidxc:


shevek-google (commented):
Just an observation that we mostly use Jackson APIs, rather than GSON. Not a CR, though.

misolt (approved):

dawidxc:
> Just an observation that we mostly use Jackson APIs, rather than GSON. Not a CR, though.

@shevek-google I have changed the implementation from GSON to Jackson.

misolt (approved):

shevek-google (commented):
Thank you, this is elegant.

PR #422, opened 2024-06-06
Address concurrency issues in TaskSetState

misolt:
TaskSetState has a synchronization mechanism, which intends to provide mutual exclusion when accessing the private `resultMap`. However, the public method:
```
getTaskResultMap()
```
returns `resultMap` to the caller, which makes such synchronization ineffective, as the caller is able to read and write to this map freely.

The primary goal of this PR is to replace this approach with methods that have correct synchronization.

- replace getTaskResultMap with safe methods
- add classes representing return types needed by the new methods
- in the newly added methods, discourage accessing them directly from `Impl`
- inline a redundant method, remove unchecked cast
- fix a variable name referring to "tasksNotSucceeded", when it actually represents only the tasks that ended in failure
- remove the redundant `lock` object

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
@@ -79,6 +79,7 @@ public interface TaskSetState {
 
     private final Map<Task<?>, TaskResult<?>> resultMap = new HashMap<>();
 
+    @Override
     public synchronized long failedRequiredTaskCount() {
       long result = 0;
       for (Task<?> key : resultMap.keySet()) {
@@ -90,6 +91,8 @@ public interface TaskSetState {
       return result;
     }
 
+    @Nonnull
+    @Override
     public synchronized ImmutableList<TaskResultSummary> taskResultSummaries() {
       ImmutableList.Builder<TaskResultSummary> builder = ImmutableList.builder();
       resultMap.forEach(
@@ -100,6 +103,8 @@ public interface TaskSetState {
       return builder.build();
     }
 
+    @Nonnull
+    @Override
     public synchronized ImmutableList<TasksReport> tasksReports() {
       ImmutableList.Builder<TasksReport> builder = ImmutableList.builder();
       resultMap.values().stream()
@@ -124,6 +129,14 @@ public interface TaskSetState {
       resultMap.put(task, new TaskResult<>(state, exception));
     }
   }
+
+  long failedRequiredTaskCount();
+
+  @Nonnull
+  ImmutableList<TaskResultSummary> taskResultSummaries();
+
+  @Nonnull
+  ImmutableList<TasksReport> tasksReports();
 
   @CheckForNull
   public <T> TaskResult<T> getTaskResult(@Nonnull Task<T> task);
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
@@ -80,13 +80,6 @@ public interface TaskSetState {
 
     private final Object lock = new Object();
 
-    @Nonnull
-    public Map<Task<?>, TaskResult<?>> getTaskResultMap() {
-      synchronized (lock) {
-        return resultMap;
-      }
-    }
-
     public long failedRequiredTaskCount() {
       synchronized (lock) {
         return resultMap.entrySet().stream()
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
@@ -129,8 +129,8 @@ public interface TaskSetState {
       resultMap.put(task, new TaskResult<>(state, value));
     }
 
-    public synchronized <T> void setTaskException(
-        @Nonnull Task<T> task, @Nonnull TaskState state, @CheckForNull Exception exception) {
+    public synchronized void setTaskException(
+        @Nonnull Task<?> task, @Nonnull TaskState state, @CheckForNull Exception exception) {
       resultMap.put(task, new TaskResult<>(state, exception));
     }
   }
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
@@ -16,6 +16,9 @@
  */
 package com.google.edwmigration.dumper.application.dumper.task;
 
+import static com.google.edwmigration.dumper.application.dumper.task.TaskCategory.REQUIRED;
+import static com.google.edwmigration.dumper.application.dumper.task.TaskState.FAILED;
+
 import com.google.common.collect.ImmutableList;
 import java.util.HashMap;
 import java.util.Map;
@@ -77,10 +80,14 @@ public interface TaskSetState {
     private final Map<Task<?>, TaskResult<?>> resultMap = new HashMap<>();
 
     public synchronized long failedRequiredTaskCount() {
-      return resultMap.entrySet().stream()
-          .filter(e -> TaskCategory.REQUIRED.equals(e.getKey().getCategory()))
-          .filter(e -> TaskState.FAILED.equals(e.getValue().getState()))
-          .count();
+      long result = 0;
+      for (Task<?> key : resultMap.keySet()) {
+        TaskState state = resultMap.get(key).getState();
+        if (key.getCategory() == REQUIRED && state == FAILED) {
+          result++;
+        }
+      }
+      return result;
     }
 
     public synchronized ImmutableList<TaskResultSummary> taskResultSummaries() {
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
@@ -22,6 +22,7 @@ import static com.google.edwmigration.dumper.application.dumper.task.TaskState.F
 import static java.util.stream.Collectors.counting;
 import static java.util.stream.Collectors.groupingBy;
 
+import com.google.auto.value.AutoValue;
 import com.google.common.collect.ImmutableList;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


paolomorandini (commented):
- Consider reinstating `@GuardedBy("this")` if using synchronized methods
- Locking on a private object is a common pattern, but the advantages are probably not much relevant in this case
- Consider introducing smaller `synchronized` blocks

Could this also be resolved by having the getter return an immutable copy of the map?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
```diff
@@ -16,60 +16,135 @@
...
 import java.util.Map;
+import java.util.stream.Collectors;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
+import javax.annotation.ParametersAreNonnullByDefault;
 import javax.annotation.concurrent.GuardedBy;
 import net.jcip.annotations.ThreadSafe;
 
 /** @author shevek */
 public interface TaskSetState {
```
misolt, line 33: @paolomorandini Let's continue the discussion here, because github won't allow in-thread replies unless a comment is attached to a line of code.

- I've reinstated `@GuardedBy` with `this` as suggested. The annotation doesn't seem to work in my setup, but if it works for some people I don't mind keeping it.
- I don't have an opinion on this pattern in general, I agree that in this case it's not necessary
- There is room for improvement of the performance of this class, including less work within `synchronized` blocks, but I think it would be better to do this separately from the current PR and only if we have flows with a sufficiently large number of Tasks.

Returning a copy of the map would work, but I don't want to make the map part of the api, because it will prevent future optimizations. For example, `failedRequiredTaskCount` could be optimized to constant-time if we track it separately, but returning a copy of the whole map will always be linear.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
```diff
@@ -16,60 +16,135 @@
...
     @Override
-    @SuppressWarnings("unchecked")
-    public <T> TaskResult<T> getTaskResult(@Nonnull Task<T> task) {
-      synchronized (lock) {
-        return (TaskResult<T>) resultMap.get(task);
-      }
+    public synchronized ImmutableList<TasksReport> tasksReports() {
+      ImmutableList.Builder<TasksReport> builder = ImmutableList.builder();
+      resultMap.values().stream()
+          .collect(Collectors.groupingBy(TaskResult::getState, Collectors.counting()))
```
dawidxc, line 117: I would suggest static imports for `groupingBy` and `counting`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
```diff
@@ -16,60 +16,135 @@
...
+          .forEach((key, value) -> builder.add(new TasksReport(value, key)));
```
dawidxc, line 118: I would recommend creating a second stream and collecting to list instead of using the builder.

```
.entrySet().stream().map(entry -> new TaskReport(...)).collect(toImmutableList());
```

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
```diff
@@ -16,60 +16,135 @@
...
+    @Deprecated // Use TaskSetState instead of TaskSetState.Impl
     @Nonnull
-    public Map<Task<?>, TaskResult<?>> getTaskResultMap() {
-      synchronized (lock) {
-        return resultMap;
-      }
+    @Override
+    public synchronized ImmutableList<TaskResultSummary> taskResultSummaries() {
+      ImmutableList.Builder<TaskResultSummary> builder = ImmutableList.builder();
+      resultMap.forEach(
```
dawidxc, line 103: I would recommend using a stream over entrySet as in the other comment.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
```diff
@@ -16,60 +16,135 @@
...
-    @GuardedBy("lock")
+    @GuardedBy("this")
     private final Map<Task<?>, TaskResult<?>> resultMap = new HashMap<>();
 
-    private final Object lock = new Object();
+    @Deprecated // Use TaskSetState instead of TaskSetState.Impl
+    @Override
+    public synchronized long failedRequiredTaskCount() {
+      long result = 0;
+      for (Task<?> key : resultMap.keySet()) {
```
dawidxc, line 89: If both keys and values are needed, then I would suggest iterating over `entrySet`, not `keySet`, so that the reference to `resultMap` is not needed inside the loop.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
```diff
@@ -16,60 +16,134 @@
...
+      this.count = count;
+      this.state = state;
+    }
+  }
+
   @ThreadSafe
   public static class Impl implements TaskSetState {
 
-    @GuardedBy("lock")
+    @GuardedBy("this")
```
shevek-google, line 84: One should never expose one's mutexes, they should always be private objects. Exposing a mutex makes it non-private, which means someone else can lock on it, and cause deadlock.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
```diff
@@ -16,60 +16,134 @@
...
-  @CheckForNull
-  public <T> TaskResult<T> getTaskResult(@Nonnull Task<T> task);
+  long failedRequiredTaskCount();
 
   @Nonnull
-  public default TaskState getTaskState(@Nonnull Task<?> task) {
-    TaskResult<?> result = getTaskResult(task);
-    return (result == null) ? TaskState.NOT_STARTED : result.getState();
-  }
+  ImmutableList<TaskResultSummary> taskResultSummaries();
```
shevek-google, line 142: Should these be 'get' methods?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
```diff
@@ -16,60 +16,134 @@
...
 
-  @CheckForNull
-  public <T> TaskResult<T> getTaskResult(@Nonnull Task<T> task);
+  long failedRequiredTaskCount();
 
   @Nonnull
-  public default TaskState getTaskState(@Nonnull Task<?> task) {
-    TaskResult<?> result = getTaskResult(task);
-    return (result == null) ? TaskState.NOT_STARTED : result.getState();
-  }
```
shevek-google, line 74: Why not a default method?

shevek-google (dismissed):
I'm not sure any concurrency issue is actually being addressed here.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
misolt: `getTaskState` refers to `resultMap`, so a non-default method is needed to do this part of the job.

If it's beneficial, I can extract the 2nd part of `getTaskState` to a separate method and make it `default` in `TaskSetState'.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
misolt: Done, I added a `get-` prefix to the new methods.

dawidxc (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
```diff
@@ -16,60 +16,134 @@
...
-    public <T> TaskResult<T> getTaskResult(@Nonnull Task<T> task) {
-      synchronized (lock) {
-        return (TaskResult<T>) resultMap.get(task);
-      }
+    public synchronized ImmutableList<TasksReport> getTasksReports() {
+      return resultMap.values().stream()
+          .collect(groupingBy(TaskResult::getState, counting()))
+          .entrySet()
+          .stream()
+          .map((entry) -> new TasksReport(entry.getValue(), entry.getKey()))
```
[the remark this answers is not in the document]
misolt: Done, I removed the parentheses

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
```diff
@@ -16,60 +16,134 @@
...
+    @Deprecated // Use TaskSetState instead of TaskSetState.Impl
     @Nonnull
-    public Map<Task<?>, TaskResult<?>> getTaskResultMap() {
-      synchronized (lock) {
-        return resultMap;
-      }
+    @Override
+    public synchronized ImmutableList<TaskResultSummary> getTaskResultSummaries() {
+      return resultMap.entrySet().stream()
+          .map((entry) -> TaskResultSummary.create(entry.getKey(), entry.getValue()))
```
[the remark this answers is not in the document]
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
```diff
@@ -16,60 +16,134 @@
...
+    private TaskResultSummary(String exception, Task<?> task, TaskState state) {
+      this.exception = exception;
+      this.task = task;
+      this.state = state;
+    }
+
+    @Nonnull
+    static TaskResultSummary create(Task<?> task, TaskResult<?> result) {
+      String optionalException = String.format(": %s", result.getException());
+      String exceptionDescription = result.getException() == null ? "" : optionalException;
```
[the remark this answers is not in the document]
misolt: Done, I moved this logic inside `toString`. I used a similar approach as in the 1st alternative, but I kept `String.format` for the initial, fixed step. The `exception` field now just holds the Throwable instead of a String.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
```diff
@@ -16,60 +16,134 @@
...
+
+    @Nonnull
+    @Override
+    public String toString() {
+      return String.format("Task %s (%s) %s%s", state, task.getCategory(), task, exception);
+    }
+  }
+
+  @ParametersAreNonnullByDefault
+  public static class TasksReport {
```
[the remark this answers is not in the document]
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskSetState.java
misolt: The hypothetical is that two threads access some `TaskSetState` instance `x` simultaneously. One enters a block, which is `synchronized (x) {...`, while the other calls `x`'s member method, whose implementation requires a mutex.

The suggestion is that concurrent access should be allowed for these threads. The PR opts to enforce the mutex.

It's hard to be sure about the intention of someone writing `synchronized (x) {...` in this context, but it's unlikely that they would expect a different thread operating on critical sections of `x` at the same time through a different mutex.

In the deadlock example, replacing the call to a `synchronized` method on `x` with a `synchronized (x) {...` block leads to the same behavior. The risk is inherent to `synchronized` blocks which depend on some shared object, adding the method doesn't increase it.

PR #429, opened 2024-06-13
Handle null ResultSet

misolt:
### handle null resultset in AbstractJdbcTask

This is on hold, to be merged only if we confirm that null ResultSet's need to be handled.

This PR deals with the case where a null `ResultSet` is returned. Handling of this case seems to be missing, the Task fails with NPE. Add a null check and make Dumper write nothing if the `ResultSet` is null.

### remove try-with-resources for RecordProgressMonitor
[reverted]

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
@@ -113,7 +113,7 @@ public abstract class AbstractJdbcTask<T> extends AbstractTask<T> {
   public ResultSetExtractor<Summary> newCsvResultSetExtractor(@Nonnull ByteSink sink) {
     return resultSet -> {
       RecordProgressMonitor monitor = new RecordProgressMonitor(getName());
-      printWithMonitor(sink, monitor, resultSet);
+      return printWithMonitor(sink, monitor, resultSet);
     };
   }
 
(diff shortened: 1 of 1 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractJdbcTask.java
```diff
@@ -111,29 +111,38 @@ protected CSVFormat newCsvFormat(@Nonnull ResultSet rs) throws SQLException {
...
+      throws SQLException {
+    try {
+      if (resultSet != null) {
+        printAllResults(sink, resultSet, monitor);
+      }
+      return new Summary(monitor.getCount());
+    } catch (IOException e) {
+      throw new SQLException(e);
+    } finally {
+      monitor.close();
```
dawidxc, line 142: I don't see the benefit of not using `monitor` in the `try-with-resources` statement. I don't think there needs to be an external resource to justify the usage of `try-with-resources`. It can be considered a syntax sugar for calling `close` in `finally` - I liked the previous version more.

As for the other two issues that are being addressed with the PR:
- the null check can be added directly inside `printAllResults`,
- the exception handling can also be moved to the inside of `printAllResults`.

misolt: As discussed:
- I removed the `finally` block, reverting to try-with-resources
- the new method can stay, it decreases the complexity of `printAllResults` code

dawidxc (approved):

shevek-google (commented):
There's a minor code-smell here: A query should always return a nonnull ResultSet if it returns a result. What query did we execute that doesn't return a ResultSet, and what did we intend by that query?

Normative reference: https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#getResultSet--

misolt:
@shevek-google This is a good point. I didn't merge this after the initial approval, because the query that caused the null ResultSet was incorrect.

For now, I'll wait with the PR. if it turns out that this change is needed, we can just merge. Otherwise this can be closed.

If we close, I'd still add error handling for the null ResultSet. It would be better to say `"Incorrect query type: <query text>"` rather than `java.lang.NullPointerException: null at lambda$newCsvResultSetExtractor$0`

I'll also update the description in a moment

shevek-google:
I agree, thank you. It's probably very important to get earlier in the chain and diagnose (remotely) WHY a query returned a null ResultSet rather than ignoring it silently. I think sometimes we get so focused on "make the exception stop happening" that we forget the bigger picture.

PR #433, opened 2024-06-17
[b/341297017] Add table-types-dtl.sql

misolt:
part of b/341297017

### DIFF
--- a/dumper/app/src/main/resources/oracle-stats/native/table-types-dtl.sql
+++ b/dumper/app/src/main/resources/oracle-stats/native/table-types-dtl.sql
@@ -1,0 +1,89 @@
+-- Copyright 2022-2024 Google LLC
+-- Copyright 2013-2021 CompilerWorks
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+SELECT
+  D.con_id "ConId",
+  D.owner "Owner",
+  D.table_name "TableName",
+  D.partitioned "Partitioned",
+  D.iot_type "IotType",
+  D.nested "Nested",
+  D.temporary "Temporary",
+  D.secondary "Secondary",
+  D.cluster_name "ClusterName",
+  D.source "Source",
+  E.partitioning_type "PartitioningType",
+  E.subpartitioning_type "SubpartitioningType",
+  E.partition_count "PartitionCount",
+  F.count "SubpartitionCount"
+FROM (
+  SELECT
+    A.con_id,
+    A.owner,
+    A.table_name,
+    A.partitioned,
+    A.iot_type,
+    A.nested,
+    A.temporary,
+    A.secondary,
+    A.cluster_name,
+    'TAB' source
+  FROM cdb_tables A
+  WHERE A.owner <> 'SYSTEM' AND A.owner NOT LIKE '%SYS'
+  UNION ALL SELECT
+    B.con_id,
+    B.owner,
+    B.table_name,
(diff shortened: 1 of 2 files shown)


dumper/app/src/main/resources/oracle-stats/native/table-types-dtl.sql
```diff
@@ -0,0 +1,89 @@
...
+    A.partitioned,
+    A.iot_type,
+    A.nested,
+    A.temporary,
+    A.secondary,
+    A.cluster_name,
+    'TAB' source
+  FROM cdb_tables A
+  WHERE A.owner <> 'SYSTEM' AND A.owner NOT LIKE '%SYS'
+  UNION ALL SELECT
```
dawidxc, line 44: Optional: It looks like there is discrepancy in whether the query that follows the `UNION ALL` is added inside parentheses or not. Compare with:

https://github.com/google/dwh-migration-tools/blob/f3cd28ba5345ff0c8a8c19eface048772525776e/dumper/app/src/main/resources/oracle-stats/native/db-objects.sql#L29

I would suggest choosing just one style (either with parentheses or without) and applying it to all scripts.

dumper/app/src/main/resources/oracle-stats/native/table-types-dtl.sql
```diff
@@ -0,0 +1,89 @@
...
+    C.table_name,
+    C.partitioned,
+    C.iot_type,
+    C.nested,
+    C.temporary,
+    C.secondary,
+    C.cluster_name,
+    'OBJ' source
+  FROM cdb_object_tables C
+  WHERE C.owner <> 'SYSTEM' AND C.owner NOT LIKE '%SYS'
```
dawidxc, line 69: The `WHERE` condition is repeated 3 times. Would it make sense to reduce it to one (union first, filter later)?

dawidxc (approved):

dumper/app/src/main/resources/oracle-stats/native/table-types-dtl.sql
```diff
@@ -0,0 +1,89 @@
...
+    A.table_name,
+    A.partitioned,
+    A.iot_type,
+    A.nested,
+    A.temporary,
+    A.secondary,
+    A.cluster_name,
+    'TAB' source
+  FROM cdb_tables A
+  WHERE A.owner <> 'SYSTEM' AND A.owner NOT LIKE '%SYS'
```
kpiotrowski, line 43: I think we should be consistent in all the scripts and decide to exclude SYSTEM objects or not. This filtering is included in this PR but it was not in the previous one that added db_objects and used_space_details.

I'd suggest to simplify the dumper code and remove the filters. We can implement it in the assessment metadata postprocessor.

kpiotrowski (approved):

dumper/app/src/main/resources/oracle-stats/native/table-types-dtl.sql
misolt: It could be done, but I don't see a way to do this without adding another nesting level.

dawidxc: What's the downside of another nesting level?

dumper/app/src/main/resources/oracle-stats/native/table-types-dtl.sql
misolt: Done, the filters reduce the output size by about 1/3, so it's not a very big difference.

Filtering is a responsibility of Assessment, Dumper can do it where it's not practical to skip the filter, e.g. the results would be incorrect or larger by an order of magnitude.

dumper/app/src/main/resources/oracle-stats/native/table-types-dtl.sql
misolt: I usually avoid the parentheses in those queries if I see that they can be avoided. I'll try to remove the parentheses in db-objects on next change.

dumper/app/src/main/resources/oracle-stats/native/table-types-dtl.sql
misolt: It makes it slightly harder to follow the control flow. It's not a big difference, but I think that the effect on overall readability is similar to repeating this condition.

Note that this condition is now simplified, because `owner <> 'SYSTEM'` was removed.

PR #434, opened 2024-06-17
[b/342096412] Detect Java version in the main script

dawidxc:
The `dwh-migration-dumper` script worked for Java version 9 and above. This change makes it compatible with Java 8 also.

The `dwh-migration-dumper-java8` script was a special script for Java 8 support. It is no longer needed, therefore it is removed.

The Java detection is applied in Linux and Windows environments.

The new scripts in the bin directory behave the same in the source and binary releases (see the requirement in the "Binary vs Source Release" section of the Contributing guide: https://github.com/google/dwh-migration-tools/blob/c1dc8b5442230d455fb5c4f988b57e23d4f69e64/dumper/CONTRIBUTING.md#binary-vs-source-release).

This is a follow-up of PR #364.

### DIFF
--- a/bin/dwh-migration-client
+++ b/bin/dwh-migration-client
@@ -1,0 +1,3 @@
+#!/bin/sh -ex
+
+exec dwh-migration-client "$@"
--- a/bin/dwh-migration-client
+++ b/bin/dwh-migration-client
@@ -1,3 +1,0 @@
-#!/bin/sh -ex
-
-exec dwh-migration-client "$@"
--- a/bin/dwh-migration-dumper
+++ b/bin/dwh-migration-dumper
@@ -1,12 +1,31 @@
-#!/bin/sh -ex
+#!/bin/bash
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 
-BASE_DIR=$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" 2>&1 > /dev/null && cd .. && pwd)
-. "${BASE_DIR}/dumper/app/src/main/sh/java_versioning.sh"
-BASE_BIN_DIR="${BASE_DIR}/dumper/app/build/install/app/bin"
-BIN=$(is_java_greater_than_8 && echo "${BASE_BIN_DIR}/dwh-migration-dumper" || echo "${BASE_BIN_DIR}/dwh-migration-dumper-java8")
+set -e
 
-if [ ! -x "$BIN" ] ; then
-  (cd "${BASE_DIR}" && ./gradlew --parallel :dumper:app:installDist)
+BASE_DIR="$(cd -- "$(dirname "${BASH_SOURCE[0]:-$0}")" 2>&1 > /dev/null && pwd -P)"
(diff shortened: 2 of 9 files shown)


cyulysses-corp (approved):

misolt (approved):

cyulysses-corp (commented):
Nice idea!

kpiotrowski (approved):

dumper/app/src/main/sh/java_versioning.sh
```diff
@@ -1,16 +1,28 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 
 function is_java_greater_than_8() {
-  java_path=$(which java)
-  java_version_output=$($java_path -version 2>&1)
+  java_version_output="$(java -version 2>&1)"
```
shevek-google, line 17: Not all JDKs support java -version. Some of them only support javac -version. I _think_ that's a non-issue post-8, but it's worth a check, particularly on the IBM JDK we ran into the other day (J9) as we clearly have customers running some really odd Java versions.

shevek-google (commented):
This is always fiddly, and one of the things I love most about the CreateStartScripts task is that it's Not Our Problem(tm). I wonder if we might be better allowing the VM to re-execute _itself_ with additional arguments? A JVM is more easily able to introspect itself than we can introspect it from shell. WDYT?

shevek-google:
Is the second (36) answer on this post interesting: https://stackoverflow.com/questions/4159802/how-can-i-restart-a-java-application ? Or is it worth looking at https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/lang/Module.html#addOpens(java.lang.String,java.lang.Module) since we run without a SecurityManager ?

shevek-google:
Can we do this with a module declaration, and (possibly) a multi-version JAR? https://www.baeldung.com/java-illegal-reflective-access

dumper/app/src/main/sh/java_versioning.sh
dawidxc: After merging this PR, this script (i.e. `java_versioning.sh`) will no longer be used by the dumper, only by GCE launcher. I merely added a license header and applied some small fixes, but this doesn't change the script behavior.

I would suggest to apply any changes to this script in a separate PR that would focus on GCE launcher. One of the options is that we can remove this script and use the same trick from this PR (i.e. the `JDK_JAVA_OPTIONS` variable) in the GCE launcher start scripts.

dawidxc:
@shevek-google 
> I wonder if we might be better allowing the VM to re-execute itself with additional arguments? A JVM is more easily able to introspect itself than we can introspect it from shell. WDYT?

We can prepare a better solution in a separate PR (for example, as you suggested - the VM re-executing itself or module declaration or something else). This PR is focused on the simplest way to fix the Java 8 compatibility issue, but there might be more elegant solutions that we could consider later.

shevek-google:
> @shevek-google
> 
> > I wonder if we might be better allowing the VM to re-execute itself with additional arguments? A JVM is more easily able to introspect itself than we can introspect it from shell. WDYT?
> 
> We can prepare a better solution in a separate PR (for example, as you suggested - the VM re-executing itself or module declaration or something else). This PR is focused on the simplest way to fix the Java 8 compatibility issue, but there might be more elegant solutions that we could consider later.

Which Java versions and JVMs have we tested this approach against?

dawidxc:
> > @shevek-google
> > > I wonder if we might be better allowing the VM to re-execute itself with additional arguments? A JVM is more easily able to introspect itself than we can introspect it from shell. WDYT?
> > 
> > 
> > We can prepare a better solution in a separate PR (for example, as you suggested - the VM re-executing itself or module declaration or something else). This PR is focused on the simplest way to fix the Java 8 compatibility issue, but there might be more elegant solutions that we could consider later.
> 
> Which Java versions and JVMs have we tested this approach against?

I've tested this against Java 8 and 22 in Linux and in Windows.

PR #439, opened 2024-06-19
Add tests for AbstractOracleConnector

misolt:
Add several tests for the `buildUrl` method.


dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
```diff
@@ -44,4 +50,52 @@ public void buildProperties_success() {
...
     assertEquals(EXAMPLE_USER, actual.getProperty("user"));
   }
+
+  @Test
+  public void buildUrl_providedBoth_throwsException() {
+    when(arguments.getOracleServicename()).thenReturn("ORCLPDB");
+    when(arguments.getOracleSID()).thenReturn("ORCLPDB1");
+    // Act
+    ThrowingRunnable runnable = () -> AbstractOracleConnector.buildUrl(arguments);
+    // Assert
```
dawidxc, line 60: I would suggest adding a single blank line before the `// Act` and `// Assert` comments.

Alternatively, since this test method is very short, structure the test into 3 sections separated by single blank lines and remove the comments.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
```diff
@@ -44,4 +50,52 @@ public void buildProperties_success() {
...
+    assertThrows(MetadataDumperUsageException.class, runnable);
```
dawidxc, line 61: I would recommend keeping the `assertThrows` call in the Act section and inlining the runnable.

See example usage:
https://github.com/google/dwh-migration-tools/blob/531ca76b2b1cb2cb383fa4b5923a7b44e1eb34db/dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/QueryLogTableNamesResolverTest.java#L142-L144

Similarly for other tests, where `assertThrows` is used.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
```diff
@@ -44,4 +50,52 @@ public void buildProperties_success() {
...
+  }
+
+  @Test
+  public void buildUrl_providedNone_throwsException() {
+    ThrowingRunnable runnable = () -> AbstractOracleConnector.buildUrl(arguments);
+    assertThrows(MetadataDumperUsageException.class, runnable);
```
dawidxc, line 67: I would suggest asserting the error message of the exception:
1. to make sure that the exception happened for the expected reason,
2. to check that the correct message is displayed to the user in this case.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
```diff
@@ -44,4 +50,52 @@ public void buildProperties_success() {
...
+    when(arguments.getOracleServicename()).thenReturn("ORCLPDB");
+    when(arguments.getOracleSID()).thenReturn("ORCLPDB1");
+    // Act
+    ThrowingRunnable runnable = () -> AbstractOracleConnector.buildUrl(arguments);
+    // Assert
+    assertThrows(MetadataDumperUsageException.class, runnable);
+  }
+
+  @Test
+  public void buildUrl_providedNone_throwsException() {
```
dawidxc, line 65: Optional: I would recommend rewording `None` to `NoServiceNameNorSid` to make the test more self-contained. Otherwise the test is difficult to reason about without reading other test cases.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
```diff
@@ -44,4 +50,52 @@ public void buildProperties_success() {
     assertEquals(EXAMPLE_PASSWORD, actual.getProperty("password"));
     assertEquals(EXAMPLE_USER, actual.getProperty("user"));
   }
+
+  @Test
+  public void buildUrl_providedBoth_throwsException() {
```
dawidxc, line 55: I would recommend rewording `Both` to `BothServiceNameAndSid`.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
```diff
@@ -44,4 +50,52 @@ public void buildProperties_success() {
...
+    // Act
+    String url = AbstractOracleConnector.buildUrl(arguments);
+    // Assert
+    assertEquals("jdbc:oracle:thin:@//localhost:1521/ORCLPDB", url);
+  }
+
+  @Test
+  public void buildUrl_providedSid_success() {
+    when(arguments.getOracleSID()).thenReturn("ORCLPDB1");
+    when(arguments.getHost()).thenReturn("localhost");
```
dawidxc, line 84: I would suggest using a bit more unusual test data in at least one test (e.g. `sample-host-123`) to check that the `localhost` is passed from the arguments when constructing the URL and that it is not hardcoded. Similarly for port.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
```diff
@@ -44,4 +50,52 @@ public void buildProperties_success() {
...
+    when(arguments.getPort(anyInt())).thenReturn(1521);
+    // Act
+    String url = AbstractOracleConnector.buildUrl(arguments);
+    // Assert
+    assertEquals("jdbc:oracle:thin:@localhost:1521:ORCLPDB1", url);
+  }
+
+  @Test
+  public void buildUrl_providedUrl_success() {
```
dawidxc, line 93: I would suggest adding also a test case, when both URL and host command-line flags are provided.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnector.java
```diff
@@ -184,7 +184,7 @@ Properties buildProperties(@Nonnull ConnectorArguments arguments) {
   // dbc:oracle:thin://<host>:<port>/<service>
   // jdbc:oracle:thin:<host>:<port>:<SID>
   // jdbc:oracle:thin:<TNSName> (from 10.2.0.1.0)
-  private String buildUrl(ConnectorArguments arguments) {
+  static String buildUrl(ConnectorArguments arguments) {
```
dawidxc, line 187: Optional: If it makes sense to change the method to `static`, then I would recommend extracting the logic to a separate class (e.g. `OracleUrlBuilder`).

misolt: I agree with this idea, in general I think that AbstractOracleConnector could be slimmed down or even removed eventually. However I'd prefer to do this later and limit this PR to what's required for testability.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
misolt: Done, renamed to `buildUrl_serviceNameAndSidBothProvided_throwsException`

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
misolt: Done, renamed to `buildUrl_serviceNameAndSidBothProvided_throwsException`

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
misolt: Thanks for bringing this up, it seems like an important corner case. I'm not sure what's the correct behavior here.

The current approach is to silently ignore the host, which I don't like, because a) it's bad to silently ignore a flag, b) the reason why we choose the url and not the host is unclear to the user and c) we should try to limit the number of possible flag combinations as much as possible to reduce complexity.

Ideally, I'd like to throw an usage exception, but in any case I'd discuss this further before including it in tests.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
misolt: IMO this is a slightly different piece of logic. The reason why I wanted to cover this is to verify that the URLs are built correctly for valid input data, without exceptions being thrown or malformed results, like missing a slash or colon, or a character added where it doesn't belong.

If we want to check that data is passed correctly from the cli args class to the connector classes, I don't mind adding it, but I'd write a separate test.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
misolt: Done, I added a method for creating this exception and verification that they match by the message.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
misolt: Done, I moved `assertThrows` inside `// Act` to match convention

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
misolt: Done, I've added a new test that provides uncommon host, port and sid values, and checks if they are get added to the built url.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
dawidxc: I would still slightly suggest to keep the Act step with a single assignment only as is already done in other tests, following the pattern:
```
E e = assertThrows(E.class, () -> ...);
```

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
```diff
@@ -44,4 +48,68 @@ public void buildProperties_success() {
...
+  @Test
+  public void buildUrl_serviceNameAndSidBothNull_throwsException() {
+    Exception serviceOrSid = OracleConnectorExceptions.mustProvideServiceOrSid();
+
+    // Act
+    ThrowingRunnable runnable = () -> AbstractOracleConnector.buildUrl(arguments);
+    Exception exception = assertThrows(MetadataDumperUsageException.class, runnable);
+
+    // Assert
+    assertEquals(serviceOrSid.getMessage(), exception.getMessage());
```
dawidxc, line 77: I would suggest writing the actual message that you want to assert against, instead of using the result of the call to the method in `OracleConnectorExceptions`.

See examples in other test classes, e.g. https://github.com/google/dwh-migration-tools/blob/7399219ad87ccfd15bf16e2eae702e5ba00c4cff/dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/teradata/SplitTextColumnQueryGeneratorTest.java#L246-L249

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleConnectorExceptions.java
```diff
@@ -16,12 +16,19 @@
...
+import com.google.edwmigration.dumper.application.dumper.MetadataDumperUsageException;
+
 class OracleConnectorExceptions {
 
   static IllegalArgumentException invalidSqlSourcePath(String path) {
     String message = String.format("Invalid path for SQL source file: %s", path);
     return new IllegalArgumentException(message);
   }
 
+  static MetadataDumperUsageException mustProvideServiceOrSid() {
```
dawidxc, line 28: Optional: I think the `OracleConnectorExceptions` class introduces unnecessary abstraction that causes too much indirection. Unless there are some other requirements that I'm not aware of, by default I would create the exception where it is thrown, so that no other class needs to be consulted when reading the code.

My main argument is simplicity - not to overcomplicate things too much.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
misolt: Done

dawidxc (approved):

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnectorTest.java
misolt: Done

PR #447, opened 2024-06-24
[b/341297017] Integrate sql-stats (AWR) from DMA

misolt:
part of [b/341297017]


dumper/app/src/main/resources/oracle-stats/awr/sql-stats-awr.sql
```diff
@@ -0,0 +1,54 @@
...
+    sum(A.javexec_time_delta) "SumJavexecTimeDelta"
+FROM cdb_hist_sqlstat A
+JOIN (
+    SELECT B.snap_id, B.instance_number, B.dbid
+    FROM cdb_hist_snapshot B
+) B
+ON A.dbid = B.dbid
+    AND A.instance_number = B.instance_number
+    AND A.snap_id = B.snap_id
+GROUP BY
```
kpiotrowski, line 51: This is missing a grouping by a query. In DMA script there is group by `force_matching_signature` column.

In Oracle, force_matching_signature is a value associated with each SQL statement in the database. It is a hash value that represents a normalized version of the SQL text, where literals (constants) are replaced with placeholders. This allows Oracle to group together SQL statements that are essentially the same, differing only in the literal values used.

misolt: Done, I added the missing column and grouping. Thanks for catching this

kpiotrowski (approved):
LGTM 👍  Is there a way to return the timerange for sql statistics? e.g. min and max snapshot time? There are columns like `BEGIN_INTERVAL_TIME` & `END_INTERVAL_TIME` in `DBA_HIST_SNAPSHOT` table.

PR #448, opened 2024-06-25
Fix: Parameter names don't match their values

misolt:
In the method `buildSelectStarTask`, the parameters and variables prefixed with `dba_` actually refer to tables prefixed with `all_` and vice versa.

Fix this by renaming the parameters to match their actual values.


paolomorandini (approved):
Thank you.

PR #451, opened 2024-06-27
Fixes for hist-cmd-types (STATSPACK) SQL

misolt:
- remove the 'delta' columns
- make column names consistent with the AWR SQL
- make the set of columns more similar to the AWR file


kpiotrowski (approved):
This SQL looks much more simple now, great work 👍

PR #469, opened 2024-07-16
[b/352469141] Make oracle-stats duration customizable

misolt:
The user can optionally provide the number of days for query logs, similarly to existing logs-type connectors.
The end of the timeframe will be the present time, the start will be the present time minus the provided number of days.

The number of days can be a value from the inclusive range [1, 10000]

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsConnector.java
@@ -18,6 +18,7 @@ package com.google.edwmigration.dumper.application.dumper.connector.oracle;
 
 import com.google.auto.service.AutoService;
 import com.google.edwmigration.dumper.application.dumper.ConnectorArguments;
+import com.google.edwmigration.dumper.application.dumper.MetadataDumperUsageException;
 import com.google.edwmigration.dumper.application.dumper.connector.Connector;
 import com.google.edwmigration.dumper.application.dumper.task.Task;
 import com.google.edwmigration.dumper.plugin.ext.jdk.annotation.Description;
@@ -37,12 +38,31 @@ public class OracleStatsConnector extends AbstractOracleConnector {
   @Override
   public void addTasksTo(List<? super Task<?>> out, ConnectorArguments arguments) throws Exception {
     StatsTaskListGenerator taskListGenerator = new StatsTaskListGenerator();
-    out.addAll(taskListGenerator.createTasks(arguments));
+    int queryLogDays = getQueryLogDays(arguments);
+    out.addAll(taskListGenerator.createTasks(arguments, queryLogDays));
   }
 
   @Nonnull
   @Override
   public String summary(String fileName) {
     return String.format("Oracle statistics saved to %s", fileName);
+  }
+
+  static int getQueryLogDays(ConnectorArguments arguments) {
+    Integer asObject = arguments.getQueryLogDays();
+    int shortTime = 7;
+    int longTime = 30;
+    if (asObject == null || asObject == longTime) {
+      return longTime;
+    } else if (asObject == shortTime) {
+      return shortTime;
+    } else {
+      throw invalidDuration(asObject.intValue());
+    }
+  }
+
+  private static MetadataDumperUsageException invalidDuration(int days) {
+    String message = "The number of days must be either 7 or 30. Was: " + days;
+    return new MetadataDumperUsageException(message);
   }
 }
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsConnector.java
@@ -22,6 +22,7 @@ import com.google.edwmigration.dumper.application.dumper.MetadataDumperUsageExce
 import com.google.edwmigration.dumper.application.dumper.connector.Connector;
 import com.google.edwmigration.dumper.application.dumper.task.Task;
 import com.google.edwmigration.dumper.plugin.ext.jdk.annotation.Description;
+import java.time.Duration;
 import java.util.List;
 import javax.annotation.Nonnull;
 import javax.annotation.ParametersAreNonnullByDefault;
@@ -31,7 +32,8 @@ import javax.annotation.ParametersAreNonnullByDefault;
 @ParametersAreNonnullByDefault
 public class OracleStatsConnector extends AbstractOracleConnector {
 
-  static int MAX_DAYS = 10000;
+  static final int MAX_DAYS = 10000;
+  static final Duration maxDuration = Duration.ofDays(MAX_DAYS);
 
   public OracleStatsConnector() {
     super(OracleConnectorScope.STATS);
@@ -40,8 +42,8 @@ public class OracleStatsConnector extends AbstractOracleConnector {
   @Override
   public void addTasksTo(List<? super Task<?>> out, ConnectorArguments arguments) throws Exception {
     StatsTaskListGenerator taskListGenerator = new StatsTaskListGenerator();
-    int queriedDays = getQueryLogDays(arguments);
-    out.addAll(taskListGenerator.createTasks(arguments, queriedDays));
+    Duration queriedDuration = getQueriedDuration(arguments);
+    out.addAll(taskListGenerator.createTasks(arguments, queriedDuration));
   }
 
   @Nonnull
@@ -50,20 +52,26 @@ public class OracleStatsConnector extends AbstractOracleConnector {
     return String.format("Oracle statistics saved to %s", fileName);
   }
 
-  static int getQueryLogDays(ConnectorArguments arguments) {
+  static Duration getQueriedDuration(ConnectorArguments arguments) {
+    Duration queriedDuration = extractFromArgs(arguments);
+    if (queriedDuration.compareTo(Duration.ofDays(0)) == 1
+        && queriedDuration.compareTo(maxDuration) < 1) {
+      return queriedDuration;
+    } else {
+      throw new MetadataDumperUsageException(
+          String.format(
+              "The number of days must be positive and not greater than %s. Was: %s",
+              MAX_DAYS, queriedDuration.toDays()));
+    }
+  }
+
+  private static Duration extractFromArgs(ConnectorArguments arguments) {
     int queriedDays;
     if (arguments.getQueryLogDays() == null) {
       queriedDays = 30;
     } else {
       queriedDays = arguments.getQueryLogDays();
     }
-    if (queriedDays > 0 && queriedDays <= MAX_DAYS) {
-      return queriedDays;
-    } else {
-      throw new MetadataDumperUsageException(
-          String.format(
-              "The number of days must be positive and not greater than %s. Was: %s",
-              MAX_DAYS, queriedDays));
-    }
+    return Duration.ofDays(queriedDays);
   }
 }
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsConnector.java
@@ -49,15 +49,15 @@ public class OracleStatsConnector extends AbstractOracleConnector {
   }
 
   static int getQueryLogDays(ConnectorArguments arguments) {
-    Integer asObject = arguments.getQueryLogDays();
+    Integer queryLogDaysObject = arguments.getQueryLogDays();
     int shortTimeInDays = 7;
     int longTimeInDays = 30;
-    if (asObject == null || asObject == longTimeInDays) {
+    if (queryLogDaysObject == null || queryLogDaysObject == longTimeInDays) {
       return longTimeInDays;
-    } else if (asObject == shortTimeInDays) {
+    } else if (queryLogDaysObject == shortTimeInDays) {
       return shortTimeInDays;
     } else {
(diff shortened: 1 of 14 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsConnector.java
```diff
@@ -37,12 +38,31 @@ public OracleStatsConnector() {
...
   }
 
   @Nonnull
   @Override
   public String summary(String fileName) {
     return String.format("Oracle statistics saved to %s", fileName);
   }
+
+  static int getQueryLogDays(ConnectorArguments arguments) {
+    Integer asObject = arguments.getQueryLogDays();
```
dawidxc, line 52: I would suggest renaming `asObject` to `queryLogDays` to increase readability.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsConnector.java
```diff
@@ -37,12 +38,31 @@ public OracleStatsConnector() {
...
+    int shortTimeInDays = 7;
+    int longTimeInDays = 30;
+    if (asObject == null || asObject == longTimeInDays) {
+      return longTimeInDays;
+    } else if (asObject == shortTimeInDays) {
+      return shortTimeInDays;
+    } else {
+      throw invalidDuration(asObject.intValue());
```
dawidxc, line 60: I would recommend directly throwing the exception instead of delegating it to a method. The additional indirection seems unnecessary.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsQuery.java
```diff
@@ -39,9 +41,10 @@ public abstract class OracleStatsQuery {
   @Nonnull
   abstract StatsSource statsSource();
 
-  static OracleStatsQuery create(String name, StatsSource statsSource) throws IOException {
+  static OracleStatsQuery create(String name, StatsSource statsSource, int days)
```
dawidxc, line 44: I would suggest keeping the name of the accessor and the parameter the same, e.g. `queriedDays`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/StatsTaskListGenerator.java
```diff
@@ -61,20 +61,20 @@ class StatsTaskListGenerator {
       ImmutableList.of("hist-cmd-types-statspack", "sql-stats-statspack");
 
   @Nonnull
-  ImmutableList<Task<?>> createTasks(ConnectorArguments arguments) throws IOException {
+  ImmutableList<Task<?>> createTasks(ConnectorArguments arguments, int logDays) throws IOException {
```
dawidxc, line 64: How does `logDays` differ from `queriedDays` and from `queryLogDays`?

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsConnectorTest.java
```diff
@@ -17,17 +17,59 @@
...
+  public void getQueryLogDays_wrongNumberOfDays_throwsException() throws IOException {
+    ConnectorArguments arguments =
+        new ConnectorArguments("--connector", "oracle-stats", "--query-log-days", "999");
+
+    MetadataDumperUsageException exception =
+        assertThrows(
+            MetadataDumperUsageException.class,
+            () -> OracleStatsConnector.getQueryLogDays(arguments));
+
+    assertTrue(exception.getMessage().contains("999"));
```
dawidxc, line 72: What if the message contains `99999`?

In general, I would suggest asserting the whole exception message using `equals` and use other means like `contains` or `startsWith` for complex messages that are impractical to assert as a whole.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/StatsTaskListGenerator.java
misolt: Functionally there's no difference, they are different names for the same concept.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsQuery.java
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/StatsTaskListGenerator.java
dawidxc: I would suggest keeping the same name across the codebase to make it easier to read the code without remembering renames when moving across different classes/methods. The rename makes sense, if the meaning of the value changes (e.g. some method is more general). If the meaning doesn't change, keeping the name seems more straight-forward.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsConnector.java
misolt: Resolving as outdated - the set of correct values was extended and after the rewrite this variable is gone.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/StatsTaskListGenerator.java
misolt: Done, I replaced `logDays` and `queryLogDays` with `queriedDays`.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsConnectorTest.java
misolt: With more cases of invalid values added to the test, it's no longer practical to check the message with equals. I'd consider this outdated, but feel free to reopen if needed.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsConnector.java
```diff
@@ -30,19 +31,39 @@
 @ParametersAreNonnullByDefault
 public class OracleStatsConnector extends AbstractOracleConnector {
 
+  static int MAX_DAYS = 10000;
```
dawidxc, line 34: I would suggest expressing the number of max days as `Duration`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsConnector.java
```diff
@@ -30,19 +31,39 @@
...
+    out.addAll(taskListGenerator.createTasks(arguments, queriedDays));
   }
 
   @Nonnull
   @Override
   public String summary(String fileName) {
     return String.format("Oracle statistics saved to %s", fileName);
   }
+
+  static int getQueryLogDays(ConnectorArguments arguments) {
```
dawidxc, line 53: I would suggest using `Duration` as the return type.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsQuery.java
```diff
@@ -30,6 +30,8 @@
 @ParametersAreNonnullByDefault
 public abstract class OracleStatsQuery {
 
+  abstract int queriedDays();
```
dawidxc, line 33: I would suggest to keep the time duration in the `Duration` type and convert it to days only at the latest possible moment.

misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsConnector.java
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsConnector.java
misolt: Done, I added the constant `maxDays`.

dawidxc: Constants in java follow the UPPER_SNAKE_CASE convention. I would suggest renaming to `MAX_DURATION`. Also this constant can be used directly in the error message. No need to keep the `MAX_DAYS` one.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsConnector.java
```diff
@@ -30,19 +32,46 @@
...
+          String.format(
+              "The number of days must be positive and not greater than %s. Was: %s",
+              MAX_DAYS, queriedDuration.toDays()));
+    }
+  }
+
+  private static Duration extractFromArgs(ConnectorArguments arguments) {
+    int queriedDays;
+    if (arguments.getQueryLogDays() == null) {
+      queriedDays = 30;
```
kpiotrowski, line 71: I suggest to define constant variable with default Duration. You can then simplify this code:

static final Duration defaultQueriedDaysDuration = Duration.ofDays(30);

...

if (arguments.getQueryLogDays() == null) {
    return defaultQueriedDaysDuration;
}
return return Duration.ofDays(arguments.getQueryLogDays())

kpiotrowski (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsConnector.java
```diff
@@ -30,19 +32,46 @@
...
 
   @Nonnull
   @Override
   public String summary(String fileName) {
     return String.format("Oracle statistics saved to %s", fileName);
   }
+
+  static Duration getQueriedDuration(ConnectorArguments arguments) {
+    Duration queriedDuration = extractFromArgs(arguments);
+    if (queriedDuration.compareTo(Duration.ofDays(0)) == 1
```
dawidxc, line 57: The expression `queriedDuration.compareTo(Duration.ofDays(0)) == 1` can be simplified to `DurationUtils.isPositive(queriedDuration)`.

Alternatively, both comparisons could be rewritten as:
```
Range.openClosed(ZERO, MAX_DURATION).contains(queriedDuration)
```

misolt: The first suggestion has a different meaning. I'd like it to be "at least one day", rather than "positive duration".
In this case the difference doesn't change the behavior, because we operate on an integer number of days. However, the fact that these durations are always divisible by 1 day isn't clear to the reader and I don't see a good way to make it so.

The reason why the "at least one day" part is significant is that the only purpose of these Duration's is extracting the number of days. This means that if some future change led to a query with the duration being e.g. 6 hours, the actual query will use an empty time range and get no data.

Similarly, the `Range` can be used with `closed` instead of `openClosed` if we use a constant representing a 1-day duration. To me, neither option seems significantly better, please let me know if you have a preference.

misolt: Correction: the original code with `queriedDuration.compareTo(Duration.ofDays(0)) == 1` is also wrong, it should be  `queriedDuration.compareTo(Duration.ofDays(1)) >= 0` and this discussion made me realize it. I'll update the code in a moment.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsConnector.java
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsConnector.java
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsConnector.java
dawidxc: > I'd like it to be "at least one day"
In this case, you could use `Range.closed(ofDays(1), MAX_DURATION)` as you already suggested in your comment.

> the fact that these durations are always divisible by 1 day isn't clear to the reader
I wouldn't worry about this aspect. The point is that the duration is converted to days, when being passed from Java to SQL query. The precision may be lost there, but I don't think that this would be surprising, if the original duration was not divisible by 1 day.

misolt: Done

PR #477, opened 2024-07-19
Wrap IOException for oracle-stats SQL resources

misolt:
[oracle-stats connector]

The `IOException` thrown by a method in `common.io.Resources` is due to a general-purpose `read` method used by the implementation.

As the [documentation](https://docs.oracle.com/javase/8/docs/api/java/io/IOException.html) shows, the exception is very broad, covering various issues related to permissions, hosts, sockets, interrupts and other topics.

Oracle-stats uses `Resources` only to load SQLs included in the JAR. Almost all subclasses of IOException are inapplicable to this use case. The only issues with these files that we see somewhat frequently are:
- file is missing
- provided path to the file is incorrect

Neither  of these results in an IOException, because `Resources` will throw `IllegalArgumentException` before getting to the stage where an `IOException` could be thrown. An `IOException` being thrown when reading SQL for oracle-stats should be treated as programmer error, so it can use an unchecked exception class.

For the relevant cases, wrap `IOException` in an `IllegalArgumentException` that will inform about an invalid file to match the behavior of `io.Resources`

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsQuery.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsQuery.java
@@ -43,6 +43,9 @@ public abstract class OracleStatsQuery {
   abstract String name();
 
   @Nonnull
+  abstract String queryText();
+
+  @Nonnull
   abstract StatsSource statsSource();
 
   @Nonnull
@@ -62,15 +65,10 @@ public abstract class OracleStatsQuery {
 
   private static OracleStatsQuery create(
       boolean isRequired, Duration queriedDuration, String name, StatsSource statsSource) {
-    return new AutoValue_OracleStatsQuery(isRequired, queriedDuration, name, statsSource);
-  }
-
-  @Nonnull
-  String queryText() {
-    String source = statsSource().value;
-    String tenantSetup = "cdb";
-    String path = String.format("oracle-stats/%s/%s/%s.sql", tenantSetup, source, name());
-    return loadFile(path);
+        String source = statsSource.value;
+        String tenantSetup = "cdb";
+        String path = String.format("oracle-stats/%s/%s/%s.sql", tenantSetup, source, name);
+    return new AutoValue_OracleStatsQuery(isRequired, queriedDuration, name, loadFile(path), statsSource);
   }
 
   @Nonnull
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsQuery.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsQuery.java
@@ -83,8 +83,8 @@ public abstract class OracleStatsQuery {
       URL queryUrl = Resources.getResource(path);
       return Resources.toString(queryUrl, UTF_8);
     } catch (IOException e) {
-      String message = String.format("An invalid file was provided: %s.", path);
-      throw new IllegalArgumentException(message, e);
+      throw new IllegalArgumentException(
+          String.format("An invalid file was provided: %s.", path), e);
     }
   }
 }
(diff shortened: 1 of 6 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsQuery.java
```diff
@@ -42,44 +42,49 @@ public abstract class OracleStatsQuery {
...
 
-  private static String loadFile(String path) throws IOException {
-    URL queryUrl = Resources.getResource(path);
-    return Resources.toString(queryUrl, UTF_8);
+  private static String loadFile(String path) {
+    try {
+      URL queryUrl = Resources.getResource(path);
+      return Resources.toString(queryUrl, UTF_8);
+    } catch (IOException e) {
+      String message = String.format("IOException caused by invalid file: %s.", path);
```
dawidxc, line 86: Optional: I would suggest against using `IOException` in the error message, because it is duplicated information. It is already stated in the cause of IAE. The error message could simply be: "Error reading resource: '%s'.".

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsQuery.java
```diff
@@ -42,44 +42,49 @@ public abstract class OracleStatsQuery {
...
+      throw new IllegalArgumentException(message, e);
```
dawidxc, line 87: Optional: I would very slightly suggest inlining the `message` variable as using `String.format` is a common pattern when throwing an exception that is used across the dumper codebase.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsQuery.java
```diff
@@ -42,44 +42,49 @@ public abstract class OracleStatsQuery {
...
+      boolean isRequired, Duration queriedDuration, String name, StatsSource statsSource) {
+    return new AutoValue_OracleStatsQuery(isRequired, queriedDuration, name, statsSource);
+  }
+
+  @Nonnull
+  String queryText() {
+    String source = statsSource().value;
+    String tenantSetup = "cdb";
+    String path = String.format("oracle-stats/%s/%s/%s.sql", tenantSetup, source, name());
+    return loadFile(path);
```
dawidxc, line 73: I understand the change with catching `IOException` inside the `loadFile`, but it's unclear to me, why you are moving the loading from `create` to the `queryText` method. I think the previous implementation was better in this aspect. I might be missing something here. I would recommend keeping the getter simple and restructuring the code in such a way that the SQL is read just once, no matter how many times the getter is invoked. (even if in practice it is currently invoked just once)

misolt: Done, I restored the property. The `create` method throwing an exception was inconvenient to use, partly because the exception was checked and partly because it needed to reference a valid resource file whenever it's called. The first point is gone, if the second ever starts causing problems I'll take a second look at this. In that case, maybe Memoized can be used to avoid re-reading the SQL file.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsQuery.java
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleStatsQuery.java
misolt: Done, the message now says `"An invalid file was provided: [providedFileName]."`

dawidxc (approved):

PR #485, opened 2024-07-24
[b/352468287] Add non-CDB versions of db-info and data-types

misolt:
- add DBA versions of data-types and db-info
- fill in the app-schemas-pdbs implementation

### DIFF
--- a/dumper/app/src/main/resources/oracle-stats/dba/native/app-schemas-pdbs.sql
+++ b/dumper/app/src/main/resources/oracle-stats/dba/native/app-schemas-pdbs.sql
@@ -21,9 +21,9 @@ SELECT
   D.logging "Logging",
   D.con_id "ConId",
   D.con_uid "ConUid",
-  NULL "EbsOwner",
-  NULL "SiebelOwner",
-  NULL "PsftOwner",
+  F.min_owner "EbsOwner",
+  H.min_owner "SiebelOwner",
+  J.min_owner "PsftOwner",
   NULL "RdaMatchCount",
   NULL "OciAutoViewMatches",
   NULL "DbmsCloudMatches",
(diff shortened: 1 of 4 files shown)


dumper/app/src/main/resources/oracle-stats/dba/native/db-info.sql
```diff
@@ -0,0 +1,23 @@
...
+-- limitations under the License.
+SELECT
+  A.dbid "DbId",
+  A.name "Name",
+  A.db_unique_name "DbUniqueName",
+  NULL "ConId",
+  NULL "ConDbId",
+  NULL "Cdb",
+  A.database_role "DatabaseRole"
+FROM v$database A
```
kpiotrowski, line 23: v$database table is the same for CDB and non-CDB so we could keep the single SQL definition. Do you expect any issues with extracting this data from non-CDB instance or Oracle 11 (version before multi tenant was introduced).

kpiotrowski: https://docs.oracle.com/cd/E11882_01/server.112/e40402/dynviews_1097.htm#REFRN30047 looks like in oracle 11 there is no `ConId`, `ConDbId`, `Cdb` columns so we need to keep this version to support older oracle version.

misolt: True, I was using the same source

dumper/app/src/main/resources/oracle-stats/dba/native/app-schemas-pdbs.sql
```diff
@@ -13,20 +13,44 @@
...
 SELECT
-  NULL "Source",
-  NULL "DbId",
-  NULL "PdbId",
-  NULL "PdbName",
-  NULL "Status",
-  NULL "Logging",
-  NULL "ConId",
-  NULL "ConUid",
+  D.source "Source",
```
kpiotrowski, line 16: I would recommend to skip sqls related to `pdbs` for single tenant instance. This query should work for the single tenant instance with version 12c+ but it will fail on Oracle 11c due to missing dba_pdbs table.
This can addressed later. For now the highest priority is to support Oracle 12c+. We can work on version 11 later.

kpiotrowski (approved):

dumper/app/src/main/resources/oracle-stats/dba/native/app-schemas-pdbs.sql
misolt: I agree, this file was just a random example so that https://github.com/google/dwh-migration-tools/pull/482 would have something to run. As discussed, the current priority is to first move the required, native sqls and the next group will be the sqls using AWR.

PR #501, opened 2024-08-05
Add a summary success message for each stats SQL group

misolt:
The task of determining if a given SQL group was successful is complicated from the user's perspective.
For example, if they want to determine whether CDBs AWR data will be available in the archive, they need to:
- go through the list of all task statuses
- find the ones which refer to CDBs and have AWR as their statsSource
- ensure that the status is either SUCCEEDED or SKIPPED for each task

Use existing Dumper mechanisms to simplify this process. Provide a summary for each group so that after successful extraction of data from an entire SQL group, an informational message will be printed, e.g:
```
* Task SUCCEEDED (OPTIONAL) Load AWR data (cdb version).
```

This is similar to what's done in OracleMetadataConnector, with two differences: the message is also shown on success and its location is in the task list instead of in the logs.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/task/VoidTask.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/task/VoidTask.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2022-2024 Google LLC
+ * Copyright 2013-2021 CompilerWorks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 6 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/task/VoidTask.java
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+package com.google.edwmigration.dumper.application.dumper.connector.oracle.task;
+
+import com.google.edwmigration.dumper.application.dumper.task.Task;
+import com.google.edwmigration.dumper.application.dumper.task.TaskRunContext;
+import javax.annotation.ParametersAreNonnullByDefault;
+
+@ParametersAreNonnullByDefault
+abstract class VoidTask implements Task<Void> {
```
dawidxc, line 24: Optional: It's not very important, but I would slightly suggest renaming the class to `NoResultTask`, because it seems to me that the point of introducing this class is that you are trying to avoid `Void` (even though it's a generic of the implemented interface), so I would emphasize in the name that the reason is to have "no result".

dawidxc (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/task/VoidTask.java
misolt: Done
Avoiding the generic with `Void` is the intention, because `Task<Void>` is a very common type of usage. In this case having the generic doesn't really benefit us, but it adds a bit more complexity to the API.

PR #510, opened 2024-08-07
Run STATSPACK only if both versions of AWR have failed

misolt:
Currently, Dumper attempts to extract both AWR and STATSPACK data regardless of the success of the other. The same goes for CDB and non-CDB versions.

Change the generated Tasks to implement a different strategy:
- first, attempt AWR tasks with CDB support
- if failed, try to extract AWR data, but in the non-CDB version
- finally, if both earlier steps failed, try to extract STATSPACK data (for which the presence of a multi-tenant setup doesn't make a difference)

Notes:
This PR creates the class `OracleSqlListing`, which is used for AWR and STATSPACK, but unused for NATIVE SQLs. The intention is to use this class for all SQLs, including NATIVE. This gap will be addressed in a later change.

Another gap is renaming the AWR and STATSPACK files to match the other group (dropping the suffixes). I'm also planning this for a later PR.

The implementation uses message-tasks introduced earlier, together with existing Dumper architecture, so that we can have groups of Tasks that depend on other groups having been executed (or skipped). The plan looks like this:
1. AWR CDB tasks are always attempted
2. If all AWR CDB tasks succeded, an AWR CDB summary task (S1) is run
3. If the summary was skipped, AWR DBA tasks are attempted
4. If the AWR DBA tasks succeded, a second summary task (S2) is executed
5. Now we need to check separately for S1 and S2 - there are three options: only S1 was skipped, only S2 was skipped or both were skipped. If both S1 and S2 were skipped, then AWR must have failed in both versions. This is the only case where STATSPACK is run, otherwise it will be skipped.


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/task/ResultMessageTask.java
```diff
@@ -49,6 +50,11 @@ public static NoResultTask create(QueryGroup group, List<StatsJdbcTask> tasks) {
     return new ResultMessageTask(condition, group);
   }
 
+  public static NoResultTask createNoFilter(QueryGroup group, List<Task<?>> tasks) {
```
dawidxc, line 53: I try to understand the naming (in particular `NoFilter`) in the context of the `create` method above. It seems to me that the filtering (applied in the `create` method above) makes it a bit difficult to understand. I would suggest separating the filtering to another method and have just one `create` method.

Before:
```
ResultMessageTask.create(queryGroup, tasks);
```

After:
```
ResultMessageTask.create(queryGroup, queryGroup.filter(tasks))
```

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleSqlList.java
```diff
@@ -0,0 +1,88 @@
...
+import com.google.edwmigration.dumper.application.dumper.task.Task;
+import java.time.Duration;
+import java.util.List;
+import java.util.function.Function;
+import javax.annotation.ParametersAreNonnullByDefault;
+
+/** Represents a list of oracle-stats SQL files that share a common QueryGroup. */
+@AutoValue
+@ParametersAreNonnullByDefault
+abstract class OracleSqlList {
```
dawidxc, line 36: Optional: I would suggest avoiding suffixing the class `List`, if it does not implement the `List` interface. Such a naming is a bit surprising.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleSqlList.java
```diff
@@ -0,0 +1,88 @@
...
+
+  static OracleSqlList awrDba() {
+    return AWR_DBA;
+  }
+
+  static OracleSqlList statspack() {
+    return STATSPACK_CDB;
+  }
+
+  private static OracleSqlList create(QueryGroup group, ImmutableList<String> names) {
```
dawidxc, line 66: Since this autovalue has a private `create` method, have you considered introducing an enum instead of autovalue?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleSqlList.java
```diff
@@ -0,0 +1,88 @@
...
+    return new AutoValue_OracleSqlList(names, group);
```
dawidxc, line 67: nit: Not very important, but I would suggest keeping the order of fields and `create` method parameters the same.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleSqlList.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  ImmutableList<Task<?>> toTasks(Duration queriedDuration) {
+    Function<String, Task<?>> mapper =
+        x -> {
```
dawidxc, line 72: I would suggest avoiding single-letter variables. What does `x` mean in this context?

Similarly for the method below.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/task/StatsJdbcTask.java
```diff
@@ -69,6 +70,17 @@ public StatsJdbcTask onlyIfFailed(Task<?> prerequisite) {
     return new StatsJdbcTask(query, failureCondition);
   }
 
+  @Nonnull
+  public static StatsJdbcTask onlyIfAllSkipped(
```
dawidxc, line 74: When I look at the implementation without looking into more details, I would say that the method should be called `onlyIfAllSucceeded`. Why is it called `onlyIfAllSkipped`? How does it work?

dawidxc (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/StatsTaskListGenerator.java
```diff
@@ -65,47 +65,40 @@ class StatsTaskListGenerator {
...
+      ImmutableList<Task<?>> messageTasksSoFar = copyOf(messageTasks);
+      ImmutableList<Task<?>> queryTasks =
+          sqls.toTasksIfAllSkipped(queriedDuration, messageTasksSoFar);
+      Task<?> success = ResultMessageTask.createNoFilter(sqls.group(), queryTasks);
+      builder.addAll(queryTasks);
+      messageTasks.add(success);
+    }
+
+    // Summary messages
+    builder.addAll(messageTasks);
```
dawidxc, line 87: I assume that the order of tasks in the builder is:
queryTasks1, queryTasks2, queryTasks3, messageTasks, nativeTasks

IIUC the execution of the queryTasks2 depends on the first messageTask. How can queryTasks2 check the condition of messageTask, if it is scheduled to be executed later than queryTasks2?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/task/ResultMessageTask.java
misolt: Done. I've taken another look at this and it turns out that `create` and `createNoFilter` do the same thing in two ways. The version with the filter looks hackier to me - it expects a combined list of all tasks, even though we know that the tasks it will extract must have been created separately to begin with.

I've removed the filter version. Since there's only one method, it can be just called `create`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleSqlList.java
misolt: Done, I changed the suffix from `List` to `Listing`

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleSqlList.java
misolt: I suspected that we'll need to create instances of this class dynamically in the future, because the value for `names` could be dynamic, e.g. due to some files being conditionally skipped. It's not very likely, but for now I'd stay with this approach, because it's more flexible than the enum.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleSqlList.java
misolt: Done. The `x` stands for the filename without the `.sql` extension. I've renamed it to `name`, matching the `names()` getter.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleSqlList.java
misolt: Done, I reordered them in `create`

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/task/StatsJdbcTask.java
```diff
@@ -69,6 +70,17 @@ public StatsJdbcTask onlyIfFailed(Task<?> prerequisite) {
...
+      OracleStatsQuery query, List<Task<?>> skippableTasks) {
+    ImmutableList<Condition> conditions =
+        skippableTasks.stream()
+            .map(item -> new StateCondition(item, TaskState.SUCCEEDED))
```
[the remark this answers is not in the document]
misolt: I added a condition factory method to replace this lambda with a reference. The name `task` is used in the factory method instead of `item`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/StatsTaskListGenerator.java
```diff
@@ -65,47 +65,41 @@ class StatsTaskListGenerator {
...
-    QueryGroup awrAndNotCdb = QueryGroup.create(/* required= */ false, AWR, SINGLE_TENANT);
-    builder.add(ResultMessageTask.create(awrAndCdb, jdbcTasks));
-    builder.add(ResultMessageTask.create(awrAndNotCdb, jdbcTasks));
+    ArrayList<Task<?>> messageTasks = new ArrayList<>();
+    ImmutableList<OracleSqlListing> listsInOrder =
+        ImmutableList.of(awrCdb(), awrDba(), statspack());
+
+    for (OracleSqlListing sqls : listsInOrder) {
+      ImmutableList<Task<?>> messageTasksSoFar = copyOf(messageTasks);
+      ImmutableList<Task<?>> queryTasks =
```
kpiotrowski, line 80: If I understand correctly, if only one AWR CDB task from the group fails, the ResultMessageTask will be skipped and all tasks (including succeeded) will be executed again with the DBA version. What will be the result in this case? We should not override successful results.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleSqlList.java
```diff
@@ -0,0 +1,88 @@
...
+          OracleStatsQuery query = OracleStatsQuery.create(x, group(), queriedDuration);
+          return StatsJdbcTask.fromQuery(query);
+        };
+    return names().stream().map(mapper).collect(toImmutableList());
+  }
+
+  ImmutableList<Task<?>> toTasksIfAllSkipped(
+      Duration queriedDuration, List<Task<?>> skippableTasks) {
+    Function<String, Task<?>> mapper =
```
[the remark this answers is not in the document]
misolt: Generics aren't an issue here, but it seems like any way to write this method will be slightly ugly. Inlining `mapper` adds 1-2 extra lines (depending on if the mapper declaration counts) and indents the lambda body pretty far:
``` java
  ImmutableList<Task<?>> toTasksIfAllSkipped(
      Duration queriedDuration, List<Task<?>> skippableTasks) {
    return names().stream()
        .map(
            name -> {
              OracleStatsQuery query = OracleStatsQuery.create(name, group(), queriedDuration);
              return StatsJdbcTask.onlyIfAllSkipped(query, skippableTasks);
            })
        .collect(toImmutableList());
  }
```
Which could look better with different whitespace, but spotless won't allow it.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/task/StatsJdbcTask.java
misolt: `SUCCEEDED` was incorrect here, it should have been `SKIPPED`. I've fixed the condition, thanks for catching this

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleSqlList.java
dawidxc: You can always change the code later. However, if you already predict this use-case now, then you could consider an autovalue and an enum that accepts this autovalue as a parameter. They could both share the same interface and the enum can delegate all the methods to the autovalue.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/StatsTaskListGenerator.java
misolt: True, this is a bug. I forgot about the need to manually order the tasks when I added line 88.

The correct approach is to never add from `messageTasks` to `builder`. Instead, the `success` task should be added while inside the loop.

misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleSqlList.java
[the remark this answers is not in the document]
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/StatsTaskListGenerator.java
misolt: This is true, Dumper assumes that any failure of the AWR CDB queries means that there is no multi-tenant setup - if any of the required native queries had failed, that would have been a different thing.
The assumption is due to Dumper's simplified architecture, where we don't specifically handle the errors, just try to obtain all information from the parts that do succeed.

In the case when:
- all required native queries succeeded
- there is CDB/PDBs support in the client DB
- some AWR CDB queries fail

It's unclear what Dumper should do. I'd assume that this is an invalid state and therefore we can't rely on the successful AWR CDBs results to be usable.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/OracleSqlList.java
misolt: If I understand correctly, this would still mean that there's an enum API. However, if the solution is supposed to work for dynamic values of the `names` list, then any enums would have to be hidden.

It could be done if the enums were encapsulated, but I'm not sure if this addresses the comments.

PR #513, opened 2024-08-08
[b/358310586] Fix: 'url' flag does not work with Oracle connectors

misolt:


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -157,11 +157,11 @@ public class ConnectorArguments extends DefaultArguments {
           .describedAs("com.company.Driver");
   private final OptionSpec<String> optionUri =
       parser
-          .accepts(OPT_URI, "JDBC driver URI (overrides host, port, etc if given)")
+          .accepts(OPT_URI, "JDBC driver URI")
```
dawidxc, line 160: What is the reason for this change? The option is used by various connectors. Do you want to make this change for all of them?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -568,14 +568,20 @@ public String getUri() {
...
   }
 
+  @Nonnull
+  public String getHostOrDefault() {
+    return getHost(OPT_HOST_DEFAULT);
+  }
+
   @CheckForNull
-  public String getHost() {
+  public String getHostOrNull() {
```
dawidxc, line 577: I would suggest keeping the `getHost` name. It is implied in the type that the return value might be `null`. It's fine to keep two methods: `getHostOrDefault` and `getHost`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -568,14 +568,20 @@ public String getUri() {
...
     return getOptions().valueOf(optionHost);
   }
 
   @Nonnull
   public String getHost(@Nonnull String defaultHost) {
-    return MoreObjects.firstNonNull(getHost(), defaultHost);
+    String nullableHost = getHostOrNull();
+    return nullableHost != null ? nullableHost : OPT_HOST_DEFAULT;
```
dawidxc, line 584: I would suggest keeping the previous implementation.

1. The `nullableHost` should be called `host`, because the nullability is implied in the type.
2. After changing to:

```
String host = getHost();
return host != null ? host : OPT_HOST_DEFAULT;
```

you can then inline it to: `return MoreObjects.firstNonNull(getHost(), OPT_HOST_DEFAULT);`

3. The `defaultHost` parameter is not used, therefore further simplification leads to the original code:

```
return MoreObjects.firstNonNull(getHost(), defaultHost);
```

dawidxc (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
misolt: The overriding behavior isn't changed in this PR, but the removed description was incorrect. In practice each connector is expected to implement its own handling of this collision.

In most cases host/port/sid, etc. is ignored as promised, but this can't be guaranteed by ConnectorArguments.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
misolt: Done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
misolt: Without context it's a better name, but it also could be misleading for someone used to the codebase, because the previous version of `getHost` did provide a default.

kpiotrowski: I would keep the original name mostly because the same convention is used for other flags and methods like `getPort`. Introducing `OrNull` suffix is yet another convention that duplicates information provided in `@CheckForNull`

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnector.java
```diff
@@ -200,7 +200,7 @@ static String buildUrl(ConnectorArguments arguments) {
   }
 
   private static void checkNonUriFlags(ConnectorArguments arguments) {
-    if (arguments.getHost() != null) {
+    if (arguments.getHostOrNull() != null) {
       throw extraFlagProvided(OPT_HOST);
```
kpiotrowski, line 204: IIRC there was no exception logged in the dumper output, or did I miss something?

kpiotrowski (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
dawidxc: Right, it can't be guaranteed by `ConnectorArguments` class. Rather the help string is just a description of the behavior of the app (even if internally it is implemented in many places). The overriding behavior could potentially be refactored to a common class, so that the rule would be implemented in just one place.

Was there a connector that broke this promise?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
dawidxc: > (...) it also could be misleading for someone used to the codebase (...)

I would avoid biasing ourselves too much with previous versions of the code and instead focus more on making the codebase consistent in terms of `OrDefault` and `OrNull` suffixes.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
misolt: In general I think it's better to break this promise - when there's ambiguity because the user passed a redundant flag, failing is better than arbitrarily choosing one option.

Changing all old connectors would be too risky, but Oracle connectors are meant to fail if there's ambiguous configuration - this includes things like providing service name together with sid or providing the full url together with the host name.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/oracle/AbstractOracleConnector.java
misolt: I did see the exception when I tried to reproduce the issue. Dumper's behavior can have some quirks depending on where it's run. Is it possible in your setup to verify whether the exception is thrown and not printed to the output or not thrown at all?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
misolt: Done, now there are 2 methods:
- `getHost`: does what old `getHost` was supposed to do
- `getHostOrDefault`: does what old `getHost` actually did

This means that most usages of `getHost` were replaced with `getHostOrDefault` to avoid breaking changes

PR #517, opened 2024-08-13
[b/359499526] Fix Ranger services date parsing.

aleofreddi:


misolt (approved):

dumper/app/src/test/resources/ranger/services.success.jsonl
```diff
@@ -1,2 +1,2 @@
...
+{"configs":{"fs.default.name":"hdfs://hadoop-m:8020","hadoop.security.authentication":"Simple","hadoop.security.authorization":"No","password":"*****","username":"admin"},"createdBy":"Admin","description":"Hadoop hdfs service","displayName":null,"guid":"4c2780ee-e244-4ac3-b21a-3660e6895980","id":1,"name":"hadoop-dataproc","policyUpdateTime":"2024-07-26T17:58:21Z","policyVersion":9,"tagService":null,"tagUpdateTime":"2024-07-26T17:58:22Z","tagVersion":1,"type":"hdfs","updatedBy":"Admin","version":1,"createDate":1716795282000,"enabled":true,"updateDate":1716795282000}
+{"configs":{"jdbc.driverClassName":"org.apache.hive.jdbc.HiveDriver","jdbc.url":"jdbc:mysql://localhost","password":"*****","username":"admin"},"createdBy":"Admin","description":"Hadoop HIVE service","displayName":null,"guid":"1fe46519-233a-4780-be20-a7eb180f1ca6","id":2,"name":"hive-dataproc","policyUpdateTime":"2024-07-26T17:58:21Z","policyVersion":6,"tagService":null,"tagUpdateTime":"2024-07-26T17:58:22Z","tagVersion":1,"type":"hive","updatedBy":"Admin","version":1,"createDate":1716795282000,"enabled":true,"updateDate":1716795282000}
```
dawidxc, line 2: I would suggest keeping the previous format for expressing the timestamps. Would it be possible to instead convert the seconds that Ranger API returns to Instant and keep the format unchanged?

dawidxc, line 2: Are `tagUpdateTime` and `policyUpdateTime` attributes that also contain timestamps unaffected?

shevek-google: I am in favour of supporting nanoseconds; I think Dawid is suggesting an ISO-zulu date format with nanoseconds included. Is that right?

The challenge with nanoseconds in JSON is that JSON is "defined" by JavaScript, in which there is no integer type, and a lot of JSON parsers can't read more than 32 bit integers. That doesn't apply to us as long as we're using Jackson (or some other tolerably capable JSON serde).

shevek-google: I have a vague memory that Jackson has an annotation for the date format string to use, which since we're already formatting an Instant, would let us include nanos? But I also feel like I misunderstand Dawid's question.

aleofreddi: For this endpoint, Ranger API returns these fields in milliseconds. To keep it consistent with the source, I'd rather leave these as millis from epoch.

aleofreddi:
As discussed offline with @dawidxc, I'll open a new PR to remove all the Ranger API models in favor of dumping the raw API response, and move the parsing to the permissions-migration command.

PR #518, opened 2024-08-14
[b/359499526] Dump raw Ranger API entries to output jsonl files.

aleofreddi:


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ranger/RangerClient.java
```diff
@@ -199,7 +132,19 @@ private <T> T doHttpGet(String path, Map<String, String> params, Class<T> bodyCl
...
+        }
+        Object items = payload.get(payloadProperty);
+        if (!(items instanceof List)) {
+          throw new RangerException(
+              "Expected an array value for key "
+                  + payloadProperty
+                  + " in Ranger response, got "
+                  + items.getClass());
+        }
+        return ImmutableList.copyOf((List<Object>) items);
```
shevek-google, line 147: Should this be a `List<?>`, not a `List<Object>`, since we don't actually know what type is in it, but that type does exist, and is fixed?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ranger/RangerConnector.java
```diff
@@ -127,14 +123,14 @@ protected String toCallDescription() {
...
     DumpGroupsTask() {
       super(GroupsFormat.ZIP_ENTRY_NAME);
     }
 
     @Override
-    protected Iterator<Group> dataIterator(@Nonnull RangerClientHandle handle) {
+    protected Iterator<Object> dataIterator(@Nonnull RangerClientHandle handle) {
       return new RangerPageIterator<>(
           page -> {
             try {
```
shevek-google, line 136: Should the try/catch be in RangerPagerIterator.computeNext()? It would have to use its own FunctionalInterface, but that's no cost. It would make all these callers much cleaner, and reduce the number (if not the call-frequency) of 'throw new RuntimeException' calls, which is generally an antipattern.

Might be out of the scope of this PR though, as you didn't change that.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ranger/RangerConnector.java
```diff
@@ -226,22 +222,22 @@ protected String toCallDescription() {
...
+    protected abstract Iterator<Object> dataIterator(@Nonnull RangerClientHandle handle);
 
     @Override
     protected Void doRun(TaskRunContext context, @Nonnull ByteSink sink, @Nonnull Handle handle)
         throws Exception {
       RangerClientHandle rangerClientHandler = (RangerClientHandle) handle;
       LOG.info("Writing to '{}' -> '{}'", getTargetPath(), sink);
       try (Writer writer = sink.asCharSink(StandardCharsets.UTF_8).openBufferedStream()) {
-        for (Iterator<T> iterator = dataIterator(rangerClientHandler); iterator.hasNext(); ) {
+        for (Iterator<Object> iterator = dataIterator(rangerClientHandler); iterator.hasNext(); ) {
```
shevek-google, line 240: I looked briefly at this to see whether it _had_ to use Iterator, and technically, we could replace this with another type/class/pattern, even a RangerPagedCollection.forEach() pattern which would not require the exception transition. But that feels out of scope, so I'm leaving it here as an observation alone.

shevek-google (commented):
I think this is good. I had observations for the curious.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ranger/RangerConnector.java
aleofreddi: Good idea - done.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ranger/RangerClient.java
aleofreddi: Done.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ranger/RangerConnector.java
aleofreddi: The for loop body can throw `IOException` and `JsonProcessingException` - migrating to a `forEach` method (in fact, we could also reuse Iterator's `forEachRemaining`) would need another functional interface that mimics `Consumer` allowing it to throw, or adding try/catch exceptions in place.

Given that such a functional interface would be used only once and that the current form is quite simple to read, I prefer to keep it as is.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ranger/RangerPageIterator.java
```diff
@@ -40,13 +41,19 @@ static Page create(int offset, int limit) {
     abstract int limit();
   }
 
-  private final Function<Page, List<T>> fetcher;
+  @FunctionalInterface
+  public interface RangerFetcher<T> {
+
+    List<T> apply(Page page) throws RangerException;
```
paolomorandini, line 47: Nit: consider naming this method `get` or `fetch`

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ranger/RangerPageIterator.java
```diff
@@ -56,7 +63,12 @@ static Page create(int offset, int limit) {
   @Override
   protected T computeNext() {
     if (pageIterator == null || !pageIterator.hasNext() && !lastPage) {
-      List<T> entries = fetcher.apply(Page.create(offset, pageSize));
+      List<T> entries;
+      try {
+        entries = fetcher.apply(Page.create(offset, pageSize));
+      } catch (RangerException e) {
+        throw new RuntimeException("Failed to fetch data from Ranger", e);
```
paolomorandini, line 70: Nit: consider including `offset` and `pageSize` in the exception message

paolomorandini (approved):
Thank you!

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ranger/RangerPageIterator.java
aleofreddi: Renamed to `fetch`, done.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ranger/RangerPageIterator.java
aleofreddi: Ok, done.

PR #521, opened 2024-08-19
Remove dwh-migration-client

dawidxc:
I'm removing the `dwh-migration-client`, because it seems unused.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,9 +12,6 @@ distributions {
                 exclude "**/__pycache__"
                 into "client"
             }
-            from("bin/dwh-migration-client") {
-                into "bin"
-            }
             project(":dumper:app").afterEvaluate {
                 from it.tasks.installPublishedDist
             }
--- a/bin/dwh-migration-client
+++ b/bin/dwh-migration-client
@@ -1,3 +1,0 @@
-#!/bin/sh -ex
-
-exec dwh-migration-client "$@"


build.gradle
```diff
@@ -12,9 +12,6 @@ distributions {
                 exclude "**/__pycache__"
                 into "client"
             }
-            from("bin/dwh-migration-client") {
```
misolt, line 15: What was the original purpose of this client?

ajwelch4: I vote to remove this.

This used to just pass a Python script in the `client` directory to the `python3` interpreter: https://github.com/google/dwh-migration-tools/blob/5d973352a04b41da1e9c36c1063ff506c6ed6f03/bin/dwh-migration-client#L3

Then, due to a whole bunch of reasons, the Python client morphed into a more full-fledged Python project that was installed via `pip` (or your preferred Python package manager) which placed an executable script named [`dwh-migration-client`](https://github.com/google/dwh-migration-tools/blob/1e1c153373d319fb67619820c8b25f0f241972cb/client/setup.py#L44-L47) on the `PATH`. This `bin/dwh-migration-client` script was then modified to call the `dwh-migration-client` script on the `PATH` which IMHO was unnecessary and confusing but I believe it was kept to maintain backwards compatibility or parity with the dumper's `bin/dwh-migration-dumper` script.

Nowadays, the Python client is [installed](https://github.com/google/dwh-migration-tools/tree/main/client#installation) via `pip` (or your preferred Python package manager) which places an executable script named [`bqms_run`](https://github.com/google/dwh-migration-tools/tree/main/client#running-using-bqms-run-directly) on the `PATH`: https://github.com/google/dwh-migration-tools/blob/main/client/pyproject.toml#L8. There is also an opinionated [`run.sh`](https://github.com/google/dwh-migration-tools/tree/main/client#running-using-runsh) example script that calls `bqms_run`. It seems this `bin/dwh-migration-client` script was lost in the shuffle and I think it should just be deleted as opposed to modified to call `bqms_run` or the example `run.sh` script.

Ultimately, this whole Python client needs to go away.

ajwelch4 (approved):

PR #529, opened 2024-08-23
[b/353664125] Extract statuses of services

dawidxc:


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hadoop/HadoopMetadataConnector.java
```diff
@@ -82,13 +100,40 @@ public void addTasksTo(@Nonnull List<? super Task<?>> out, @Nonnull ConnectorArg
...
     addTasksForSingleLineScripts(out);
+    generateServiceScripts(out);
   }
 
   private void addTasksForSingleLineScripts(List<? super Task<?>> out) {
     HadoopScripts.extractSingleLineScripts()
         .forEach((scriptName, scriptFile) -> out.add(new BashTask(scriptName, scriptFile)));
   }
 
+  private void generateServiceScripts(List<? super Task<?>> out) {
```
misolt, line 111: I would suggest providing the finished tasks as the return value, rather than modifying the `out` parameter.

If this should have a measurable effect on performance and the copy would need to be avoided, `ImmutableList` can be replaced with a different type.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hadoop/HadoopMetadataConnector.java
```diff
@@ -82,13 +100,40 @@ public void addTasksTo(@Nonnull List<? super Task<?>> out, @Nonnull ConnectorArg
...
+    SERVICE_NAMES.stream()
+        .flatMap(
+            serviceName ->
+                Stream.of(
+                    generateBashTask(
```
misolt, line 116: Optional: The `String`s here could also be converted to a class that would represent a service and contain the methods `toServiceStatus` and `toSystemCtlStatus`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hadoop/HadoopMetadataConnector.java
dawidxc: Done.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hadoop/HadoopMetadataConnector.java
dawidxc: The code is tiny - 6 lines of code. I don't predict any further processing of these scripts in Java, so IMHO investing more time in representing the process as Java entities instead of Strings will not help in the future.

misolt (approved):

PR #531, opened 2024-08-23
[b/357517011] Extract filenames from local filesystem

dawidxc:


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hadoop/LocalFilesystemScanCommandGenerator.java
```diff
@@ -0,0 +1,52 @@
...
+import java.io.IOException;
+import java.net.URL;
+
+public class LocalFilesystemScanCommandGenerator {
+  private static final Escaper SIMPLE_BASH_ESCAPER =
+      Escapers.builder().addEscape('\'', "\\'").addEscape('\\', "\\\\").build();
+
+  public static String generate() {
+    try {
+      URL resourceUrl = Resources.getResource("hadoop-extraction/search-expressions.txt");
```
misolt, line 35: The list looks small, would it be practical to move it here and remove the txt resource? This would avoid the need for custom parsing

dawidxc: The size of the list doesn't matter. The point of the text resource file was to be able to edit the list comfortably without looking at the Java syntax. The custom parsing logic is only a few lines of code, so it doesn't affect anything significantly and it is likely that the parsing logic will be changed very rarely, while the expressions may be updated much more often.

I've changed the format to the java list.

dawidxc:
I could restrict the search to some paths (for example, `/usr`, `/etc`, `/var`), where these files are most likely to be found, but then if there is some custom environment, we would not discover these files' locations, which is undesirable. I understand that it might take hours. I don't see a way around it (besides not including such a task in the discovery process).

Shell is used only as a temporary solution, we can rewrite the shell scripts to another solution later.

I've tested it only on the small server, where scanning the whole filesystem takes a few minutes.

cyulysses-corp (approved):

PR #534, opened 2024-08-23
[b/356461225] Add HDFS progress logging

dawidxc:


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hdfs/ScanContext.java
```diff
@@ -152,24 +176,22 @@ String getFormattedStats() {
...
-            .append("\nAvg time per doc: ")
-            .append(timeSinceScanBegin.dividedBy(numFilesDivisor).toMillis() + "ms\n")
-            .append("\n[/HDFS Permission extraction stats]");
-
-    return sb.toString();
+    return new StringBuilder()
+        .append("[HDFS Permission extraction stats]")
+        .append("\nTotal: num files&dirs: " + numFiles)
+        .append("\n       num dirs found: " + numDirs)
+        .append("\n       num dirs walkd: " + numDirsWalked)
```
misolt, line 183: nit: walkd -> walked

dawidxc: Fixed.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hdfs/ScanContext.java
```diff
@@ -152,24 +176,22 @@ String getFormattedStats() {
...
+        .append("[HDFS Permission extraction stats]")
+        .append("\nTotal: num files&dirs: " + numFiles)
+        .append("\n       num dirs found: " + numDirs)
+        .append("\n       num dirs walked: " + numDirsWalked)
+        .append("\nTotal File Size: " + accumulatedFileSize)
+        .append("\nAvg File Size: " + accumulatedFileSize / numFilesDivisor)
+        .append("\nTotal time: ")
+        .append(timeSinceScanBegin.getSeconds() + "s")
+        .append("\nTotal time in listStatus(..): ")
+        .append(timeSpentInListStatus.getSeconds() + "s")
```
shevek-google, line 189: This should probably be a chained append(), not a String-+ in the middle, as that creates an extra StringBuilder and String.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hdfs/ScanContext.java
```diff
@@ -116,6 +130,16 @@ void endWalkDir(FileStatus dir, long nFiles, long nDirs, long accumFileSize) thr
     }
   }
 
+  private void logProgress() {
+    long now = System.currentTimeMillis();
+    if (now - lastLogTime > LOG_PROGRESS_DELAY.toMillis()) {
+      lastLogTime = now;
+      long percentFinished =
+          Longs.constrainToRange(this.numDirsWalked * 100 / totalDirectoryCount, 0, 99);
+      partialProgressLogger.logProgress("HDFS scan " + percentFinished + "% completed");
```
shevek-google, line 139: A lot of this looks very similar to the shared/common code in [Concurrent]RecordProgressMonitor. Would it be possible to use that?

shevek-google (changes requested):
Could this use https://github.com/google/dwh-migration-tools/blob/main/dumper/lib-common/src/main/java/com/google/edwmigration/dumper/plugin/ext/jdk/progress/ConcurrentRecordProgressMonitor.java instead, particularly if the 'total' field was updated to be a mutable LongAdder? If we want an ETA, I think adding that to ConcurrentProgressMonitor would be a lovely feature update.

Note that System.currentTimeMillis() is a very expensive call, which does not show up in a stack-sampling profiler, because it's not a safepoint; BlockProgressMonitor is designed to reduce that overhead.

vladislav-sidorovich:
Actual PR is here https://github.com/google/dwh-migration-tools/pull/673

PR #536, opened 2024-08-23
Add sanity check for script content

dawidxc:
Make sure that the script was successfully written to the disk before executing it.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hadoop/HadoopScripts.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hadoop/HadoopScripts.java
@@ -25,6 +25,7 @@ import static org.apache.hadoop.thirdparty.com.google.common.collect.Maps.immuta
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.io.Resources;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hadoop/HadoopScripts.java
```diff
@@ -126,10 +128,20 @@ static Path create(String scriptFilename, byte[] scriptBody) throws IOException
...
     scriptPath.toFile().setExecutable(true);
     return scriptPath;
   }
 
+  private static void sanityCheckFileContents(Path scriptPath, byte[] expectedScriptBody)
+      throws IOException {
+    byte[] scriptBody = Files.readAllBytes(scriptPath);
+    Preconditions.checkState(
+        Arrays.equals(expectedScriptBody, scriptBody),
+        "Sanity check failed: File contents doesn't match the script body. Path='%s'.",
```
misolt, line 141: What are some potential causes for this contents mismatch?

dawidxc: Hard to tell. Some unexpected behavior, where writing would succeed, but reading would yield a different result. It's a sanity check to make sure that the script being executed contains the expected body. The scripts are very small, so it should not affect the performance.

kporzycki: I expect that if `Files.write` finishes without an exception, then the file has been written correctly and the check is not needed.

Unless we can create a reasonable unit test case that would trigger this scenario, I suggest that this check is not needed.

sten1ee (approved):

PR #540, opened 2024-08-26
[b/361371764] Check minimal Java version at startup

dawidxc:


shevek-google:
It may be worth considering:
try { Runtime.version().feature() ... } catch (NoSuchMethodError e) { AccessController.doPrivileged(System.getProperty("java.specification.version")) ... } because that limits both the set of versions which need to be handled in the catch clause and uses builtin specifications. The try will run in everything after JDK10. There's a way to build this using e.g. multiversion JARs, too which makes it much more reliable, and will elegantly handle JDK9.

shevek-google (commented):
It may be worth considering:
try { Runtime.version().feature() ... } catch (NoSuchMethodError e) { AccessController.doPrivileged(System.getProperty("java.specification.version")) ... } because that limits both the set of versions which need to be handled in the catch clause and uses builtin specifications. The try will run in everything after JDK10. There's a way to build this using e.g. multiversion JARs, too which makes it much more reliable, and will elegantly handle JDK9.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MinimumJavaVersionChecker.java
```diff
@@ -0,0 +1,52 @@
...
+import org.slf4j.LoggerFactory;
+
+public class MinimumJavaVersionChecker {
+  private static final Logger LOG = LoggerFactory.getLogger(MinimumJavaVersionChecker.class);
+
+  static void check(String version) {
+    if (!version.startsWith("1.")) {
+      return;
+    }
+    Pattern oldVersionPattern = Pattern.compile("1\\.([0-9]+)");
```
shevek-google, line 31: This is a non-anchored pattern used with find(), so it will match things like 21.7 and return 1.7

shevek-google (changes requested):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MinimumJavaVersionChecker.java
dawidxc: It will not match `21.7`, because there is `version.startsWith("1.")` check in the `if` above.

dawidxc: I've added the anchor.

dawidxc:
I've added checking the version via `Runtime#version` method first.

dawidxc:
Closing as @shevek-google pointed out that the code won't even run on Java < 8, because the dumper is compiled to target bytecode Java 8.

PR #541, opened 2024-08-27
ParallelTaskGroup fixes

misolt:
- use the correct ExecutorManager method, rather than suppressing a warning
- remove a comment with incorrect information about ParallelTaskGroup being used only for Redshift (it is also used for Greenplum)
- (x2) replace `throws Exception` with a more precise declaration
- remove redundant extension of `Callable`
- remove redundant generics
- add missing annotations

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/ParallelTaskGroup.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/ParallelTaskGroup.java
@@ -43,7 +43,7 @@ public class ParallelTaskGroup extends TaskGroup {
     super.addTask(task);
   }
 
-  private static class TaskRunner<T> implements Callable<T> {
+  private static class TaskRunner<T> {
 
     private final TaskRunContext context;
     private final Task<T> task;
@@ -55,7 +55,6 @@ public class ParallelTaskGroup extends TaskGroup {
       this.printer = printer;
     }
 
-    @Override
     public T call() throws IOException {
       T result = context.runChildTask(task);
       TaskState state = context.getTaskState(task);
@@ -73,7 +72,8 @@ public class ParallelTaskGroup extends TaskGroup {
     // We safely publish the CSVPrinter to the ExecutorManager.
     try (ExecutorManager executorManager = new ExecutorManager(context.getExecutorService())) {
       for (Task<?> task : getTasks()) {
-        executorManager.execute(new TaskRunner<>(context, task, printer));
+        TaskRunner<?> runner = new TaskRunner<>(context, task, printer);
+        executorManager.execute(runner::call);
       }
     }
     // We now, by the t-w-r, safely collect the CSVPrinter from the sub-threads.
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/ParallelTaskGroup.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/ParallelTaskGroup.java
@@ -22,6 +22,8 @@ import com.google.edwmigration.dumper.plugin.ext.jdk.concurrent.ExecutorManager;
 import java.io.IOException;
 import java.util.concurrent.ExecutionException;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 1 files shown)


dawidxc (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/ParallelTaskGroup.java
```diff
@@ -66,15 +68,16 @@ public T call() throws Exception {
...
+      @Nonnull TaskRunContext context, @Nonnull CSVPrinter printer, @Nonnull Handle handle)
+      throws ExecutionException, InterruptedException {
     // Throws ExecutionException if any sub-task threw. However, runChildTask() is nothrow, so that
     // never happens.
     // We safely publish the CSVPrinter to the ExecutorManager.
     try (ExecutorManager executorManager = new ExecutorManager(context.getExecutorService())) {
       for (Task<?> task : getTasks()) {
-        executorManager.submit(new TaskRunner<>(context, task, printer));
+        TaskRunner runner = new TaskRunner(context, task, printer);
+        executorManager.execute(runner::call);
```
shevek-google, line 80: The only real purpose of TaskRunner was to implement Callable/Runnable. This ::call constructs a new, anonymous Callable/Runnable which can never have a useful toString() and doubles the number of allocations performed by this method.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/ParallelTaskGroup.java
```diff
@@ -42,21 +44,21 @@ public void addTask(Task<?> task) {
...
+        @Nonnull TaskRunContext context, @Nonnull Task<?> task, @Nonnull CSVPrinter printer) {
       this.context = context;
       this.task = task;
       this.printer = printer;
     }
 
-    @Override
-    public T call() throws Exception {
-      T result = context.runChildTask(task);
+    public @Nullable Object call() throws IOException {
```
shevek-google, line 60: It might not be Nullable. Whether it's Nullable or not can be captured from Task<T> by a type-use annotation on T, rather than being re-specified here.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/ParallelTaskGroup.java
misolt: Regarding `toString`: The current version returns something like `TaskRunner@1234`. Should I add a `toString` to the callable being passed here?

misolt: About the number of allocations - I understand that this means having two allocations per loop iteration instead of one, not including work done inside `execute` or `call` itself. I looked for evidence to support this, but I don't really get anything like an object initialization (e.g. `<init>`) in the results of compiling similar examples.

The `execute` here wants a call site, which we provide. Neither side needs to create a new object wrapping the reference.

Maybe this worked differently before Java 8?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/ParallelTaskGroup.java
shevek-google: It's not a reflective call. It's a compiler type-system plugin.

misolt: I assumed the intention was to get the annotation at run-time. At compile-time it seems like there's no information available, because the type parameter to the generic `TaskRunner` is always `?`.

shevek-google: It _could_ be inspected at runtime, but I don't know if I've ever seen anyone do that. This is all about compile-time correctness, a shift-left of a runtime failure. That's why I'm sometimes a bit jumpy about these subtleties: Compile-time error detection is ideal, and I'd love to have the entire team understand more about how effective it can be.

PR #549, opened 2024-08-29
Cleanup loggers in HDFS connector

dawidxc:
As per convention, the logger is instantiated in the class it is used in (with rare exceptions like progress logger).

Also, use placeholder instead of string concatenation in logger.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hdfs/SingleDirScanJob.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hdfs/SingleDirScanJob.java
@@ -55,10 +55,11 @@ class SingleDirScanJob implements Callable<Void> {
         }
       }
       scanCtx.endWalkDir(dir, numFiles, numDirs, accumFileSize);
-    } catch (Exception exn) {
+    } catch (Exception e) {
       LOG.error(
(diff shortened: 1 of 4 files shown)


misolt (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hdfs/SingleDirScanJob.java
```diff
@@ -55,10 +55,11 @@ public Void call() {
...
       }
       scanCtx.endWalkDir(dir, numFiles, numDirs, accumFileSize);
-    } catch (Exception exn) {
+    } catch (Exception e) {
       LOG.error(
-          "Unexpected exception while scanning HDFS folder " + dir.getPath().toUri().getPath(),
-          exn);
+          "Unexpected exception while scanning HDFS folder '{}'",
+          dir.getPath().toUri().getPath(),
+          e);
```
shevek-google, line 62: Are you sure this change to this call does what we want it to? The three-argument error() call does not take a Throwable as the third argument. The two-argument does take a Throwable as the second argument.

dawidxc: Slf4j will replace the placeholder `{}` with the path and then it will additionally process the exception, because it is the last argument.

https://www.slf4j.org/faq.html#paramException

PR #552, opened 2024-09-03
[b/364035944] Prevent a NullPointerException where a result's toString has returned null

misolt:
If there is an object in the ResultSet, for which `toString` returns null, the task involving the object may fail, depending on the specific class of this object.

In all cases where a ResultSet element object's `toString` would return null, replace the object with null, which will be written as an empty csv item without error.


kpiotrowski (approved):

shevek-google (commented):
This method can still return null. Is it the case that CSVFormat can handle null objects, but not objects with a null toString()?

Do we know what object it was that violated toString()'s rules? Can we put that in a comment?

shevek-google (dismissed):

PR #554, opened 2024-09-09
Extract Cloudera metaconnector in-dump paths to SPI.

damian-compilerworks:


dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/test/DummyOutputHandle.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+package com.google.edwmigration.dumper.application.dumper.test;
+
+import com.google.common.io.ByteSink;
+import com.google.edwmigration.dumper.application.dumper.io.OutputHandle;
+import com.google.edwmigration.dumper.application.dumper.task.AbstractTaskTest.MemoryByteSink;
+import java.io.IOException;
+import javax.annotation.Nonnull;
+
+public class DummyOutputHandle implements OutputHandle {
```
mjquinn-google, line 25: Would a better name for this not be `InMemoryOutputHandle`?

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/test/InMemoryByteSinkFactory.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.google.edwmigration.dumper.application.dumper.io.OutputHandle;
+import com.google.edwmigration.dumper.application.dumper.io.OutputHandleFactory;
+import java.util.HashMap;
+import java.util.Map;
+import javax.annotation.Nonnull;
+
+public class InMemoryByteSinkFactory implements OutputHandleFactory {
+
+  private final Map<String, DummyOutputHandle> producedHandlers = new HashMap<>();
```
mjquinn-google, line 27: Related to previous comment; it's somewhat confusing that at first read, this factory of in-memory byte sinks doesn't actually seem to factorize any such sinks, until one goes and reads `DummyOutputHandle` to understand that it is just a wrapper around an in-memory byte sink.

mjquinn-google (approved):

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/test/DummyOutputHandle.java
damian-compilerworks: Agreed.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/test/InMemoryByteSinkFactory.java
damian-compilerworks: You're right. I renamed this class to `InMemoryOutputHandleFactory` to make it clear. It's debatable whether `InMemoryOutputHandle` class should at all be visible outside of the factory, as we might be able to perform the test assertions on concrete handles via the factory object itself, but I left it as is since it gives more flexibility and is a much more straight design.

PR #555, opened 2024-09-09
Provide a more descriptive error message when driver JAR does not exist

misolt:


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/AbstractJdbcConnector.java
```diff
@@ -93,6 +91,19 @@ public ClassLoader run() throws Exception {
...
   }
 
+  private static URI getDriverUri(String driverPath) {
+    File result = new File(driverPath);
+    String absolutePath = result.getAbsolutePath();
+    if (!result.exists()) {
+      String message = String.format("File does not exist: '%s'", absolutePath);
+      throw new IllegalArgumentException(message);
+    } else if (!result.isFile()) {
+      String message = String.format("Not a driver JAR file: %s", absolutePath);
```
shevek-google, line 101: "Not a regular file"? - that's what we know. It might be a directory.

misolt: Done, I changed the original message to provide more information.

dominik-klimczak (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/AbstractJdbcConnector.java
misolt: It would also help UX if we're more concise when printing exceptions here. This IllegalArgumentException will be wrapped in an SQLException later. Then the SQLException causes the connector to fail, which causes the Dumper to fail, both failures causing us to print both the SQLException and the Illegal-Arg cause, for a total of 4 copies of the error message in the logs.

We could replace the Illegal-Argument with a checked usage exception, drop the SQLException wrapper and not catch exceptions in connector, leaving their handling to MetadataDumper alone.

PR #558, opened 2024-09-20
[b/368572924] Fix: missing data of public synonym objects

misolt:
Fix the issue by removing the filter

### DIFF
--- a/dumper/app/src/main/resources/oracle-stats/cdb/native/db-objects-synonym-public.sql
+++ b/dumper/app/src/main/resources/oracle-stats/cdb/native/db-objects-synonym-public.sql
@@ -21,11 +21,18 @@ SELECT
   -- "Count" is kept for backwards compatibility
   CASE WHEN A.object_name LIKE '/%' THEN 0
     WHEN A.object_name LIKE 'BIN$%' THEN 0
-    ELSE 1 END "Count"
+    ELSE 1 END "Count",
+  C.table_owner "TableOwner"
 FROM cdb_objects A
-  WHERE A.object_type = 'SYNONYM'
-    AND A.owner = 'PUBLIC'
+LEFT OUTER JOIN cdb_synonyms C
+  ON A.owner = C.owner
+  AND C.table_owner IS NOT NULL
+  AND A.object_name = C.synonym_name
+  AND A.con_id = C.con_id
+WHERE A.object_type = 'SYNONYM'
+  AND A.owner = 'PUBLIC'
 GROUP BY
   A.con_id,
   A.editionable,
-  A.object_name
+  A.object_name,
+  C.table_owner
--- a/dumper/app/src/main/resources/oracle-stats/cdb/native/db-objects-synonym-public.sql
+++ b/dumper/app/src/main/resources/oracle-stats/cdb/native/db-objects-synonym-public.sql
@@ -19,9 +19,7 @@ SELECT
   A.editionable "Editionable",
   A.object_name "ObjectName",
   -- "Count" is kept for backwards compatibility
-  CASE WHEN A.object_name LIKE '/%' THEN 0
-    WHEN A.object_name LIKE 'BIN$%' THEN 0
-    ELSE 1 END "Count",
+  1 "Count",
   C.table_owner "TableOwner"
 FROM cdb_objects A
 LEFT OUTER JOIN cdb_synonyms C
@@ -30,6 +28,8 @@ LEFT OUTER JOIN cdb_synonyms C
   AND A.object_name = C.synonym_name
   AND A.con_id = C.con_id
 WHERE A.object_type = 'SYNONYM'
+  AND A.object_name NOT LIKE '/%'
+  AND A.object_name NOT LIKE 'BIN$%'
   AND A.owner = 'PUBLIC'
 GROUP BY
   A.con_id,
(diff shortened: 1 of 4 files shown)


shevek-google (commented):
I randomly picked this PR to look at for no good reason, and my thought is: Would these be easier to review and/or would the code review process be more reliable if a PR like this came with a proof of behaviour on some Oracle version(s)?

The reviewer has less information than the developer, because they don't have the ability to test it, and to approve its behaviour, their only option is to detangle the code and try to figure it out.

Compare this to the TT system in CW, where every committed code change comes with a YAML file capturing its behaviour in detail, and that behaviour can be reviewed and approved, reducing the code review to a minimum.

Nothing I say here has any bearing on this PR in particular; this is not a review, as-such, but perhaps this is an inspiration to think about the question of using evidence to make reviews faster, easier and more useful, and put some ideas in the back of your mind.

Happy to talk about TT any time you like if that's interesting.

dumper/app/src/main/resources/oracle-stats/cdb/native/db-objects.sql
```diff
@@ -18,7 +18,8 @@ SELECT
   A.object_name "ObjectName",
   A.object_type "ObjectType",
   A.editionable "Editionable",
-  count(1) "Count"
+  count(1) "Count",
+  NULL "TableOwner"
```
kpiotrowski, line 22: I think we shouldn't introduce TableOwner column to db-objects because it does not make sense in this context and won't be used by assessment. TableOwner is requires to filter synonyms only that are ingested to the separate temporary table. After the filtering assessment should take care about merging synonyms to db-objects.

dumper/app/src/main/resources/oracle-stats/cdb/native/db-objects-synonym-public.sql
```diff
@@ -13,44 +13,26 @@
...
-  WHERE C.owner = 'PUBLIC'
-    AND C.table_owner IS NOT NULL
-) D ON B.object_name = D.synonym_name
-  AND B.con_id = D.con_id
-WHERE D.table_owner IS NULL
-    AND B.owner = 'PUBLIC'
+  A.editionable "Editionable",
+  A.object_name "ObjectName",
+  -- "Count" is kept for backwards compatibility
+  CASE WHEN A.object_name LIKE '/%' THEN 0
```
kpiotrowski, line 22: I suggest to remove Count column from the synonyms sql. We can easily add this filtering on the assessment side. This will reduce dumper complexity.

dumper/app/src/main/resources/oracle-stats/cdb/native/db-objects.sql
misolt: The column supplies no data, but on the Assessment end we rely on these two CSVs having the same structure.

Even after the filtering, this column will be present in the tmp table, so we'll need to specify which columns are getting merged and which aren't. I think allowing the redundant column with nulls is a safer option.

dumper/app/src/main/resources/oracle-stats/cdb/native/db-objects-synonym-public.sql
misolt: This is a good idea, but it must wait until Assessment no longer relies on this column. Even though the information in the column can be obtained elsewhere, current-version Assessment could still try to read it and get errors.

kpiotrowski: How about we handle this condition in the same way as it is in db-objects.sql SQL and add `WHERE A.object_name NOT LIKE 'BIN$%' AND A.object_name NOT LIKE '/%'`? We can then keep simple `count(1)` for backward compatibility and remove it later after changing assessment.

dumper/app/src/main/resources/oracle-stats/cdb/native/db-objects.sql
misolt: I think either way this will be a non-trivial decision with some future consequences, so we should discuss it with the team before merging the PR or the CL.

dominik-klimczak: Let's not deviate from the dumper design principles that the tool should be as simple and as reliable as possible. It's better to add extra processing in the service than increase the complexity here.

vladislav-sidorovich: Colleagues, I see that we agreed on the statements:
1. Keep dumper as simple as possible
2. Add additional logic (complexity) on server if needed

Could you please clarify what specific actions follow from this?

dominik-klimczak: The discussion also took place in the ingestion code. The outcome is we don't add a null-filled column "TableOwner" in the dumper.

misolt: The discussion here is a little out of date, the PR is currently on hold because we can't make a release at the moment.

I'll update it before reopening, but it's as Dominik said, the TableOwner will only be present in the synonym-public version of DB Objects.

dumper/app/src/main/resources/oracle-stats/cdb/native/db-objects-synonym-public.sql
misolt: Done, I used the constant `1` instead of counting the items, because without grouping by name the number won't be useful.

kpiotrowski (approved):

PR #560, opened 2024-09-24
Actual query logs dates in summary.

sayuzbas-google:
Output actual query log start and end date in summary section.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
@@ -717,6 +717,10 @@ public class ConnectorArguments extends DefaultArguments {
     return getOptions().has(optionDryrun);
   }
 
+  public boolean isLogConnector() {
+    return getConnectorName().contains("logs");
+  }
+
   @CheckForNull
   @Deprecated
   public String getQueryLogEarliestTimestamp() {
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
@@ -276,8 +276,8 @@ public class MetadataDumper {
     ZonedDateTime actualQueryLogStartDate = QueryLogDates.getQueryLogStartDate();
     ZonedDateTime actualQueryLogEndDate = QueryLogDates.getQueryLogStartDate();
 
-    DateTimeFormatter formatter = DateTimeFormatter.ofPattern("YYYY-MM-DD");
-
+    DateTimeFormatter formatter = DateTimeFormatter.ofPattern("YYYY-MM-DDThh:mm");
+    linePrinter.println(actualQueryLogStartDate + " ");
     if (queryLogStartDate == null && queryLogEndDate == null) {
       linePrinter.println(
           "Query log contains data from '%s' to '%s'",
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
@@ -24,6 +24,7 @@ import com.google.common.base.Stopwatch;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.io.Closer;
 import com.google.common.io.Files;
+import com.google.edwmigration.dumper.application.dumper.SummaryPrinter.SummaryLinePrinter;
 import com.google.edwmigration.dumper.application.dumper.connector.Connector;
 import com.google.edwmigration.dumper.application.dumper.handle.Handle;
 import com.google.edwmigration.dumper.application.dumper.io.FileSystemOutputHandleFactory;
@@ -43,6 +44,8 @@ import java.nio.file.FileSystems;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Clock;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -193,6 +196,7 @@ public class MetadataDumper {
           checkRequiredTaskSuccess(summaryPrinter, state, outputFileLocation);
       logFinalSummary(
           summaryPrinter,
+          arguments,
           state,
           outputFileLength,
           stopwatch,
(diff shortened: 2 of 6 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -717,6 +717,10 @@ public boolean isDryRun() {
     return getOptions().has(optionDryrun);
   }
 
+  public boolean isLogConnector() {
```
vladislav-sidorovich, line 720: Do we really need such method? I think we can use this one `isDefinitionSpecified`

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/JdbcSelectTask.java
```diff
@@ -65,11 +80,20 @@ protected Summary doInConnection(
...
+    updateQueryLogDates(summary);
+    return summary;
   }
 
   @Override
   public String describeSourceData() {
     return createSourceDataDescriptionForQuery(getSql());
   }
+
+  private void updateQueryLogDates(Summary summary) {
```
vladislav-sidorovich, line 93: I'm not sure, but should this method be thread-safe?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
```diff
@@ -265,8 +270,34 @@ private boolean checkRequiredTaskSuccess(
...
 
+  private void outputCorrectLogStartAndEndDates(
+      SummaryLinePrinter linePrinter, ConnectorArguments connectorArguments) {
+
+    ZonedDateTime queryLogStartDate = connectorArguments.getQueryLogStart();
+    ZonedDateTime queryLogEndDate = connectorArguments.getQueryLogEnd();
+    ZonedDateTime actualQueryLogStartDate = QueryLogDates.getQueryLogStartDate();
+    ZonedDateTime actualQueryLogEndDate = QueryLogDates.getQueryLogStartDate();
+
+    DateTimeFormatter DATE_FORMAT =
```
vladislav-sidorovich, line 281: It looks like this object should be defined as `static` and reused across multiple method calls.
Otherwise, if you prefer to keep it local and lazy, I'd suggest moving it under the `if` statement below.

vladislav-sidorovich (changes requested):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
sayuzbas-google: `isDefinitionSpecified` checks if property was specified. Here I want to know, if we are running query log assessment, by checking connector. Even if `query-log-start` is not specified, but it is related to query log connector, we want to output message.

vladislav-sidorovich: I see. Sorry, I misunderstand the `getConnectorName()` call.

Is the containing in name of the connector `logs` robust enough? Maybe it will better to introduce a flag/property to connectors or has an explicit list of connectors?

sayuzbas-google: I checked that all logs connector contains `log` in the name, so they are following that pattern.
Having explicit list connectors, means that, if we add new log connector, that need to be updated as well. Can be option as well. What other developers think? Should I add general set with all log connectors name or do you think there could be better approach to check if connector is log related?

vladislav-sidorovich: If there is an existing naming convention that everyone should follow, that's fine. I just haven't seen any checks in the code that could prevent future bugs.
However, I don't have enough knowledge about the project yet. Let's wait for other opinions.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeLogsConnector.java
```diff
@@ -373,17 +373,21 @@ public void addTasksTo(List<? super Task<?>> out, ConnectorArguments arguments)
...
-    queryLogIntervals.forEach(interval -> addJdbcTask(out, interval, queryHistoryTask));
+    queryLogIntervals.forEach(interval -> addJdbcTask(out, interval, queryHistoryTask, true));
 
     List<TaskDescription> timeSeriesTasks = createTimeSeriesTasks(arguments);
     Duration duration = Duration.ofDays(1);
     ZonedIntervalIterableGenerator.forConnectorArguments(
             arguments, duration, IntervalExpander.createBasedOnDuration(duration))
-        .forEach(interval -> timeSeriesTasks.forEach(task -> addJdbcTask(out, interval, task)));
+        .forEach(
+            interval -> timeSeriesTasks.forEach(task -> addJdbcTask(out, interval, task, false)));
```
sayuzbas-google, line 383: These time series tasks are running with frequency of 1 day. So we should not consider these days as log dates and need to be filter out.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/JdbcSelectTask.java
```diff
@@ -35,6 +37,10 @@ public class JdbcSelectTask extends AbstractJdbcTask<Summary> {
 
   @Nonnull private final TaskCategory taskCategory;
 
+  private boolean shouldCalculateQueryLogDates;
```
sayuzbas-google, line 40: I introduced this variable to differentiate the cases when we should not consider dates as first or last entry. For now it is used in SnowflakeLogsConnector, where we run time series task with frequency of 1 day, so they consider aggregated values. Because that tasks are part of logs connector work, we should filter out that dates. More comment [here](https://github.com/google/dwh-migration-tools/pull/560/files#r1777341288)

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/JdbcSelectTask.java
```diff
@@ -35,6 +37,10 @@ public class JdbcSelectTask extends AbstractJdbcTask<Summary> {
 
   @Nonnull private final TaskCategory taskCategory;
 
+  private boolean calculateQueryLogDates;
```
sayuzbas-google, line 40: Introduced this to differentiate cases when we don't need consider dates as first or last query log entry.
For instance in SnowflakeLogsConnector there are time series tasks, which run by 1 day interval, so dates are coming here and calculating as first entry and last entry, which is incorrect.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/QueryLogDateUtil.java
```diff
@@ -0,0 +1,56 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.edwmigration.dumper.application.dumper.utils;
+
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import javax.annotation.Nonnull;
+
+public class QueryLogDateUtil {
```
dominik-klimczak, line 23: 1. It is not a utility class as the name suggests, it's a shared state class, so please rename it.
2. The state is shared between tasks, so it should be thread-safe (if this is the design we need to follow).
3. Can't we avoid a shared state here and implement the whole change in a "functional" way. Each task returns a summary that can contain the first and the last entry. At the end, we just need to find MIN and MAX from all summaries.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
```diff
@@ -24,6 +24,7 @@
 import com.google.common.collect.ImmutableMap;
```
misolt, line 24: @sayuzbas-google Can this PR be closed?

sayuzbas-google: Yes, closing

PR #567, opened 2024-10-02
Output first and last query log entries for Teradata in summary section using shared state class

sayuzbas-google:
Finding first and last query log entries for Teradata and output in summary section of Dumper.

Current summary section:
Screenshot 2024-10-04 at 14 44 48

Summary section with query log entries:
Screenshot 2024-10-04 at 14 15 57

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
@@ -44,6 +44,8 @@ import java.nio.file.FileSystems;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Clock;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -61,6 +63,9 @@ public class MetadataDumper {
 
   private static final Pattern GCS_PATH_PATTERN =
       Pattern.compile("gs://(?<bucket>[^/]+)/(?<path>.*)");
+
+  private static DateTimeFormatter OUTPUT_DATE_FORMAT =
+      DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm").withZone(ZoneOffset.UTC);
 
   public boolean run(String... args) throws Exception {
     ConnectorArguments arguments = new ConnectorArguments(JsonResponseFile.addResponseFiles(args));
@@ -274,11 +279,13 @@ public class MetadataDumper {
     }
 
     linePrinter.println(
-        "First query log entry is '%s' and last query log entry is '%s'",
-        QueryLogSharedState.queryLogEntries.get(
-            QueryLogSharedState.QueryLogEntry.QUERY_LOG_FIRST_ENTRY),
-        QueryLogSharedState.queryLogEntries.get(
-            QueryLogSharedState.QueryLogEntry.QUERY_LOG_LAST_ENTRY));
+        "The first query log entry is '%s' and the last query log entry is '%s'",
+        QueryLogSharedState.queryLogEntries
+            .get(QueryLogSharedState.QueryLogEntry.QUERY_LOG_FIRST_ENTRY)
+            .format(OUTPUT_DATE_FORMAT),
+        QueryLogSharedState.queryLogEntries
+            .get(QueryLogSharedState.QueryLogEntry.QUERY_LOG_LAST_ENTRY)
+            .format(OUTPUT_DATE_FORMAT));
   }
 
   private void logFinalSummary(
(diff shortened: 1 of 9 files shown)


sayuzbas-google:
@misolt could you please review it? I implemented shared state approach and applied it for Teradata logs connectors for now, as this approach was the most voted one during discussion. If it is fine, later can apply for the rest of connectors.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
```diff
@@ -61,6 +64,9 @@ public class MetadataDumper {
   private static final Pattern GCS_PATH_PATTERN =
       Pattern.compile("gs://(?<bucket>[^/]+)/(?<path>.*)");
 
+  private static DateTimeFormatter OUTPUT_DATE_FORMAT =
+      DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm").withZone(ZoneOffset.UTC);
```
paolomorandini, line 68: Consider including the timezone in the pattern or in the log message to let user know UTC is assumed.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
```diff
@@ -281,6 +305,10 @@ private void logFinalSummary(
                   + state.getTasksReports().stream()
                       .map(taskReport -> taskReport.count() + " " + taskReport.state())
                       .collect(joining(", ")));
+          // For now, it will return true only for TeradataLogsConnector and Terada14LogsConnector
```
misolt, line 308: Why doesn't it return `true` for all Logs Connectors?

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/QueryLogSharedStateTest.java
```diff
@@ -0,0 +1,94 @@
...
+public class QueryLogSharedStateTest {
+
+  @Before
+  public void beforeEachTest() {
+    QueryLogSharedState.clearQueryLogEntries();
+  }
+
+  @Test
+  public void queryLogFirstEntryUpdatedSuccessfully() {
+    // Arrange
```
misolt, line 39: The `// Arrange` comment can be omitted in tests, because it can be assumed that the first section of the test will be `Arrange`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
sayuzbas-google: Because for now I implemented shared state logic only for Teradata. I can make all log connectors return `true`, it will work because of this line `if (QueryLogSharedState.queryLogEntries.size() == 0) {
      return;
    }` in `outputFirstAndLastQueryLogEnries` method. Just wanted to be consistent with main logic for this change.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
```diff
@@ -265,9 +272,26 @@ private boolean checkRequiredTaskSuccess(
     return true;
   }
 
+  private void outputFirstAndLastQueryLogEnries(SummaryLinePrinter linePrinter) {
```
misolt, line 275: It would be helpful to add tests for the logic inside this method

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
misolt: It should be consistent with the connector output type - if it retrieves query logs, then a method called `isLogConnector` should return `true`, in all other cases it should be `false`. I think the simplest fix is to rename the method.
For example, if no extension is planned for the future, then `isTeradataLogConnector` would be better, otherwise `shouldOutputFirstAndLastQueryLog` is an option.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/QueryLogSharedState.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+package com.google.edwmigration.dumper.application.dumper;
+
+import com.google.common.annotations.VisibleForTesting;
+import java.time.ZonedDateTime;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+public class QueryLogSharedState {
+  public static final ConcurrentMap<QueryLogEntry, ZonedDateTime> queryLogEntries =
```
misolt, line 25: The concurrent collection suggests that this class is designed for use from multiple threads. If this is the case, please document how this is done - is it fully thread-safe and what rules need to be followed by developers using it. If any guarantees provided by the class aren't obvious then it's also good to document them.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
sayuzbas-google: Potential extension is planned for other connectors. So if we generally go with this approach, then in future might be implementing for others as well. So in that case `shouldOutputFirstAndLastQueryLog` naming would be more suitable. Will change to that.

PR #572, opened 2024-10-04
Move license overrides to a util class

misolt:
### Description
The list of license links has been growing gradually through the project lifetime.

Currently, `dumper/app`'s build script has 1000+ lines, 80% of which is used for the license links mapping.
Declutter the main script by moving the license links to a separate file in a part of the project more related to its purpose.

### Notes
The new Groovy file had to be force-added, because of a .gitignore issue:
- the package name is `com.google.edwmigration.dumper.build.licensereport`
- the project's .gitignore has [a rule](https://github.com/google/dwh-migration-tools/blob/412897cf34fa90355783fb7f27523948cab32a90/.gitignore#L1) which simply says `build`.

It would be nice to replace this rule with something that refers to specific "build" directories instead.


buildSrc/src/main/groovy/com/google/edwmigration/dumper/build/licensereport/OverridesUtil.groovy
```diff
@@ -0,0 +1,794 @@
+package com.google.edwmigration.dumper.build.licensereport
+
+class OverridesUtil {
+
+    static Map<String, Map<String, String>> projectOverrides() {
```
vladislav-sidorovich, line 5: I would suggest to return an `immutable` map.

vladislav-sidorovich (approved):

buildSrc/src/main/groovy/com/google/edwmigration/dumper/build/licensereport/OverridesUtil.groovy
misolt: It's safe to modify this map, because the method creates a new copy each time. The api design with a Map is not ideal, but I'm not sure that making it immutable will solve the problems.

PR #590, opened 2024-10-14
Class design cleanup for Snowflake Logs

misolt:
This is a similar change to https://github.com/google/dwh-migration-tools/pull/586 but for the `snowflake-logs` connector.

Limit visibility and overriding of `SnowflakeLogsConnector`, replace unnecessary overrides with constructor parameters.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeLogsConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeLogsConnector.java
@@ -40,7 +40,6 @@ import com.google.edwmigration.dumper.application.dumper.task.Task;
 import com.google.edwmigration.dumper.application.dumper.task.TaskCategory;
 import com.google.edwmigration.dumper.plugin.ext.jdk.annotation.Description;
 import com.google.edwmigration.dumper.plugin.lib.dumper.spi.SnowflakeLogsDumpFormat;
-import com.google.errorprone.annotations.ForOverride;
 import java.time.Duration;
 import java.time.ZoneOffset;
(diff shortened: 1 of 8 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeLogsConnector.java
```diff
@@ -163,15 +166,22 @@ public Class<? extends Enum<? extends ConnectorProperty>> getConnectorProperties
...
-  protected String newQueryFormat(@Nonnull ConnectorArguments arguments)
+  private String newQueryFormat(@Nonnull ConnectorArguments arguments)
       throws MetadataDumperUsageException {
     // Docref: https://docs.snowflake.net/manuals/sql-reference/functions/query_history.html
     // Per the docref, Snowflake only retains/returns seven trailing days of logs.
-    return createQueryFromInformationSchema(arguments);
+    switch (inputSource) {
+      case USAGE_THEN_SCHEMA:
+        throw new IllegalArgumentException("Unsupported input source for Snowflake logs.");
+      case USAGE_ONLY:
```
kpiotrowski, line 176: nit: I think the current naming is confusing because it does not reflect the snowflake table name.  `SCHEMA_ONLY` may suggest that only some kind of schema will be extracted but it actually changes source of query logs to `INFORMATION_SCHEMA` table.

misolt: Done, I renamed the enum values and added Javadoc to clarify. New names include the word `source` to distinguish "schema" (information schema) as data source from schema as type of data.

PR #594, opened 2024-10-15
[b/373568138] Change timestamp format used in zip entry files to work on windows

kpiotrowski:
Update `redshift-raw-logs`, `teradata-logs` and `snowflake-logs` connectors to use "yyyy-MM-dd'T'HHmmss'Z'" pattern instead of default `ISO_OFFSET_DATE_TIME`. This format was already used for Redshift `cluster_usage_metrics_`, but not applied for all the other files.

The timestamp suffix is added to each time based file generated by the dumper. Currently we use the `ISO_OFFSET_DATE_TIME` format that includes colon characters. Windows file system does not accept colons in the file names what results in following exception:

```
java.nio.file.InvalidPathException: Illegal char <:> at index 21: ddltext_2024-10-09T14:00:00Z.csv
```

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/ArchiveNameUtil.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/ArchiveNameUtil.java
@@ -23,7 +23,11 @@ import java.time.format.DateTimeFormatter;
 
 public class ArchiveNameUtil {
 
-  public static final String ZIP_ENTRY_SUFFIX = ".csv";
+  private static final String ZIP_ENTRY_SUFFIX = ".csv";
+  private static final DateTimeFormatter ENTRY_FILE_DATE_FORMAT =
+      DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HHmmss'Z'");
+  private static final DateTimeFormatter ARCHIVE_FILE_DATE_FORMAT =
+      DateTimeFormatter.ofPattern("yyyyMMdd'T'HHmmss");
 
   /**
    * Generate the archive file name that includes creation time. Typically this is used with a logs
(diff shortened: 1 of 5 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/ArchiveNameUtil.java
```diff
@@ -36,6 +40,17 @@ public static String getFileNameWithTimestamp(String name, Clock clock) {
     return formatFileName(name, timeSuffix);
   }
 
+  public static String getEntryFileNameWithTimestamp(String prefix, ZonedInterval interval) {
+    return getEntryFileNameWithTimestamp(prefix, interval, ZIP_ENTRY_SUFFIX);
+  }
+
+  public static String getEntryFileNameWithTimestamp(
+      String prefix, ZonedInterval interval, String suffix) {
+    DateTimeFormatter formatter =
```
vladislav-sidorovich, line 49: I think `formatter` can be `static final`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/ArchiveNameUtil.java
```diff
@@ -36,6 +40,17 @@ public static String getFileNameWithTimestamp(String name, Clock clock) {
     return formatFileName(name, timeSuffix);
   }
 
+  public static String getEntryFileNameWithTimestamp(String prefix, ZonedInterval interval) {
+    return getEntryFileNameWithTimestamp(prefix, interval, ZIP_ENTRY_SUFFIX);
+  }
+
+  public static String getEntryFileNameWithTimestamp(
```
vladislav-sidorovich, line 47: Let's add a unit-test for the method with some explanation. To make sure that in 2-3 years from now someone will not decide to change the format back.

vladislav-sidorovich (approved):
In the scope of current issue the changes are ok. However with the current approach we should keep in mind that `getEntryFileNameWithTimestamp` should be called for all the future `*Task` we will add. 

Should we introduce some `persistent` tier and keep this responsibility in this tier instead of keep file naming logic in each `*Task`?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/ArchiveNameUtil.java
kpiotrowski: done, I added simple unit test for this.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/utils/ArchiveNameUtil.java
kpiotrowski: I moved it together with second formatter used for archive name.

PR #598, opened 2024-10-16
Fix bugs in ZonedIntervalIterableTest

misolt:
- The variable `durationOfHour` stores the duration of a day, fix it to store the duration of an hour
- Despite the above, tests didn't fail, because `hourExpander` was also configured to expand to days, not hours - fix this as well
- A constant of 168 **days** is introduced without explanation. A week has 168 **hours**. I'm assuming it's supposed to be 1 week as in the assertion that comes after and the arguments were incorrect as in the assertions before. Fix it by using the duration and expander which refer to hours, not days.
- Remove the `Nonnegative` annotation applied to an `Iterable` parameter. `Iterable`'s don't have a sign.
- Extract tested values to named variables to document their meaning

### DIFF
--- a/dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/ZonedIntervalIterableTest.java
+++ b/dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/ZonedIntervalIterableTest.java
@@ -32,7 +32,6 @@ import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.temporal.ChronoUnit;
 import java.util.Locale;
-import javax.annotation.Nonnegative;
 import javax.annotation.Nonnull;
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -49,7 +48,7 @@ public class ZonedIntervalIterableTest {
   private static final ZonedParser zonedParserStart = ZonedParser.withDefaultPattern(START_OF_DAY);
   private static final ZonedParser zonedParserEnd = ZonedParser.withDefaultPattern(END_OF_DAY);
 
-  private void testIterable(int expectCount, @Nonnegative ZonedIntervalIterable iterable) {
+  private void testIterable(int expectCount, ZonedIntervalIterable iterable) {
     LOG.debug("Testing {}", iterable);
     int actualCount = 0;
     for (ZonedInterval interval : iterable) {
--- a/dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/ZonedIntervalIterableTest.java
+++ b/dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/ZonedIntervalIterableTest.java
@@ -87,8 +87,8 @@ public class ZonedIntervalIterableTest {
     Duration durationOfDay = Duration.ofDays(1);
     IntervalExpander dayExpander = IntervalExpander.createBasedOnDuration(durationOfDay);
 
-    Duration durationOfHour = Duration.ofDays(1);
-    IntervalExpander hourExpander = IntervalExpander.createBasedOnDuration(durationOfDay);
+    Duration durationOfHour = Duration.ofHours(1);
+    IntervalExpander hourExpander = IntervalExpander.createBasedOnDuration(durationOfHour);
 
     testIterable(
         169,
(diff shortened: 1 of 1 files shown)


dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/ZonedIntervalIterableTest.java
```diff
@@ -86,23 +85,25 @@ private void assertIterations(
...
-        ZonedIntervalIterableGenerator.forTimeUnitsUntilNow(24 * 7, durationOfDay, dayExpander));
-    testIterable(
-        8, ZonedIntervalIterableGenerator.forTimeUnitsUntilNow(7, durationOfDay, dayExpander));
-
-    testIterable(
-        1,
+    Duration durationOfHour = Duration.ofHours(1);
+    IntervalExpander hourExpander = IntervalExpander.createBasedOnDuration(durationOfHour);
+    ZonedIntervalIterable weekInHours =
+        ZonedIntervalIterableGenerator.forTimeUnitsUntilNow(24 * 7, durationOfHour, hourExpander);
```
vladislav-sidorovich, line 91: Let's use something like that `TimeUnit.DAYS.toHours(7);` to be more clear what we mean, like in your example with `168` days or hours.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/ZonedIntervalIterableTest.java
```diff
@@ -49,7 +48,7 @@ public class ZonedIntervalIterableTest {
   private static final ZonedParser zonedParserStart = ZonedParser.withDefaultPattern(START_OF_DAY);
   private static final ZonedParser zonedParserEnd = ZonedParser.withDefaultPattern(END_OF_DAY);
 
-  private void testIterable(int expectCount, @Nonnegative ZonedIntervalIterable iterable) {
+  private void testIterable(int expectCount, ZonedIntervalIterable iterable) {
```
vladislav-sidorovich, line 51: Most probably the intention was to use `@Nonnull` instead of `@Nonnegative`.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/ZonedIntervalIterableTest.java
```diff
@@ -86,23 +85,25 @@ private void assertIterations(
...
             ZonedDateTime.of(2020, 1, 1, 0, 0, 0, 0, ZoneId.of("UTC")),
             1,
             durationOfHour,
-            hourExpander));
+            hourExpander);
+    ZonedIntervalIterable dayInHours =
+        ZonedIntervalIterableGenerator.forTimeUnitsUntilNow(24, durationOfHour, hourExpander);
+
+    // Act + Assert
+    testIterable(8, weekInDays);
```
vladislav-sidorovich, line 104: I'm a bit confused. There is 7 days in a week, isn't ?

vladislav-sidorovich (approved):

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/ZonedIntervalIterableTest.java
misolt: Yes, that didn't change ;)
The logic of `ZonedIntervalIterable` which is tested here is a bit odd. The idea is that we have time separated into segments (days or hours) and we receive an arbitrary period of time.

The goal is to select the minimum set of segments that covers our time period. Let's say that the time period is until now (13:30 on Thursday), from exactly a week ago and we would like to use segment size = 1 day.

This is 7 days, but creates 8 segments - we have to select all of this Thursday, all of last Thursday and the 6 one-day segments in between. OTOH it will be 7 segments if run exactly at midnight.

Readability of the test is probably not much improved by how this method call looks a bit like an assertion, with `expected=8, actual=weekInDays`. I'd write these tests differently, but I wanted to limit this PR to bug fixes, without any rewriting.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/ZonedIntervalIterableTest.java
misolt: I prefer the constant here to minimize logic in tests, but you're right that it could be explained more clearly.
I added a clarifying comment.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/ZonedIntervalIterableTest.java
misolt: It could be the case, but it's fine for the test to make `iterable` null - it's an invalid result from generator and it makes the test fail.

PR #619, opened 2024-10-23
[b/368222012] Report AccessControlException(s) in HdfsExtractionTask separately

sten1ee:


kporzycki (approved):
Before merging, please test the changes on an environment where the original problem with denied access can be reproduced.

PR #621, opened 2024-10-24
[b/367261072] Cloudera API integration - main data

vladislav-sidorovich:
Separate connector to dump data from Cloudera specific API.
The main flow is implemented. 
cli interface:
```
cloudera-manager - Dumps metadata from Cloudera Manager.
        --url        The Cloudera's Manager API HTTP(s) endpoint.
        --user       The username for Cloudera's Manager.
        --password   The password for Cloudera's Manager. (Required If not specified as an argument, will use a secure prompt.)
        --cluster    The name of Cloudera's cluster. (Required Only if you need to dump data for a single Cloudera cluster.)
```
Examples:
dump all the clusters cli:
```
--connector cloudera-manager --url https://localhost:7183/api/v55/ --user Bruce --password batman
```
dump single clusters cli:
```
--connector cloudera-manager --url https://localhost:7183/api/v55/ --user Bruce --password batman --cluster cldr2-aw-dl
```
The data:
1. Services in the clusters like Kafka, Spark, Impala and so on
2. Nodes (hosts) per cluster
2.1 CPU, RAM, Disk - total and current usage

The next step: Average utilization for date ranges: now - 1d, now - 7d, now - 30d, now - 90d

Note: ~license check is temporary disabled~ The own DTOs used because of license questions

Please leave your comments, likes and subscribe to the channel for the future content.


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaManagerTask.java
```diff
@@ -0,0 +1,51 @@
...
+package com.google.edwmigration.dumper.application.dumper.connector.cloudera.manager;
+
+import com.google.common.io.ByteSink;
+import com.google.edwmigration.dumper.application.dumper.handle.Handle;
+import com.google.edwmigration.dumper.application.dumper.task.AbstractTask;
+import com.google.edwmigration.dumper.application.dumper.task.TaskRunContext;
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+
+public abstract class AbstractClouderaManagerTask extends AbstractTask<Void> {
```
misolt, line 26: I understand that some new classes in this package were designed for inheritance. For each method, I'd recommend to ether document or prohibit overriding it.

For most methods it should be as quick as making them `private` or `final`, but in such cases where subclasses are supposed to override the method, please document the consequences and reasons for overriding.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerHandle.java
```diff
@@ -0,0 +1,94 @@
...
+    return apiURI.resolve("/");
+  }
+
+  public CloseableHttpClient getHttpClient() {
+    return httpClient;
+  }
+
+  @CheckForNull
+  public synchronized List<ClouderaClusterDTO> getClusters() {
+    return clusters;
```
misolt, line 54: If `clusters` is null I suggest to either return an empty list or fail, whichever fits better.

misolt: More importantly, this allows external code to do:
```
List<?> l1 = handle.getClusters();
l1.clear();
```
Which could be fixed by e.g. making a copy of `clusters`

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerHandle.java
```diff
@@ -0,0 +1,94 @@
...
+  public URI getBaseURI() {
+    return apiURI.resolve("/");
+  }
+
+  public CloseableHttpClient getHttpClient() {
+    return httpClient;
+  }
+
+  @CheckForNull
+  public synchronized List<ClouderaClusterDTO> getClusters() {
```
misolt, line 53: nit: Synchronizing on `this` has been controversial in dwh-tools, though I personally have nothing against it.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerHandle.java
```diff
@@ -0,0 +1,94 @@
...
+    this.clusters = ImmutableList.copyOf(clusters);
+  }
+
+  @Override
+  public void close() throws IOException {
+    if (httpClient != null) {
+      try {
+        httpClient.close();
+      } catch (IOException ignore) {
+
```
misolt, line 70: Could you explain what are the exceptions that are caught here in practive? Is it related to some faulty code outside our control?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerLoginHelper.java
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.http.client.entity.UrlEncodedFormEntity;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.client.methods.HttpPost;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.message.BasicNameValuePair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ClouderaManagerLoginHelper {
```
dominik-klimczak, line 34: 1. Javadoc for the class would make the purpose clearer.
2. It seems it's a stateless utility class, so it's worth to follow the pattern: make the class final, private constructor and static method.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerConnector.java
```diff
@@ -0,0 +1,117 @@
...
+      URI baseURI, CloseableHttpClient httpClient, String user, String password) throws Exception {
+    ClouderaManagerLoginHelper loginHelper = new ClouderaManagerLoginHelper();
+    loginHelper.login(baseURI, httpClient, user, password);
+  }
+
+  private HttpClientBuilder disableSSLVerification(HttpClientBuilder builder) throws Exception {
+    // Cloudera Manager API SSL certificate is not in list of know certificates.
+    // So, switch off SSL certificate validation.
+    // It is  expected that Dumper will work  in internal private network (probably localhost calls)
+    builder.setSSLContext(
```
dominik-klimczak, line 112: Can't it be solved in an environment by adding the certificate to the trusted ones? Maybe it's worth to add a parameter that controls this behavior (in a separate pr)?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaAPIHostsTask.java
```diff
@@ -0,0 +1,67 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+import javax.annotation.Nonnull;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.util.EntityUtils;
+
+/**
+ * The tasks dum data from the <a
```
sayuzbas-google, line 33: typo `dump`

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractTask.java
```diff
@@ -103,7 +103,6 @@ public AbstractTask<?> onlyIfAllFailed(@Nonnull Task<?>... tasks) {
    * @throws Exception If the task fails.
    */
   @CheckForNull
-  // @VisibleForTesting @ForOverride
```
misolt, line 106: Can't say about `VisibleForTesting`, but I can see `ForOverride` being meaningful in this context. IMO it would be best to get info from the original authors what purpose they meant for some of the annotations in dwh-tools.

Do you know of any cases where a subclass needs to call `doRun` directly?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaManagerTask.java
sayuzbas-google, line 26: Naming a bit confusing for this class. `ClouderaClustersTask` also extends it, but does not have `manager` suffix. Can it be renamed to `AbstractClouderaTask`?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaServicesTask.java
```diff
@@ -0,0 +1,63 @@
...
+    try (Writer writer = sink.asCharSink(StandardCharsets.UTF_8).openBufferedStream()) {
+      for (ClouderaClusterDTO cluster : clusters) {
+        String servicesPerCluster =
+            handle.getApiURI() + "/clusters/" + cluster.getName() + "/services";
+
+        try (CloseableHttpResponse services = httpClient.execute(new HttpGet(servicesPerCluster))) {
+          String json = EntityUtils.toString(services.getEntity());
+          writer.write(jsonToJsonl(json));
+          writer.write('\n');
+        }
```
dominik-klimczak, line 58: I'd recommend to write all errors to a separate file - similar to exceptions.txt produced by the dumper for jdbc connectors.
This way it will be easier to debug problems like not all services present in the dump. (That can be done in a separate PR)

dominik-klimczak: This applies to other tasks too.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaAPIHostsTask.java
vladislav-sidorovich: Thanks, fixed

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaManagerTask.java
vladislav-sidorovich: Well, you are absolutely right, the current naming with Cloudera is confusing. 
We have `cloudera` connector which works just with Hadoop. Fix this confusion is planned in the next PR to keep the scope of PRs smaller. 

The scope of the current task is to add connector which will work with Cloudera Manager API and will be irrelevant for any other Hadoop vendors. 
So, this abstract task is parent for each specific task which works with Cloudera Manager API, kind of grouping and some utility help. 

Current naming:
For example `ClouderaClustersTask` dump data related to Cloudera's `Hadoop clusters. It's done by Cloudera Manager API, but still it's it's Cloudera's Hadoop clusters, not Cloudera Manager clusters. 

I'm open to any ideas to fix the confusion and improve the naming, please share, I will consider the ideas and try to fix the issue in the next PR.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerLoginHelper.java
vladislav-sidorovich: 1. done
2. yes, initial I even though about private `login` method in `ClouderaManagerConnector`, but eventually decided to extract it into separate class for unit testing simplicity. Updated to utility pattern

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaAPIHostsTask.java
```diff
@@ -0,0 +1,67 @@
...
+      throw new MetadataDumperUsageException(
+          "Cloudera clusters must be initialized before hosts dumping.");
+    }
+
+    try (Writer writer = sink.asCharSink(StandardCharsets.UTF_8).openBufferedStream()) {
+      for (ClouderaClusterDTO cluster : clusters) {
+        String hostPerClusterUrl = handle.getApiURI() + "/clusters/" + cluster.getName() + "/hosts";
+
+        try (CloseableHttpResponse hosts = httpClient.execute(new HttpGet(hostPerClusterUrl))) {
+          String json = EntityUtils.toString(hosts.getEntity());
```
rocgranada, line 59: Have you considered an internal interface / dto to sit between the Cloudera Manager API json response and our dumper output? I guess we should be safe from breaking changes between having a pinned version and the API Client not having updates for 7 years. But it has been common for most dumps to have one: https://github.com/google/dwh-migration-tools/blob/8fc9238b3920ecf8a0d08cfb9e89c43add07f061/dumper/lib-dumper-spi/src/main/java/com/google/edwmigration/dumper/plugin/lib/dumper/spi/BigQueryMetadataDumpFormat.java#L58 

As another example, Hive's output for tables has currently two formats:

- Dumper's internally defined format (https://github.com/google/dwh-migration-tools/blob/8fc9238b3920ecf8a0d08cfb9e89c43add07f061/dumper/lib-dumper-spi/src/main/java/com/google/edwmigration/dumper/plugin/lib/dumper/spi/HiveMetadataDumpFormat.java#L60)
- RAW response from the Thrift call to Hive (https://github.com/google/dwh-migration-tools/blob/8fc9238b3920ecf8a0d08cfb9e89c43add07f061/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hive/HiveMetadataConnector.java#L342)

I don't have a particuarly strong opinion on the topic: extra boilerplate vs more control / readability. But I thought it would be good to consider.

rocgranada: This comment applies to all Tasks that output a file.

dumper/app/build.gradle
```diff
@@ -154,21 +166,22 @@ task cloudExtractorStartScripts(type: CreateStartScripts) {
...
-        new com.github.jk1.license.filter.LicenseBundleNormalizer(bundlePath: rootProject.file("gradle/license-bundle-normalizer.json"), createDefaultTransformationRules: true)
-    ]
-    renderers = [
-        new com.google.edwmigration.dumper.build.licensereport.CsvReportRenderer(),
-        new com.github.jk1.license.render.JsonReportRenderer('index.json', false),
-        new com.github.jk1.license.render.InventoryHtmlReportRenderer("index.html", "Licenses of Third Party Dependencies")
-    ]
-    allowedLicensesFile = rootProject.file("gradle/license-allowed.json")
-}
+//todo temporary disable license check. Clouder Case Number:  1077072
```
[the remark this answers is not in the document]
vladislav-sidorovich: Correct, but I prefer to revert all the changes in this file once license question will be resolved. 
I will keep this conversation `unresolved` till this moment.

dumper/app/build.gradle
```diff
@@ -154,21 +166,22 @@ task cloudExtractorStartScripts(type: CreateStartScripts) {
...
+//        new com.google.edwmigration.dumper.build.licensereport.CsvReportRenderer(),
+//        new com.github.jk1.license.render.JsonReportRenderer('index.json', false),
+//        new com.github.jk1.license.render.InventoryHtmlReportRenderer("index.html", "Licenses of Third Party Dependencies")
+//    ]
+//    allowedLicensesFile = rootProject.file("gradle/license-allowed.json")
+//}
 
 tasks.named('check') {
-    dependsOn 'checkLicense'
+//    dependsOn 'checkLicense'
```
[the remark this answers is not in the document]
vladislav-sidorovich: There are no plans to keep these comments/changes around longer than the current PR exists. I'll use the conversation with @dominik-klimczak above as a reminder to revert changes before merging.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerConnector.java
vladislav-sidorovich: Yes, it's possible. But if we will take into account our use case it will mean that our customers will need to do an additional not trivial step to use our tool. At the same time I don't see a lot of value on this because the tool should be used in their private datacenter/network. 

Let's keep it as is for usability simplification.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerHandle.java
vladislav-sidorovich: I agree, the field is updated to provide better readability for the code.

dumper/app/build.gradle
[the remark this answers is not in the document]
vladislav-sidorovich: Cloudera's support team doesn't have required information.  The library is removed from deps for safety.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerHandle.java
vladislav-sidorovich: In practice I do not expect any exceptions here, however the interface declare that it _might_ be an exception and the idea here is to keep plain program execution flow running because we have nothing to do with the exception.

In general, this pattern is similar to try/with-resource implementation The spec: https://docs.oracle.com/javase/specs/jls/se16/html/jls-14.html#jls-14.20.3

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractTask.java
vladislav-sidorovich: I removed useless comment from the code. Do you suggest to improve  the code and add `@ForOverride` ?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaAPIHostsTask.java
vladislav-sidorovich: I did. In general, the idea of the Dumper is to be a client and to be as simple as possible. All the logic to support different versions of API or fields value should be done on server. It's why I'm trying to avoid strict schemas here and write the responses as is.

There are exceptions:
1. I want to be sure that response is valid json/jsonl format. 
2. I introduces dto `com.google.edwmigration.dumper.application.dumper.connector.cloudera.manager.dto.ApiClusterListDto` for `ClouderaClustersTask`  because this task may do requests to 2 different APIs and  to make sure that success output will be always in single structure `dto` is suitable.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractTask.java
misolt: I know this is a tiny change, but the class is an API used by the entire `dumper.app` including parts not used by us and the annotation determines what is the correct usage of this API.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaManagerTask.java
vladislav-sidorovich: In this class there is `final` method which should not be overridden and the `abstract` which must be overridden.  Additionally, there is a helper method that can be overridden if needed in a subclass (at least I don't expect that right now).  I think it's what you expect. 

Do you expect something different?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerHandle.java
vladislav-sidorovich: I don't see a big issues with it. So, let's keep it as is for now. It will be simple to change in case of any future needs/requests.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/AbstractTask.java
vladislav-sidorovich: It was commented for 2 years. 
There is a good practice to clean-up the code when you work with it in the scope for your tasks, otherwise the cost of support will be only growing.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaManagerTask.java
vladislav-sidorovich: This `abstract` class marked as package-private. 
If you will find any documents with the motivation to make subclasses private please share and I will just the code

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerHandle.java
vladislav-sidorovich: The comment with the explanation added.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaServicesTask.java
vladislav-sidorovich: As discussed with @dominik-klimczak it should work out of the box.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaClustersTask.java
```diff
@@ -0,0 +1,124 @@
...
+      try (CloseableHttpResponse clustersResponse =
+          httpClient.execute(new HttpGet(handle.getApiURI() + "/clusters"))) {
+        String clustersJson = EntityUtils.toString(clustersResponse.getEntity());
+
+        clusterList = objectMapper.readValue(clustersJson, ApiClusterListDto.class);
+      }
+    }
+
+    try (Writer writer = sink.asCharSink(StandardCharsets.UTF_8).openBufferedStream()) {
+      writer.write(objectMapper.writeValueAsString(clusterList));
```
misolt, line 86: This seems to only write `ApiClusterListDto@abc123` rather than the data in this dto.

My suggestion is to log the items directly without wrapping.

If you will change this behavior, please also include it in unit tests.

vladislav-sidorovich: No, it's write json with the structure of the class, such behavior is the motivation to introduce `ApiClusterListDto`. 

It's also covered by unit test `com.google.edwmigration.dumper.application.dumper.connector.cloudera.manager.ClouderaClustersTaskTest:142`

misolt: True, I forgot that jackson can convert this to string by itself with reflection. I've tested this, the output looks like: `{"items": [ obj#1, obj#2, ... ]}` where `obj#n` are items as objects

Still, I'd suggest to simplify this format and write just the list without the single-entry dictionary - `[ obj#1, obj#2, ... ]`. To me it looks like we won't need the wrapping ListDto.

For the tests, I meant comparing this to a fixed expected value. The test you've mentioned checks if we're able convert both ways without errors, but adding a value assertion would be helpful to see what we're converting to.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaAPIHostsTask.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.util.EntityUtils;
+
+/**
+ * The tasks dump data from the <a
+ * href="https://archive.cloudera.com/cm7/7.11.3.0/generic/jar/cm_api/apidocs/json_ApiHostList.html">Hosts
+ * API</a> which doesn't contain usage and disk data and collected as a fallback.
+ */
+public class ClouderaAPIHostsTask extends AbstractClouderaManagerTask {
```
misolt, line 37: For this and other subclasses of `AbstractClouderaManagerTask`, inheritance needs to be addressed. The options are:
- design and document to support future inheritance
- prevent inheritance
(Effective Java 3rd ed, Item 19)

Both options are valid. Personally I'd use the second, because it seems faster to implement - all that is needed would be making the classes `final` and minor visibility changes (removing `protected`, because all subclasses are in the same package anyway).

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerHandle.java
```diff
@@ -0,0 +1,96 @@
...
+      } catch (IOException ignore) {
+        // The intention is to do graceful shutdown and try to release the resource.
+        // In case of errors we do not need to interrupt the execution flow
+        // because the e2e use case might be successful
+      }
+    }
+  }
+
+  @AutoValue
+  public abstract static class ClouderaClusterDTO {
```
misolt, line 85: I recommend dropping the `-DTO` suffix for this class.
The class is not used to transfer data "over the wire" and the project doesn't use MVC or a similar architectural pattern.

If you're concerned that this class doesn't contain a full description of a cluster, a suffix such as `-Summary` or `-Info` can be added optionally.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaClustersTaskTest.java
```diff
@@ -0,0 +1,271 @@
...
+          .write(
+              (String)
+                  argThat(
+                      content -> {
+                        try {
+                          ApiClusterListDto listDto =
+                              objectMapper.readValue((String) content, ApiClusterListDto.class);
+                          assertNotNull(listDto.getItems());
+                        } catch (JsonProcessingException e) {
+                          throw new RuntimeException(e);
```
misolt, line 145: This test makes some assertions, but it also makes a check without assertions, following the steps:
- Make a conversion from format A to B
- Convert the results back from B to A
- Check that no exceptions were thrown

This kind of check can detect some errors, but doesn't verify or document much about what format B actually looks like. It would be very helpful to have a more typical test:
- Take an input in format A
- Make a one-way conversion
- Compare the result with expected output in format B

Rewriting existing tests is not required, but I'd ask you to add a test of the second type for this logic.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerHandle.java
misolt: Let's go back to the initial comment: I'd suggest to either fail or return an empty list if `clusters` wasn't initialized. This is a safer alternative to returning null.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaManagerTask.java
```diff
@@ -0,0 +1,47 @@
...
+
+  protected abstract Void doRun(
+      TaskRunContext context, @Nonnull ByteSink sink, @Nonnull ClouderaManagerHandle handle)
+      throws Exception;
+
+  protected final String jsonToJsonl(String json) {
+    json = json.replaceAll("\n", "");
+    json = json.replaceAll("\r", "");
+    return json;
+  }
```
shevek-google, line 46: This, and its usages, feel a little bit fragile. Do we think it's worth doing an unchecked Jackson parse into a tree, and then a re-stringify with the no-pretty-print flag set?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaClustersTask.java
```diff
@@ -0,0 +1,124 @@
...
+        LOG.warn(
+            "Can't receive cluster [{}] status by url [{}]. Response status is [{}] and body: {}",
+            clusterName,
+            requestUrl,
+            clusterStatus.getStatusLine().getStatusCode(),
+            responseBody);
+
+        return null;
+      } else {
+        JsonNode jsonNode = objectMapper.readTree(responseBody);
```
shevek-google, line 118: Should we have parsed this from the entity's InputStream, and only used EntityUtils.toString() in the error-path?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerConnector.java
```diff
@@ -0,0 +1,128 @@
...
+    String password = arguments.getPasswordOrPrompt();
+    doClouderaManagerLogin(handle.getBaseURI(), httpClient, user, password);
+
+    return handle;
+  }
+
+  private void doClouderaManagerLogin(
+      URI baseURI, CloseableHttpClient httpClient, String user, String password) throws Exception {
+    ClouderaManagerLoginHelper.login(baseURI, httpClient, user, password);
+  }
```
shevek-google, line 117: Helper method called from only one location - should it inline?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerLoginHelper.java
```diff
@@ -0,0 +1,75 @@
...
+      // verify success login by home page access
+      try (CloseableHttpResponse home = httpClient.execute(new HttpGet(baseURI + "/cmf/home"))) {
+        if (HttpStatus.SC_OK != home.getStatusLine().getStatusCode()) {
+          LOG.error(
+              "Login to Cloudera Manager wasn't successful. "
+                  + "The response code for /cmf/home page is [{}] and response: {}",
+              home.getStatusLine().getStatusCode(),
+              home);
+
+          throw new MetadataDumperUsageException("Login wasn't successful: " + login);
```
shevek-google, line 69: Does this `+ login` print something useful, in case it failed?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/dto/ApiClusterDto.java
```diff
@@ -0,0 +1,134 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+
+/**
+ * DTO class for the official Cloudera API <a
+ * href="https://archive.cloudera.com/cm7/7.11.3.0/generic/jar/cm_api/apidocs/json_ApiCluster.html">json</a>
+ *
+ * <p>Note: The license for <a
+ * href="https://mvnrepository.com/artifact/com.cloudera.api.swagger/cloudera-manager-api-swagger/7.11.0">generated</a>
+ * code is unclear, the own model for public schema used instead of it.
```
shevek-google, line 29: I don't understand this comment.

Also, is this class generated?

Since we use Jackson's ObjectMapper to deserialize this, it might be worth using AutoValue and Builder, which I think Jackson can honour. It will greatly reduce the amount of "boring" code here: https://stackoverflow.com/questions/43364266/jackson-autovalue-builder-bean-validation

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaAPIHostsTaskTest.java
```diff
@@ -0,0 +1,178 @@
...
+              argThat(
+                  get -> get != null && get.getURI().toString().endsWith("/second-cluster/hosts"))))
+          .thenReturn(responseCldClst);
+
+      mockedUtils
+          .when(() -> EntityUtils.toString(entityId1))
+          .thenReturn("{'clusterName':'first-cluster'}");
+      mockedUtils
+          .when(() -> EntityUtils.toString(entityCldClst))
+          .thenReturn("{\n'clusterName':'second-cluster'\n\r}");
```
shevek-google, line 114: Observation: This is fine, but as a general note, I'm always a bit cautious using mocks, because it's never quite clear what one is testing. Out of interest, I did find https://github.com/square/okhttp/tree/master/mockwebserver but I've no experience with it. My previous experience is mostly with Spring's MockMvc, which is brilliant, and doesn't use the network stack. So this observation might be interesting, but might not be useful.

shevek-google (commented):
This looks solid. I think if you were to look into any of these items, it would be, in priority order:
1. The jsonl thing, and other serialization notes: Better to use a proper serializer.
2. The AutoValue possibility (optional)
Everything else is probably ignorable unless you like it.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerConnector.java
vladislav-sidorovich: Initially it was private method in this class, then I decided to extract it into separate class for testing simplifications.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerHandle.java
vladislav-sidorovich: Could you explain what is the problem with returning `null`?  It's straightforward logic to validate and it's quite readable as well. 

Fail in this class is too much responsibilities for this class, the decision to fail should be done in particular task depends on the task logic. 
Empty list can be a good solution as null-object pattern, but in current case I need just simple check.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaAPIHostsTask.java
vladislav-sidorovich: I'm not sure that this advice somehow improves the current design in the Dumper. Moreover this practice is not adopted, so this connector will look different compare to others. 
If you think that this point is essential, let's discuss it with the entire team and create a separate ticket to apply the practice if we will decide to adopt it.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaManagerTask.java
vladislav-sidorovich: I understand that this code might looks weird when you see it first time. However it works according to specification https://jsonlines.org/.

Using Jackson for this purpose might be more costly in terms of footprint, but as I understand the Dumper use case the additional footprint of Jackson usage will not be visible in real usage.  So, I agree, let's use Jackson to keep the code more natural for developers. Also we have unit-test for `jsonl` so it's quite safe change.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerLoginHelper.java
vladislav-sidorovich: Yes, it prints response with some useful information (like redirects and etc).

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/dto/ApiClusterDto.java
vladislav-sidorovich: The classes are created according to the schema, not generated by a tool. I will play with autovalue+Jackson in background, if it will reduce boilerplate I will adopt it here.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaClustersTask.java
vladislav-sidorovich: Well, good point.  It might be combined with mockserver in tests and should looks good. 
I added it to the scope http://b/377651080

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaClustersTaskTest.java
vladislav-sidorovich: Could you please clarify why this test: `doesn't verify or document much about what format B actually looks like` ?

This test asserts:
1. The string in file is valid `json`
2. The `json` can be parsed into structure `ApiClusterListDto` with required fields (only one to be honest).

The order of fields in `json` or formatting with tabs or spaces is not important it's why it's not asserted here.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaManagerTask.java
vladislav-sidorovich: I think we can do parsing of string into `JsonNode` first and then write it back as `json`. Is it what you mean?

vladislav-sidorovich: done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaClustersTask.java
vladislav-sidorovich: done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerLoginHelper.java
vladislav-sidorovich: The example of `.toString()`
```
HttpResponseProxy{HTTP/1.1 401 Unauthorized [Date: Fri, 08 Nov 2024 09:40:12 GMT, WWW-Authenticate: Negotiate, Content-Language: en-US, Content-Type: text/html;charset=utf-8, Cache-Control: max-age=0, no-cache, no-store, Pragma: no-cache, Expires: Tue, 01 Jan 1980 1:00:00 GMT, X-UA-Compatible: IE=edge, Strict-Transport-Security: max-age=31536000 ; includeSubDomains, X-XSS-Protection: 1; mode=block, X-Frame-Options: SAMEORIGIN, X-Content-Type-Options: nosniff, Content-Length: 546] ResponseEntityProxy{[Content-Type: text/html;charset=utf-8,Content-Length: 546,Chunked: false]}}
```

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -156,9 +158,15 @@ public class ConnectorArguments extends DefaultArguments {
           .describedAs("com.company.Driver");
   private final OptionSpec<String> optionUri =
       parser
-          .accepts(OPT_URI, "JDBC driver URI")
+          .accepts(
+              OPT_URI,
+              "The main Connector's URI to dump metadata. "
+                  + "It can JDBC driver URI or HTTP endpoint URL, must be specified "
```
rocgranada, line 164: nit: "It can be a JDBC ..."

rocgranada (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerHandle.java
vladislav-sidorovich: Resolved. It works as expected if we will find more use cases we will change it.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaClustersTaskTest.java
vladislav-sidorovich: Resolved. Tests are valid, however I agree that readability is not the best one. If any idea for improvement will come, let's consider them in the scope of http://b/377651080

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerHandle.java
vladislav-sidorovich: Resolved. Let's leave it as is, it's very limited scope of the usage.

PR #634, opened 2024-11-04
Remove another connector's configuration from snowflake-lite

misolt:
The TCO lite connector was based on the metadata connector. Some parts of the old connector are redundant for the new use case and can now be removed.

- Remove redundant inheritance from the metadata dump format
- Simplify a helper method because the parameter list was getting too big (8 arguments)

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeLiteConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeLiteConnector.java
@@ -77,8 +77,7 @@ public final class SnowflakeLiteConnector extends AbstractSnowflakeConnector
     out.add(new DumpMetadataTask(arguments, FORMAT_NAME));
     out.add(new FormatTask(FORMAT_NAME));
 
-    boolean INJECT_IS_FAULT = arguments.isTestFlag('A');
-    final String IS = INJECT_IS_FAULT ? "__NONEXISTENT__" : "INFORMATION_SCHEMA";
+    final String IS = "INFORMATION_SCHEMA";
     final String AU = "SNOWFLAKE.ACCOUNT_USAGE";
     final String AU_WHERE = " WHERE DELETED IS NULL";
 
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeLiteConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeLiteConnector.java
@@ -101,8 +101,9 @@ public final class SnowflakeLiteConnector extends AbstractSnowflakeConnector
 
     if (arguments.isAssessment()) {
       for (AssessmentQuery item : planner.generateAssessmentQueries()) {
-        String AU = "SNOWFLAKE.ACCOUNT_USAGE";
-        String query = String.format(item.formatString, AU, /* an empty WHERE clause */ "");
+        String usageSchema = "SNOWFLAKE.ACCOUNT_USAGE";
+        String query =
+            String.format(item.formatString, usageSchema, /* an empty WHERE clause */ "");
         String zipName = item.zipEntryName;
         Task<?> task = new JdbcSelectTask(zipName, query).withHeaderTransformer(item.transformer());
         out.add(task);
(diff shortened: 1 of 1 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeLiteConnector.java
```diff
@@ -50,72 +55,54 @@ public SnowflakeLiteConnector() {
...
-      @Nonnull String format,
-      @Nonnull String schemaZip,
-      @Nonnull String alteredSchemaView,
-      @Nonnull String usageZip,
-      @Nonnull String usageView,
-      @Nonnull String usageFilter) {
+  private ImmutableList<Task<?>> sqlTasksWithInfoSchemaFallback(
+      Class<? extends Enum<?>> header, String format, String schemaZip, String usageZip) {
+    String usageView = "SNOWFLAKE.ACCOUNT_USAGE";
+    String usageFilter = " WHERE DELETED IS NULL";
```
sayuzbas-google, line 65: nit: these 3 Strings could be marked as `final`

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeLiteConnector.java
```diff
@@ -50,72 +55,54 @@ public SnowflakeLiteConnector() {
...
+            TablesFormat.Header.class,
+            "SELECT table_catalog, table_schema, table_name, table_type, row_count, bytes,"
+                + " clustering_key FROM %1$s.TABLES%2$s",
+            TablesFormat.IS_ZIP_ENTRY_NAME,
+            TablesFormat.AU_ZIP_ENTRY_NAME));
 
     if (arguments.isAssessment()) {
       for (AssessmentQuery item : planner.generateAssessmentQueries()) {
-        String query = String.format(item.formatString, AU, /* an empty WHERE clause */ "");
+        String usageSchema = "SNOWFLAKE.ACCOUNT_USAGE";
```
sayuzbas-google, line 103: this string is the same as in line 64, so could be moved as field as well. Also it creates the same string in a loop, so worth to moved out of the loop.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeLiteConnector.java
```diff
@@ -30,18 +31,22 @@
...
+import javax.annotation.Nullable;
+import javax.annotation.ParametersAreNonnullByDefault;
 
 @AutoService(Connector.class)
-public final class SnowflakeLiteConnector extends AbstractSnowflakeConnector
-    implements SnowflakeMetadataDumpFormat {
+@ParametersAreNonnullByDefault
+public final class SnowflakeLiteConnector extends AbstractSnowflakeConnector {
 
   private static final String NAME = "snowflake-lite";
```
sayuzbas-google, line 47: nit: `CONNECTOR_NAME`

sayuzbas-google (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeLiteConnector.java
misolt: The repetition is bigger than the two occurrences, because it's also in the metadata connector. I'm planning to remove all of these variables, but it's a slightly bigger change so I'd leave it to the next PR. In this PR I can bring back the parameter, which would remove one repeated occurrence. WDYT?

The string literal is a constant, so an object is not created, this only sets a reference to the constant. Using the minimal necessary scope is an advantage here.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeLiteConnector.java
misolt: This is intended to match the `name` super-constructor parameter for the `getName` method in `AbstractConnector` and another parameter in `ArchiveNameUtil` which is also called `name`. It's not used in anywhere else, so the current version is more consistent.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeLiteConnector.java
misolt: For final attributes I agree. For final variables I think we should decide to either use them whenever possible or to never do it. Using final variables from time to time decreases readability - it's hard to tell if a missing `final` means the variable will be modified or the keyword was simply skipped in this case.

Current state of the Dumper seems closer to never using final variables than to always using them, so I'm following this trend.

PR #635, opened 2024-11-07
Enable Kerberos keytab auth for HdfsConnector

sten1ee:
Dumper/HdfsConnector now takes two more (optional) args, that should always go together:
--kerberos-principal _service-principal_
--kerberos-keytab-path _path-to-keytab-file_

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
@@ -389,7 +389,7 @@ public class ConnectorArguments extends DefaultArguments {
               OPT_KERBEROS_PRINCIPAL,
               "Set Kerberos principal. It is usually a Service principal of the form: "
                   + "DUMPER/webserver.example.com@EXAMPLE.COM . (No default value)")
-          .withOptionalArg()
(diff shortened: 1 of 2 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -381,6 +382,24 @@ public class ConnectorArguments extends DefaultArguments {
...
           .defaultsTo(OPT_THREAD_POOL_SIZE_DEFAULT);
 
+  // Kerberos principal
+  private final OptionSpec<String> optionKerberosPrincipal =
+      parser
+          .accepts(
+              OPT_KERBEROS_PRINCIPAL,
+              "Set Kerberos principal. It is usually a Service principal of the form: "
+                  + "DUMPER/webserver.example.com@EXAMPLE.COM . (No default value)")
+          .withOptionalArg()
```
aleofreddi, line 392: If there's no default, shouldn't this be `withRequiredArgs` as the keytab option below?

sten1ee: Right!
Fixed
Thanks!

PR #666, opened 2024-12-03
upgraded to apache avro to 1.11.4

kaxuna:
upgraded and forced to apache avro to 1.11.4

### DIFF
--- a/dumper/app/build.gradle
+++ b/dumper/app/build.gradle
@@ -41,6 +41,7 @@ dependencies {
 
     annotationProcessor libs.auto.value
 
+//    implementation("org.apache.avro:avro:1.11.4")
     implementation project(':dumper:lib-common')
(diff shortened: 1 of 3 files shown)


dumper/app/build.gradle
```diff
@@ -41,6 +41,7 @@ dependencies {
 
     annotationProcessor libs.auto.value
 
+    implementation("org.apache.avro:avro:1.11.4")
```
vladislav-sidorovich, line 44: You already defined the version in [libs.versions.toml](https://github.com/google/dwh-migration-tools/pull/666/files#diff-697f70cdd88ba88fe77eebda60c7e143f6ad1286bca75017421e93ad84fb87df), so you can use reference here.

kaxuna: will delete this line. duplication. line 54 does the same thing

PR #671, opened 2024-12-06
Adding 2 new columns for SegmentStatistics table

sayuzbas-google:
Add 2 new columns for SegmentStatistics table.
They will be used to calculate correct space usage by available databases


misolt (approved):

dominik-klimczak (approved):

kpiotrowski (approved):

vladislav-sidorovich:
i'm wondering, if we will remove some columns from the queries, no tests will fail.  
Why we do not use `*` in the queries in such case?

PR #679, opened 2024-12-30
[b/382467945] Move parts of JdbcHandle logic to a util class

misolt:
JdbcHandle is a class which provides common Jdbc logic to connectors that choose to use it as their handle.

Redshift extraction needs (and currently uses) the general-purpose logic from JdbcHandle, but it also needs to define its own handle, for aws-specific connection setup (at the moment this setup is done in Tasks).

Make the JdbcHandle methods available as a util class so that a custom handle planned for Redshift will be able to use them.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/handle/HandleUtil.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/handle/HandleUtil.java
@@ -51,7 +51,7 @@ public class HandleUtil {
     }
   }
 
-  public static HikariConfig createConfig(DataSource dataSource, int threadPoolSize) {
+  public static HikariConfig createHikariConfig(DataSource dataSource, int threadPoolSize) {
     HikariConfig config = new HikariConfig();
     // providing "0" sets it to Integer.MAX_VALUE
     config.setConnectionTimeout(0);
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/handle/HandleUtil.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/handle/HandleUtil.java
@@ -30,7 +30,7 @@ public class HandleUtil {
 
   private static final Logger LOG = LoggerFactory.getLogger(HandleUtil.class);
 
-  public static JdbcTemplate templateFromDataSource(DataSource dataSource) {
+  public static JdbcTemplate createJdbcTemplate(DataSource dataSource) {
     JdbcTemplate template = new JdbcTemplate(dataSource);
     template.setFetchSize(1024);
     return template;
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/handle/HandleUtil.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/handle/HandleUtil.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2022-2024 Google LLC
+ * Copyright 2013-2021 CompilerWorks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/handle/HandleUtil.java
```diff
@@ -0,0 +1,73 @@
...
+import java.sql.Connection;
+import java.sql.SQLException;
+import javax.annotation.ParametersAreNonnullByDefault;
+import javax.sql.DataSource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.jdbc.core.JdbcTemplate;
+
+@ParametersAreNonnullByDefault
+public class HandleUtil {
```
sayuzbas-google, line 30: nit: `JdbcHandleUtil`

sayuzbas-google (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/handle/HandleUtil.java
misolt: I'd like to keep all handle-related utils in a single class for now, it hasn't gotten very big yet.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/handle/HandleUtilTest.java
```diff
@@ -0,0 +1,72 @@
...
+@RunWith(MockitoJUnitRunner.class)
+public class HandleUtilTest {
+
+  @Test
+  public void templateFromDataSource_success_fetchSizeMatches() {
+    DataSource dataSource = mock(DataSource.class);
+
+    JdbcTemplate template = templateFromDataSource(dataSource);
+
+    assertEquals(1024, template.getFetchSize());
```
sayuzbas-google, line 44: it is not so obvious, why `fetchSize` should be equal to `1024`. And it is set to 1024 inside code, not even sure, if it is worth to have test for that.

misolt: This was the value before. Unfortunately it's not documented why 1024 was chosen.

The intent of this test is to verify that the developer didn't forget to set the custom fetch size e.g. by creating a `JdbcTemplate` directly, without going through the util method.

This verification could be done with a named value or using mocks, but those options would complicate the test code.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/handle/HandleUtil.java
```diff
@@ -0,0 +1,67 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(HandleUtil.class);
+
+  public static JdbcTemplate templateFromDataSource(DataSource dataSource) {
```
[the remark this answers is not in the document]
misolt: Done, but I also removed the `FromDataSource`, because the other method that created a template is now removed.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/handle/HandleUtil.java
```diff
@@ -0,0 +1,67 @@
...
+        // This used to be thrown by Preconditions and was kept for compatibility.
+        // TODO: Replace with a better error.
+        throw new NullPointerException();
+      } else {
+        LOG.debug("Connection test succeeded");
+      }
+    }
+  }
+
+  public static HikariConfig createConfig(DataSource dataSource, int threadPoolSize) {
```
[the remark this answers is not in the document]
misolt: Done

PR #680, opened 2024-12-30
Remove DummyHandle, make AbstractHandle.close final

misolt:
### Make AbstractHandle.close() final
* The interface `Handle` extends `AutoCloseable`, which means subclasses need to be closed.
* `Handle` is implemented by `AbstractHandle`, where `close` is added as an empty method, so subclasses of  `AbstractHandle` don't need to be closed, despite implementing `Handle`.
* In `JdbcHandle` this changes once more: `JdbcHandle` and any potential subclasses do need to be closed (additionally, the empty `AbstractHandle.close()` is called when closing a `JdbcHandle`).

Remove the 3rd point by bypassing `AbstractHandle`. Prevent this pattern in the future by making the empty `close()` in `AbstractHandle` final. If a `Handle` needs to be closed, it has no need to extend `AbstractHandle`, because `AbstractHandle` doesn't provide anything other than the empty `close()` method.

### Remove DummyHandle

The class is used only once, to create an empty handle in a test, which can be replaced with a lambda.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/handle/AbstractHandle.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/handle/AbstractHandle.java
@@ -22,5 +22,5 @@ import java.io.IOException;
 public abstract class AbstractHandle implements Handle {
 
   @Override
-  public void close() throws IOException {}
+  public final void close() throws IOException {}
 }
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/handle/AbstractHandle.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/handle/AbstractHandle.java
@@ -22,5 +22,8 @@ import java.io.IOException;
 public abstract class AbstractHandle implements Handle {
 
   @Override
-  public final void close() throws IOException {}
+  public void close() throws IOException {
+    // Intentionally empty to simplify implementation of some Handles.
+    // If your subclass needs to release resources, consider implementing Handle directly.
+  }
 }
--- a/dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/test/TestConnector.java
+++ b/dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/test/TestConnector.java
@@ -22,7 +22,6 @@ import com.google.edwmigration.dumper.application.dumper.connector.AbstractConne
 import com.google.edwmigration.dumper.application.dumper.connector.Connector;
 import com.google.edwmigration.dumper.application.dumper.connector.ConnectorProperty;
 import com.google.edwmigration.dumper.application.dumper.connector.MetadataConnector;
-import com.google.edwmigration.dumper.application.dumper.handle.AbstractHandle;
 import com.google.edwmigration.dumper.application.dumper.handle.Handle;
 import com.google.edwmigration.dumper.application.dumper.task.Task;
 import java.util.List;
@@ -30,7 +29,7 @@ import javax.annotation.Nonnull;
 
 @AutoService({Connector.class, MetadataConnector.class})
 public class TestConnector extends AbstractConnector implements MetadataConnector {
-  private static final Handle DUMMY_HANDLE = new AbstractHandle() {};
+  private static final Handle DUMMY_HANDLE = () -> {};
 
   public TestConnector() {
     super("test");
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/handle/JdbcHandle.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/handle/JdbcHandle.java
@@ -31,7 +31,7 @@ import org.springframework.jdbc.core.JdbcTemplate;
  *
  * @author shevek
  */
-public class JdbcHandle extends AbstractHandle {
+public class JdbcHandle implements Handle {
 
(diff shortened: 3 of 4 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/handle/AbstractHandle.java
```diff
@@ -22,5 +22,5 @@
 public abstract class AbstractHandle implements Handle {
 
   @Override
-  public void close() throws IOException {}
+  public final void close() throws IOException {}
```
vladislav-sidorovich, line 25: It doesn't look like correct direction for `AbstractHandle`. This change means that any subclass of `AbstractHandle` will not be able to close any resource properly in `close` method. 

We can:
1. leave empty `close()` method.  - In this case the motivation of abstract classes is to override `close` only if needed
2. remove `AbstractHandle` completely. - In this case subclasses will explicitly override `close()`. 
3. add something like `default void close() throws IOException {}` to `Handle`. It's not very popular practice because `default` was introduced for backward compatibility with older java versions and such usage is something like `hack`, but it's still valid.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/test/TestConnector.java
```diff
@@ -22,15 +22,14 @@
...
-import com.google.edwmigration.dumper.application.dumper.handle.AbstractHandle;
 import com.google.edwmigration.dumper.application.dumper.handle.Handle;
 import com.google.edwmigration.dumper.application.dumper.task.Task;
 import java.util.List;
 import javax.annotation.Nonnull;
 
 @AutoService({Connector.class, MetadataConnector.class})
 public class TestConnector extends AbstractConnector implements MetadataConnector {
-  private static final Handle DUMMY_HANDLE = new AbstractHandle() {};
+  private static final Handle DUMMY_HANDLE = () -> {};
```
vladislav-sidorovich, line 32: This change looks good, I like it.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/handle/AbstractHandle.java
shevek-google: Agree with Vladislav - this shouldn't be final, it would defeat the point.

I would keep AbstractHandle - I _always_ have an interface and an abstract class, even if it's empty, for any subclass hierarchy. One always needs it sooner or later.

Side note: I don't think of default as a hack, it's a great way of writing mixin functionality based on a small set of underlying state methods.

shevek-google (changes requested):
I'm not sure we should do this change overall; I would both keep DummyHandle and AbstractHandle as-is. While, technically, they can be optimized out, this looks like premature optimization, and may not be in the interests of maximum future maintainability.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/handle/AbstractHandle.java
misolt: I also wouldn't remove AbstractHandle. It would be a large refactor and I don't think it would improve the code.

At first my understanding of this hierarchy was that:
1. `Handle` represents resources which _may_ need to be closed
2. We've decided to always make the closing step (because a redundant `close` is better than a missing one)
3. For some `Handle`s the closing step will be trivial, for which we use `AbstractHandle`

I see now that 3. is not true, because the actual reason for AbstractHandle is to follow a more general convention about interfaces.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/test/TestConnector.java
misolt: This is not affected by changes in `AbstractHandle`, it would have to be a change in `Handle`.

This is also a test fake, only for cases where we need to pass something that resembles a handle for the code to run. If the value of the handle was relevant, we should use a real handle instead of a dummy class or `() -> {}`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/handle/AbstractHandle.java
vladislav-sidorovich: It seems that during this conversation we come to the conclusion that we should leave the `AbstractHandle` without any changes. 
I'm totally ok with it.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/handle/AbstractHandle.java
```diff
@@ -22,5 +22,8 @@
 public abstract class AbstractHandle implements Handle {
 
   @Override
-  public void close() throws IOException {}
+  public void close() throws IOException {
+    // Intentionally empty to simplify implementation of some Handles.
+    // If your subclass needs to release resources, consider implementing Handle directly.
```
vladislav-sidorovich, line 27: I'm not sure we need any comment here. The `close()` comes from `Closeable` and it's already well documented. 

```
Closes this stream and releases any system resources associated with it. If the stream is already closed then invoking this method has no effect.
As noted in AutoCloseable.close(), cases where the close may fail require careful attention. It is strongly advised to relinquish the underlying resources and to internally mark the Closeable as closed, prior to throwing the IOException.
```
As an option we can use `{@inheritDoc}`,

PR #681, opened 2025-01-02
[b/387249832] Move Redshift URL utilities to a Util class

misolt:
### Extract URL logic:
- Postgres URL builder
- Simple RS URL builder
- Redshift IAM URL builder
- default port constant

### Fixes in Connector:
- move commented-out code to a method
- remove redundant comments
- remove an unused method
- remove redundant `SuppressWarnings`

### Purpose of change

The moved code represented an additional responsibility of the Redshift connector besides managing the handle in `open` and building the task list in `addTasksTo`. Simplify the class to make working on its main features easier.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftUrlUtil.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftUrlUtil.java
@@ -34,7 +34,7 @@ class RedshiftUrlUtil {
   static String makeJdbcUrlPostgresql(ConnectorArguments arguments)
       throws MetadataDumperUsageException, UnsupportedEncodingException {
     String password = arguments.getPasswordIfFlagProvided().orElse(null);
-    return "jdbc:postgresql://"
+    return makeScheme("postgresql")
         + arguments.getHostOrDefault()
         + ":"
         + arguments.getPort(DEFAULT_PORT)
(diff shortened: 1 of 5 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftUrlUtil.java
```diff
@@ -0,0 +1,103 @@
...
+import java.io.UnsupportedEncodingException;
+import javax.annotation.Nonnull;
+import javax.annotation.ParametersAreNonnullByDefault;
+
+@ParametersAreNonnullByDefault
+class RedshiftUrlUtil {
+
+  private static final int DEFAULT_PORT = 5439;
+  // for annotations
+  static final String OPT_PORT_DEFAULT = "5439";
```
kpiotrowski, line 31: How about keeping only the private static variable and adding the getter?

kpiotrowski (approved):
Changes looks good, but can you explain what is the purpose? Do you plan to reuse the extracted util methods?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/redshift/RedshiftUrlUtil.java
misolt: The annotation requires a constant expression, which can't call any methods.

Another question here is why not use e.g. `String.valueOf(DEFAULT_PORT)`. This also doesn't work, because even though `OPT_PORT_DEFAULT` is still a static final field, it now _depends_ on a function being called, making it not a constant expression.

misolt: I could keep the field, but change the value from `"5439"` to `"" + DEFAULT_PORT`, keeping the previous trick to convert types at compile-time. I don't think either option is significantly better than the other.

PR #709, opened 2025-01-21
[381225149] upgraded jetty-server to 12.0.12

kaxuna:
[381225149] upgraded jetty-server to 12.0.12


shevek-google:
Good luck. Jetty changed a lot between 9 and 12. Also, I think we never actually run it; we might do just as well excluding the dependency.

kpiotrowski (approved):

PR #710, opened 2025-01-21
381225149 removed jetty from hive module

kaxuna:
completely removed jetty libraries from hive module


shevek-google (approved):
Seems fine to me. There's probably more one can pull out, without much effort.

kaxuna:
@shevek-google If you point some libraries I can pay them the attention. Otherwise I am reviewiving libraries under the MOSS

PR #713, opened 2025-01-27
Feat [b/382469906]: Add a task to dump Cloudera YARN applications

zaldis:
Add two optional CLI arguments:
- `--yarn-application-types RANGER,MAPREDUCE`
- `--pagination-page-size 1000` (it's 1000 by default)

Create 2 new zipped files:
- yarn-applications-90d.jsonl
- yarn-application-types-90d.jsonl

YARN application has the following structure:
```
{
    "clusterName": "cldr3-data-hub",
    "applicationId": "application_1736775069229_0005",
    "name": "livy-session-0",
    "state": "FINISHED",
    "startTime": "2025-01-13T14:24:26.887Z",
    "endTime": "2025-01-13T15:48:39.360Z",
    "user": "user-name",
    "pool": "default",
    "applicationTags": [
      "livy-session-0-0iw4scjj"
    ],
    "allocatedMemorySeconds": 77433697,
    "allocatedVcoreSeconds": 15121,
    "attributes": {}
  },
```

YARN application type has the following structure:
```
{
    "applicationId": "application_1736775069229_0950",
    "applicationType": "TEZ"
}
```


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaJobsTask.java
```diff
@@ -0,0 +1,66 @@
...
+
+    YarnClient yarnClient = YarnClient.createYarnClient();
+    yarnClient.init(conf);
+    LOG.warn(conf.get("hadoop.security.authentication"));
+    yarnClient.start();
+
+    List<ApplicationReport> applicationReports = yarnClient.getApplications();
+    applicationReports.forEach(
+        appReport ->
+            System.out.printf(
```
vladislav-sidorovich, line 61: Please use logs, not `System.out`.  
https://stackoverflow.com/questions/8601831/do-not-use-system-out-println-in-server-side-code

zaldis: Got it, thanks!

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
```diff
@@ -0,0 +1,181 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** The task dumps YARN applications from Cloudera Manager API */
+public class ClouderaYarnApplicationsTask extends AbstractClouderaManagerTask {
+  private static final Logger LOG = LoggerFactory.getLogger(ClouderaYarnApplicationsTask.class);
+  private static final DateTimeFormatter isoDateTimeFormatter =
+      DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
+
+  private final String[] predefinedAppTypes = {
```
vladislav-sidorovich, line 54: It's really unexpected, but in java all arrays are mutable. So, it's possible to do something like that `predefinedAppTypes[0] = "wrong value"`.
I think, your intention here is to have a `constant`. So, it's better to have `ImmutableList` instead of an array.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
```diff
@@ -0,0 +1,181 @@
...
+    private final String host;
+    private final CloseableHttpClient httpClient;
+
+    public YarnApplicationsLoader(String host, CloseableHttpClient httpClient) {
+      this.host = host;
+      this.httpClient = httpClient;
+    }
+
+    public List<ApiYARNApplicationDTO> load(String clusterName, String appType) {
+      List<ApiYARNApplicationDTO> yarnApplications = new LinkedList<>();
```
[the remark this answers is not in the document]
zaldis: Now I see, thanks!

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
```diff
@@ -0,0 +1,181 @@
...
+
+    List<ApiYARNApplicationDTO> totalYARNApplications = new LinkedList<>();
+    YarnApplicationsLoader appLoader =
+        new YarnApplicationsLoader(handle.getApiURI().toString(), handle.getHttpClient());
+    for (ClouderaClusterDTO cluster : handle.getClusters()) {
+      String clusterName = cluster.getName();
+      List<String> yarnAppTypes = fetchYARNApplicationTypes(handle, clusterName);
+      for (String yarnAppType : yarnAppTypes) {
+        LOG.info(
+            String.format(
```
vladislav-sidorovich, line 78: Formatting should be done with logging framework, not with `String.format`. 
```
LOG.info("Dump YARN applications with {} type from {} cluster", yarnAppType, clusterName));
```

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
```diff
@@ -0,0 +1,181 @@
...
+                "Dump YARN applications with %s type from %s cluster", yarnAppType, clusterName));
+        List<ApiYARNApplicationDTO> yarnApplications = appLoader.load(clusterName, yarnAppType);
+        totalYARNApplications.addAll(yarnApplications);
+        LOG.info(
+            String.format(
```
vladislav-sidorovich, line 83: The same here. Please use logger framework.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
zaldis: Yeah, great spot! Thanks.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
```diff
@@ -0,0 +1,181 @@
...
+                "Dumped %d YARN applications with %s type from %s cluster",
+                yarnApplications.size(), yarnAppType, clusterName));
+      }
+    }
+
+    String yarnAppsJson = parseObjectToJsonString(totalYARNApplications);
```
vladislav-sidorovich, line 89: It looks like we don't need `totalYARNApplications`. We can request application and write a response to the file immediately. 
We do not need to collect all responses to `totalYARNApplications` first and then write all them to file at once. In theory we can eve do it asynchronies, but let's not overcomplicate it for now.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
```diff
@@ -0,0 +1,181 @@
...
+
+  private String buildISODateTime(int deltaInDays) {
+    ZonedDateTime dateTime =
+        ZonedDateTime.of(LocalDateTime.now().minusDays(deltaInDays), ZoneId.of("UTC"));
+    return dateTime.format(isoDateTimeFormatter);
+  }
+
+  private List<String> fetchYARNApplicationTypes(ClouderaManagerHandle handle, String clusterName) {
+    List<String> yarnApplicationTypes = new ArrayList<>();
+    Collections.addAll(yarnApplicationTypes, predefinedAppTypes);
```
vladislav-sidorovich: You will remove it with replacement `predefinedAppTypes` type from array to immutable list.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
```diff
@@ -0,0 +1,181 @@
...
+
+    String yarnAppTypesUrl =
+        handle.getApiURI().toString() + "clusters/" + clusterName + "/serviceTypes";
+    CloseableHttpClient httpClient = handle.getHttpClient();
+    JsonNode yarnAppTypesJson;
+    ApiYARNApplicationTypeDTO yarnAppTypesFromClouderaAPI;
+    try (CloseableHttpResponse chart = httpClient.execute(new HttpGet(yarnAppTypesUrl))) {
+      yarnAppTypesJson = readJsonTree(chart.getEntity().getContent());
+      yarnAppTypesFromClouderaAPI =
+          parseJsonStringToObject(yarnAppTypesJson.toString(), ApiYARNApplicationTypeDTO.class);
```
vladislav-sidorovich, line 111: It seems we do not use `ApiYARNApplicationTypeDTO` anywhere outside this private method. So, we can avoid this class with something like this. 
```
JsonNode at = jsonNode.at("/items");
List<String> strings = objectMapper.convertValue(at, new TypeReference<>() {});
```

vladislav-sidorovich: it's also unclear why the json is 2 times:
1. read `JsonNode` from response
2. convert it to string
3. read object from string - here we also have implicit parsing of string, similar to step #1 

This method should be simplified.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
```diff
@@ -0,0 +1,177 @@
...
+      TaskRunContext context, @Nonnull ByteSink sink, @Nonnull ClouderaManagerHandle handle)
+      throws Exception {
+    List<ClouderaClusterDTO> clusters = handle.getClusters();
+    Preconditions.checkNotNull(
+        clusters, "Clusters must be initialized before fetching YARN applications.");
+
+    List<ApiYARNApplicationDTO> totalYARNApplications = new ArrayList<>();
+    YarnApplicationsLoader appLoader =
+        new YarnApplicationsLoader(handle.getApiURI().toString(), handle.getHttpClient());
+    for (ClouderaClusterDTO cluster : handle.getClusters()) {
```
vladislav-sidorovich, line 73: You already have `List<ClouderaClusterDTO> clusters` variable above. Please reuse, `handle.getClusters()` is synchronised.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
```diff
@@ -0,0 +1,177 @@
...
+        List<ApiYARNApplicationDTO> newLoad = load(clusterName, appType, limit, offset);
+        yarnApplications.addAll(newLoad);
+        nextLoad = (!newLoad.isEmpty());
+        offset += limit;
+      }
+
+      return yarnApplications;
+    }
+
+    private List<ApiYARNApplicationDTO> load(
```
vladislav-sidorovich, line 144: We assume that job history can be big. So, it's better: 
1. request page
2. write response to file
3. repeat

instead collecting all responses into single list in RAM.  We can tune page size, for example accept as an optional CLI parameter and keep `100` as default value.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
```diff
@@ -0,0 +1,177 @@
...
+          host + "clusters/" + clusterName + "/services/yarn/yarnApplications";
+      String fromDate = buildISODateTime(includedLastDays);
+      URI yarnApplicationsURI;
+      try {
+        URIBuilder uriBuilder = new URIBuilder(yarnApplicationsUrl);
+        uriBuilder.addParameter("limit", String.valueOf(limit));
+        uriBuilder.addParameter("offset", String.valueOf(offset));
+        uriBuilder.addParameter("from", fromDate);
+        uriBuilder.addParameter("filter", String.format("applicationType=%s", appType));
+        yarnApplicationsURI = uriBuilder.build();
```
vladislav-sidorovich, line 156: Logic of building `yarnApplicationsURI` looks like separate private method.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
```diff
@@ -0,0 +1,177 @@
...
+      } catch (URISyntaxException ex) {
+        throw new RuntimeException(ex.getMessage(), ex);
+      }
+
+      try (CloseableHttpResponse resp = httpClient.execute(new HttpGet(yarnApplicationsURI))) {
+        JsonNode yarnApplicationsRespJson = readJsonTree(resp.getEntity().getContent());
+        ApiYARNApplicationListDTO yarnAppListDto =
+            parseJsonStringToObject(
```
vladislav-sidorovich, line 164: It looks like we don't need this method. We can build object directly from `InuputStream` and avoid double parsing.
```
objectMapper.readValue(httpResponse.getEntity().getContent(), MyDTO.class)
```

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
```diff
@@ -0,0 +1,177 @@
...
+      }
+
+      return yarnApplications;
+    }
+
+    private List<ApiYARNApplicationDTO> load(
+        String clusterName, String appType, int limit, int offset) {
+      String yarnApplicationsUrl =
+          host + "clusters/" + clusterName + "/services/yarn/yarnApplications";
+      String fromDate = buildISODateTime(includedLastDays);
```
vladislav-sidorovich, line 148: This variable can be build only once for all class.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
```diff
@@ -0,0 +1,177 @@
...
+      yarnAppTypesFromClouderaAPI =
+          parseJsonStringToObject(yarnAppTypesJson.toString(), ApiYARNApplicationTypeDTO.class);
+      yarnApplicationTypes.addAll(yarnAppTypesFromClouderaAPI.getItems());
+    } catch (IOException ex) {
+      throw new RuntimeException(ex.getMessage(), ex);
+    }
+    return yarnApplicationTypes;
+  }
+
+  private class YarnApplicationsLoader {
```
vladislav-sidorovich, line 119: We don't need this class. As I see the logic of `ClouderaYarnApplicationsTask`:
For each cluster:
1. fetch all application types
2. fetch all jobs per application type with pagination. 

So, it's more like 2 private method of this task but not class.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
```diff
@@ -0,0 +1,177 @@
...
+      String yarnApplicationsUrl =
+          host + "clusters/" + clusterName + "/services/yarn/yarnApplications";
+      String fromDate = buildISODateTime(includedLastDays);
+      URI yarnApplicationsURI;
+      try {
+        URIBuilder uriBuilder = new URIBuilder(yarnApplicationsUrl);
+        uriBuilder.addParameter("limit", String.valueOf(limit));
+        uriBuilder.addParameter("offset", String.valueOf(offset));
+        uriBuilder.addParameter("from", fromDate);
+        uriBuilder.addParameter("filter", String.format("applicationType=%s", appType));
```
vladislav-sidorovich, line 155: I'm not sure if we need this parameter at all. it looks like we can remove `filter` and all applications will be returned, it will simplify the code.

vladislav-sidorovich:
As already discussed. The usage of Cloudera API for jobs history looks reasonable, there is a limitation with engine types so let's split 1 task in to 2 tasks:
1. Collect all jobs history without job-type
2. Collect mapping of will know job-types to job id

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
zaldis: Thanks, updated.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
zaldis: This parameter gives allows to get application with specific type. Otherwise it's not possible to identify which engine type is related to the specific YARN application. But it will be moved to the separated task as discussed (this task will be responsible for mappings between applications and application types).

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
zaldis: I added this class to simplify amount of parameters which should be passed to the private methods. As we need to pass pagination data (offset, limit), common parameters (URL host, httpClient), clusterName, application type.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
zaldis: Done. Moved this into the callback.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
zaldis: Done.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
zaldis: Moved.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
zaldis: Fixed.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
zaldis: This class has been moved into the abstract class.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
zaldis: Fixed.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
zaldis: Fixed.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
zaldis: Done.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaManagerTask.java
```diff
@@ -65,6 +65,10 @@ protected <T> T parseJsonStringToObject(String jsonString, Class<T> type)
     return objectMapper.readValue(jsonString, type);
   }
 
+  protected <T> T parseJsonStreamToObject(InputStream src, Class<T> type) throws IOException {
+    return objectMapper.readValue(src, type);
+  }
+
   protected String parseObjectToJsonString(Object obj) throws JsonProcessingException {
```
vladislav-sidorovich, line 72: It's part of a different PR, but object to string it's not `parse`, it serialization. I would like to suggest to reuse the names from    `objectMapper`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    public int load(String clusterName, Consumer<List<ApiYARNApplicationDTO>> callback) {
+      return load(clusterName, null, callback);
+    }
+
+    public int load(
+        String clusterName,
+        @Nullable String appType,
+        Consumer<List<ApiYARNApplicationDTO>> callback) {
```
vladislav-sidorovich, line 77: In general callback is fine, but you already introduced `abstract` class, so now you can introduce abstract method `writeBatch` or `writePage` and remove callback from method parameters.

vladislav-sidorovich: uh, the load is in separate class, not int `*ClouderaYarnApplicationTask`. Let's keep callback in this case.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
```diff
@@ -0,0 +1,131 @@
...
+        }
+        ApiYARNApplicationListDTO yarnAppListDto =
+            parseJsonStreamToObject(resp.getEntity().getContent(), ApiYARNApplicationListDTO.class);
+        return yarnAppListDto.getApplications();
+      } catch (IOException ex) {
+        throw new RuntimeException(ex.getMessage(), ex);
+      }
+    }
+
+    private URI buildYARNApplicationURI(String clusterName, @Nullable String appType) {
```
vladislav-sidorovich, line 112: `buildNextPageYARNApplicationURI`, with such name it will be more clear why this method called in a cycle many times with the same parameters.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
```diff
@@ -0,0 +1,131 @@
...
+      while (nextLoad) {
+        URI yarnAppsURI = buildYARNApplicationURI(clusterName, appType);
+        List<ApiYARNApplicationDTO> newLoad = load(yarnAppsURI);
+        if (!newLoad.isEmpty()) {
+          callback.accept(newLoad);
+        } else {
+          nextLoad = false;
+        }
+        amountOfLoadedApps += newLoad.size();
+        offset += limit;
```
vladislav-sidorovich, line 91: Should it be `newLoad.size()` instead of `limit`?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationTypeTask.java
```diff
@@ -0,0 +1,130 @@
...
+    } catch (IOException ex) {
+      throw new RuntimeException("Error: Can't dump YARN application types", ex);
+    }
+  }
+
+  private List<String> fetchYARNApplicationTypes(ClouderaManagerHandle handle, String clusterName) {
+    String yarnAppTypesUrl =
+        handle.getApiURI().toString() + "clusters/" + clusterName + "/serviceTypes";
+    CloseableHttpClient httpClient = handle.getHttpClient();
+    try (CloseableHttpResponse chart = httpClient.execute(new HttpGet(yarnAppTypesUrl))) {
```
vladislav-sidorovich, line 103: it's not `chart`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
```diff
@@ -0,0 +1,79 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** The task dumps YARN applications from Cloudera Manager API */
+public class ClouderaYarnApplicationsTask extends AbstractClouderaYarnApplicationTask {
+  private static final Logger LOG = LoggerFactory.getLogger(ClouderaYarnApplicationsTask.class);
+
+  public ClouderaYarnApplicationsTask(int days) {
+    super("yarn-applications", days);
+    Preconditions.checkArgument(
```
vladislav-sidorovich, line 38: You can check it only once in `AbstractClouderaYarnApplicationTask`

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/dto/ApiYARNApplicationDTO.java
```diff
@@ -0,0 +1,157 @@
...
+ */
+package com.google.edwmigration.dumper.application.dumper.connector.cloudera.manager.dto;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+import java.util.Map;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class ApiYARNApplicationDTO {
```
vladislav-sidorovich, line 25: Please add a link to Cloudera specification for this object.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
vladislav-sidorovich: `callback` is too generic, let's call it `onPageSuccessLoad` or something like that.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationTaskTest.java
```diff
@@ -0,0 +1,154 @@
...
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ClouderaYarnApplicationTaskTest {
+
+  private final ClouderaYarnApplicationsTask task = new ClouderaYarnApplicationsTask(30);
+
+  private ClouderaManagerHandle handle;
+  private static WireMockServer server;
```
vladislav-sidorovich, line 60: static fields should go before non-static.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaManagerTask.java
zaldis: I agree about `parsing` vs `serialization`. But some names in `objectMapper` are not obvious. For example the method `.writValueAsString(obj)`. This method only creates a String and doesn't write anything into the output stream.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.List;
+import java.util.function.Consumer;
+import javax.annotation.Nullable;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.client.utils.URIBuilder;
+import org.apache.http.impl.client.CloseableHttpClient;
+
+public abstract class AbstractClouderaYarnApplicationTask extends AbstractClouderaManagerTask {
+  private static final DateTimeFormatter isoDateTimeFormatter =
```
sayuzbas-google, line 38: name of final variables should uppercase with underscore:
`ISO_DATETIME_FORMATTER`

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
```diff
@@ -0,0 +1,131 @@
...
+      DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
+
+  private final String fromDate;
+
+  public AbstractClouderaYarnApplicationTask(String fileName, int includedLastDays) {
```
sayuzbas-google, line 43: nit: better to rename to `lastDaysToInclude` or just `daysToInclude`

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
```diff
@@ -0,0 +1,131 @@
...
+    super(String.format("%s-%dd.jsonl", fileName, includedLastDays));
+    Preconditions.checkArgument(
+        includedLastDays > 1,
```
sayuzbas-google, line 46: it is checking for values higher than 1, but message says "must be positive number". In that case it should be `>=1`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
zaldis: Great idea! I missed it when moved limit/offset parameters into class state. Thanks!

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
```diff
@@ -0,0 +1,131 @@
...
+        String.format("Amount of days must be a positive number. Get %d.", includedLastDays));
+    this.fromDate = buildISODateTime(includedLastDays);
+  }
+
+  private String buildISODateTime(int deltaInDays) {
```
sayuzbas-google, line 51: `convertToISODateTimeString`. Also name of argument is not understandable, better to rename.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
```diff
@@ -0,0 +1,131 @@
...
+  class PaginatedClouderaYarnApplicationsLoader {
+    private final String host;
+    private final CloseableHttpClient httpClient;
+    private final int limit;
+    private int offset;
+
+    public PaginatedClouderaYarnApplicationsLoader(ClouderaManagerHandle handle) {
+      this.host = handle.getApiURI().toString();
+      this.httpClient = handle.getHttpClient();
+      this.limit = 500;
```
sayuzbas-google, line 66: If this numbers are hardcoded, so better set them during declaration of field:
`private final int LIMIT = 500`

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
zaldis: Actually it doesn't matter. It's expected that API returns no more than `limit` rows per request. And in the last call the API should return an empty list.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
```diff
@@ -0,0 +1,131 @@
...
+      this.offset = 0;
```
sayuzbas-google, line 67: by default int types have 0 as value, no need to set it here.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
```diff
@@ -0,0 +1,131 @@
...
+    public int load(String clusterName, Consumer<List<ApiYARNApplicationDTO>> callback) {
+      return load(clusterName, null, callback);
+    }
+
+    public int load(
+        String clusterName,
+        @Nullable String appType,
+        Consumer<List<ApiYARNApplicationDTO>> callback) {
+      int amountOfLoadedApps = 0;
+      offset = 0;
```
sayuzbas-google, line 79: You have `offset` as class field and also setting to `0` every time when this `load` method is called and use it in `buildYARNApplicationURI`. So instead of it, you can just have it as local variable and pass it to `buildYARNApplicationURI` as argument, in this case, it will be easier to understand usage of it.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationTypeTask.java
```diff
@@ -0,0 +1,130 @@
...
+        new PaginatedClouderaYarnApplicationsLoader(handle);
+
+    try (Writer writer = sink.asCharSink(StandardCharsets.UTF_8).openBufferedStream()) {
+      for (ClouderaClusterDTO cluster : clusters) {
+        final String clusterName = cluster.getName();
+        Set<String> yarnAppTypes = new HashSet<>(fetchYARNApplicationTypes(handle, clusterName));
+        yarnAppTypes.addAll(predefinedAppTypes);
+        for (String yarnAppType : yarnAppTypes) {
+          LOG.info("Dump YARN applications with {} type from {} cluster", yarnAppType, clusterName);
+          int loadedAppsCnt =
```
sayuzbas-google, line 69: `loadedAppsCount`

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
zaldis: The main idea here was to minimise the amount of variables in the method. It's okay to reset the state of object as it's another call from it. Like running several times the iterator object. It's possible to do iteration on the same one or create two separated objects.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
```diff
@@ -0,0 +1,131 @@
...
+      return load(clusterName, null, onPageLoad);
+    }
+
+    public int load(
+        String clusterName,
+        @Nullable String appType,
+        Consumer<List<ApiYARNApplicationDTO>> onPageLoad) {
+      int amountOfLoadedApps = 0;
+      offset = 0;
+      boolean nextLoad = true;
```
kaxuna, line 80: renaming to shouldLoadNext would underline that this var is boolean

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
```diff
@@ -0,0 +1,131 @@
...
+      while (nextLoad) {
+        URI yarnAppsURI = buildNextYARNApplicationPageURI(clusterName, appType);
+        List<ApiYARNApplicationDTO> newLoad = load(yarnAppsURI);
+        if (!newLoad.isEmpty()) {
+          onPageLoad.accept(newLoad);
+        } else {
+          nextLoad = false;
+        }
+        amountOfLoadedApps += newLoad.size();
+        offset += limit;
```
kaxuna, line 91: should we increase offset even when newLoad is empty?

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationTaskTest.java
```diff
@@ -0,0 +1,153 @@
...
+    queryParams.put("offset", matching("500"));
+    stubYARNApplicationsAPI("test-cluster", queryParams, "{\"applications\": []}");
+
+    task.doRun(context, sink, handle);
+
+    server.verify(0, getRequestedFor(urlEqualTo("/api/vTest/clusters/test-cluster/serviceTypes")));
+    server.verify(
+        2,
+        getRequestedFor(
+            urlPathMatching("/api/vTest/clusters/test-cluster/services/yarn/yarnApplications.*")));
```
kajgol, line 104: This doesn't really test if the task was successful, only that the correct url was called.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
```diff
@@ -0,0 +1,131 @@
...
+      }
+      return amountOfLoadedApps;
+    }
+
+    private List<ApiYARNApplicationDTO> load(URI yarnAppURI) {
+      try (CloseableHttpResponse resp = httpClient.execute(new HttpGet(yarnAppURI))) {
+        final int statusCode = resp.getStatusLine().getStatusCode();
```
kaxuna, line 98: final is extra since scope of statusCode is this function

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
zaldis: Done.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
zaldis: Great spot! Thanks!

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
zaldis: Updated.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
zaldis: Updated as a parameter.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
zaldis: Fixed.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationTypeTask.java
zaldis: Done.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
zaldis: Done.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
zaldis: It doesn't break anything, but I can move it into the `if` block

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
zaldis: Fixed.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
```diff
@@ -0,0 +1,131 @@
...
+  private static final DateTimeFormatter isoDateTimeFormatter =
+      DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
+
+  private final String fromDate;
+
+  public AbstractClouderaYarnApplicationTask(String fileName, int includedLastDays) {
+    super(String.format("%s-%dd.jsonl", fileName, includedLastDays));
+    Preconditions.checkArgument(
+        includedLastDays > 1,
+        String.format("Amount of days must be a positive number. Get %d.", includedLastDays));
```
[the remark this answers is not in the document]
zaldis: Thanks!

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
```diff
@@ -0,0 +1,131 @@
...
+    public int load(String clusterName, Consumer<List<ApiYARNApplicationDTO>> onPageLoad) {
+      return load(clusterName, null, onPageLoad);
+    }
+
+    public int load(
+        String clusterName,
+        @Nullable String appType,
+        Consumer<List<ApiYARNApplicationDTO>> onPageLoad) {
+      int amountOfLoadedApps = 0;
+      offset = 0;
```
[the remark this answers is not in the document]
zaldis: Hmm, in current implementation it's not 100%, but I think it's possible to refactor load logic to use only one variable.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
```diff
@@ -0,0 +1,131 @@
...
+    private int offset;
+
+    public PaginatedClouderaYarnApplicationsLoader(ClouderaManagerHandle handle) {
+      this.host = handle.getApiURI().toString();
+      this.httpClient = handle.getHttpClient();
+      this.limit = 500;
+      this.offset = 0;
+    }
+
+    public int load(String clusterName, Consumer<List<ApiYARNApplicationDTO>> onPageLoad) {
```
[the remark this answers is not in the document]
zaldis: It's not a single function, but a method in the PaginatedClouderaYarnApplicationsLoader class. So it already should give a context that YARN applications should be loaded.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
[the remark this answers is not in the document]
zaldis: Fixed.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
```diff
@@ -0,0 +1,131 @@
...
+      return load(clusterName, null, onPageLoad);
+    }
+
+    public int load(
+        String clusterName,
+        @Nullable String appType,
+        Consumer<List<ApiYARNApplicationDTO>> onPageLoad) {
+      int amountOfLoadedApps = 0;
```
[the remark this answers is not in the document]
zaldis: Updated.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
```diff
@@ -0,0 +1,131 @@
...
+    this.fromDate = buildISODateTime(includedLastDays);
+  }
+
+  private String buildISODateTime(int deltaInDays) {
+    ZonedDateTime dateTime =
+        ZonedDateTime.of(LocalDateTime.now().minusDays(deltaInDays), ZoneId.of("UTC"));
+    return dateTime.format(isoDateTimeFormatter);
+  }
+
+  class PaginatedClouderaYarnApplicationsLoader {
```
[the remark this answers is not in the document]
zaldis: Unfortunately it can't. This class is related to the object based field `fromDate`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationTypeTask.java
```diff
@@ -0,0 +1,130 @@
...
+      return StreamSupport.stream(json.get("items").spliterator(), false)
+          .map(JsonNode::asText)
+          .collect(Collectors.toList());
+    } catch (IOException ex) {
+      throw new RuntimeException(ex.getMessage(), ex);
+    }
+  }
+}
+
+class ApplicationTypeToYarnApplication {
```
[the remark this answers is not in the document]
zaldis: Great idea! Thanks.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
```diff
@@ -0,0 +1,131 @@
...
+        offset += limit;
+      }
+      return amountOfLoadedApps;
+    }
+
+    private List<ApiYARNApplicationDTO> load(URI yarnAppURI) {
+      try (CloseableHttpResponse resp = httpClient.execute(new HttpGet(yarnAppURI))) {
+        final int statusCode = resp.getStatusLine().getStatusCode();
+        if (!isStatusCodeOK(statusCode)) {
+          throw new RuntimeException(
```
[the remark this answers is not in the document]
zaldis: For what reason this exception will be useful. There is no custom handling for yarn application exceptions.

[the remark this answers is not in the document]
kajgol: Have a read: https://engdoc.corp.google.com/eng/doc/devguide/java/practices/exceptions.md?cl=head#abstract_throwables

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
```diff
@@ -0,0 +1,131 @@
...
+        ApiYARNApplicationListDTO yarnAppListDto =
+            parseJsonStreamToObject(resp.getEntity().getContent(), ApiYARNApplicationListDTO.class);
+        return yarnAppListDto.getApplications();
+      } catch (IOException ex) {
+        throw new RuntimeException(ex.getMessage(), ex);
+      }
+    }
+
+    private URI buildNextYARNApplicationPageURI(String clusterName, @Nullable String appType) {
+      String yarnApplicationsUrl =
```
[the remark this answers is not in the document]
zaldis: This method overrides already provided url path. It's might be a collision if ClouderaManagerHandle::host contains part of URI path.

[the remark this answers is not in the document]
zaldis: Updated.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationTaskTest.java
zaldis: Added asserts. Also I'm still working on another test cases.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/dto/ApiYARNApplicationDTO.java
zaldis: Added.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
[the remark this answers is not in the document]
kajgol: The field can be passed in the constructor. It also uses some methods from the base object, which arguably should also be static since they do not depend on the object state.

sayuzbas-google (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/AbstractClouderaYarnApplicationTask.java
```diff
@@ -0,0 +1,127 @@
+/*
```
misolt, line 1: It's a pretty big change :)
I'll start reviewing, but it may take a while. My suggestion would be to break this up into smaller parts, in the 200-300 lines range. If the PR is done or almost done, it might not be worth it, but in that case I'd still keep that in mind for the future.

Also note that the entire PR will appear as a single commit after the merge to main.

zaldis: PR is done. But yeah I agree to separate it into smaller parts in future.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaYarnApplicationsTask.java
```diff
@@ -0,0 +1,79 @@
...
+import java.io.IOException;
+import java.io.Writer;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+import javax.annotation.Nonnull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** The task dumps YARN applications from Cloudera Manager API */
+public class ClouderaYarnApplicationsTask extends AbstractClouderaYarnApplicationTask {
```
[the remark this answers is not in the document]
zaldis: Thanks!

PR #728, opened 2025-02-04
[ops] Schedule dependabot weekly

vladislav-sidorovich:


.github/dependabot.yml
```diff
@@ -3,7 +3,10 @@ updates:
   - package-ecosystem: "gradle"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
+      day: "monday"
+      time: "05:00"
```
kajgol, line 8: Nit: Probably wouldn't make any difference here, but a good practice is to pick some random time (like 5:13) to avoid a pileup at the start of an hour.

kajgol (approved):

zaldis (approved):

.github/dependabot.yml
vladislav-sidorovich: good point. I will keep it as is for the current minor process, but in general the point is good.

PR #730, opened 2025-02-05
[b/360552594] Remove redundant logger frameworks and use single version of logback

vladislav-sidorovich:
Before this change we had 3 versions of logback 1.0.9, 1.2.10, 1.3.15 in different scopes and 3 reloaders at runtime.

1. Module `lib-ext-hive-metastore`
Exclude `co.cask.tephra:*` from Hive client. Dumper doesn't work with HBase at the moment, so it's safe. As the result transient dependency on old version of logback is removed.

2. Module `app`
Exclude transient reloaders from `hadoop.common`, so one version of logback is used as implementation of logger framework.


rocgranada (approved):

misolt (approved):

kajgol (approved):

dumper/app/build.gradle
```diff
@@ -69,16 +72,13 @@ dependencies {
     implementation libs.aws.java.sdk.redshift
     implementation libs.aws.java.sdk.cloudwatch
 
-    runtimeOnly libs.logback.classic
+    implementation libs.logback.classic
```
shevek-google, line 75: This should remain runtimeOnly - it's the runtime implementation, not the API. The API is all in the slf4j packages.

If you need this in the implementation, something else is wrong. I assume you're familiar with the images in https://docs.gradle.org/current/userguide/java_library_plugin.html ?

shevek-google (changes requested):

dumper/app/build.gradle
shevek-google: Followup notes: If it's not obvious, what caused you to put this in the implementation? Did you get a LinkageError of some sort? Did someone statically link against a logback class? Happy to help, just please let me know what triggered this change, please.

vladislav-sidorovich: changed back to `runtimeOnly`

zaldis (approved):

PR #749, opened 2025-02-21
Make descriptions mandatory for Snowflake connectors

misolt:


vladislav-sidorovich:
The approach is definitely working, but I didn't get the motivation for changes.

```
@Description("Some useful text")
class MyLovelyConnector {}
```
vs
```
class MyLovelyConnector {
  public String getDescription() {
    return "Some useful text";
  }
}
```

To be honest, the previous approach looks more suitable. Annotations for meta-information, methods - behaviour. 

What the problem we are trying to solve here?

misolt:
> The approach is definitely working, but I didn't get the motivation for changes.
> 
> ```
> @Description("Some useful text")
> class MyLovelyConnector {}
> ```
> 
> vs
> 
> ```
> class MyLovelyConnector {
>   public String getDescription() {
>     return "Some useful text";
>   }
> }
> ```
> 
> To be honest, the previous approach looks more suitable. Annotations for meta-information, methods - behaviour.
> 
> What the problem we are trying to solve here?

But to the developer it's not really useful. Most of the time it's just `FooBarConnector -> "Dumps bar from Foo."`. Even if it was useful to us, I'd still say Javadoc is better than the annotation.

The main purpose of this description is behavior, because description is what the user sees in help text. With the getter we know that if it compiles, then no description is missing.

In general an annotation is more powerful than a getter. For example, we don't need the functionality of "iterate over every class that has a description and do X for it".

There's also no need to import an interface from another subproject if the only use case of the interface is to return the same string that was provided to it earlier.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/AbstractConnector.java
```diff
@@ -30,6 +31,16 @@ public AbstractConnector(@Nonnull String name) {
     this.name = Preconditions.checkNotNull(name, "Name was null.");
   }
 
+  @Nonnull
+  public String getDescription() {
+    Description description = getClass().getAnnotation(Description.class);
+    if (description != null) {
+      return description.value();
+    } else {
+      return "";
```
sayuzbas-google, line 40: nit: you can remove `else`, just `return ""`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/AbstractConnector.java
```diff
@@ -30,6 +31,16 @@ public AbstractConnector(@Nonnull String name) {
     this.name = Preconditions.checkNotNull(name, "Name was null.");
   }
 
+  @Nonnull
```
sayuzbas-google, line 34: should not it this method has `@Override` annotation, as it overrides `getDescription` from `Connector`

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/AbstractConnector.java
misolt: Done

sayuzbas-google (approved):

PR #750, opened 2025-02-21
Add sizeIfKnown for GCS source and also fix some other issues

samdgupi:
1. Added sizeIfKnown for GCS byte source 
2. Fixed naming for cloud run jobs
3. Made cloud run job names random to isolate runs. We should make the command parameteraized if possible to create one job and then update different parameters for different instances.

### DIFF
--- a/dbsync/client/src/main/java/com/google/edwmigration/dbsync/client/GcsClientMain.java
+++ b/dbsync/client/src/main/java/com/google/edwmigration/dbsync/client/GcsClientMain.java
@@ -16,6 +16,12 @@ public class GcsClientMain {
             .ofType(String.class)
             .required();
 
+    private final OptionSpec<String> locationOptionSpec =
+        parser.accepts("location", "Specifies the gcp location")
+            .withRequiredArg()
+            .ofType(String.class)
+            .required();
+
     private final OptionSpec<String> targetOptionSpec =
         parser.accepts("target_file", "Specifies the target file")
             .withRequiredArg()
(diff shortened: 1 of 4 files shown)


dbsync/client/src/main/java/com/google/edwmigration/dbsync/client/GcsClientMain.java
```diff
@@ -56,14 +68,18 @@ public String getStagingBucket() {
...
+    // This is needed as URI.resolve do not add '/' automatically
+    String stagingBucket = arguments.getStagingBucket().endsWith("/") ? arguments.getStagingBucket()
+        : arguments.getStagingBucket() + "/";
     try {
       client.putRsync(
           arguments.getProject(),
+          arguments.getLocation(),
           new URI(arguments.getSourceUri()),
-          new URI(arguments.getStagingBucket()),
+          new URI(stagingBucket),
```
misolt, line 79: Repeating slashes are valid, so it might be simpler to do:
``` java
// This is needed as URI.resolve does not add '/' automatically
new URI(arguments.getStagingBucket() + "/"),
```
and remove lines 71-73

misolt (approved):

dbsync/client/src/main/java/com/google/edwmigration/dbsync/client/GcsClientMain.java
samdgupi: Makes sense and tested

PR #758, opened 2025-02-25
[OPS] simplify git ignore files

vladislav-sidorovich:
Simplify `.gitignore` rules and hierarchy


kajgol (approved):

.gitignore
```diff
@@ -21,3 +21,6 @@ node_modules
 /*.tmp
 .DS_Store
 /progress.log.*
+/dbsync/**/bin
```
misolt, line 24: If there is a package named `bin` in `src`, will it also be ignored under the new rules?

vladislav-sidorovich: yes.

sayuzbas-google (approved):

.gitignore
misolt: - The code is divided into 11 subprojects and I don't know why it should be more than 2 or 3, so I don't know if it should be single `.gitignore` or separate files - no opinion here.
- Even if it should be a single file, my personal preference is to write the boilerplate and never worry about path edge cases. It should be <20 lines in total.

I remember when I changed it before, it used to be even broader, something like `**/bin` iirc

vladislav-sidorovich: To keep single file is much easier to support if we will have some edge cases we will fix it if once it will happen. Specially in this case something will not be committed, so it will be compile error in other words it's safe even if someone will miss it at first look.

misolt: I looked for some examples of this, if the wildcard is used maybe we can follow the example in this [template](https://github.com/github/gitignore/blob/main/Gradle.gitignore)?

``` .gitignore
**/build/
!src/**/build/
```

misolt (approved):

PR #760, opened 2025-02-25
[OPS] Update README.md

vladislav-sidorovich:
Java 8 is required for compiling

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ and generating an archive of DDL metadata or logs for consumption by Assessment
 or the Translation Service. For more information, [read the tool
 documentation](https://cloud.google.com/bigquery/docs/generate-metadata).
 
-    The Dumper requires Java 8 or higher. To check Java version run the command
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -13,7 +13,7 @@ and generating an archive of DDL metadata or logs for consumption by Assessment
 or the Translation Service. For more information, [read the tool
 documentation](https://cloud.google.com/bigquery/docs/generate-metadata).
 
-    The Dumper requires Java 8 or higher. To check Java version run the command
+    Compiling the Dumper from source requires `Java 8`, running the Dumper requires `Java 8` or higher. To check Java version run the command
```
misolt, line 16: Java 11 also compiles

vladislav-sidorovich: that's true, but our target is Java 8.  For example Cloudera/Hadoop env works with Java 8, so we will not able to run Dumper compiled by Java 11 in Hadoop envs.

PR #762, opened 2025-02-25
[b/390590500] url param added to Airflow connector

vladislav-sidorovich:
Airflow connector support url parameter or host,port,schema.

Next steps:
1. check all tables dumped for clarified requirements
2. add start/end date parameters

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
@@ -704,6 +704,10 @@ public class ConnectorArguments extends DefaultArguments {
     return getOptions().valueOf(optionUri);
   }
 
+  public boolean hasUri() {
+    return has(optionUri);
+  }
+
   @Nonnull
   public String getHostOrDefault() {
     return getHost(OPT_HOST_DEFAULT);
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
@@ -768,6 +768,10 @@ public class ConnectorArguments extends DefaultArguments {
         .collect(toImmutableList());
   }
 
+  public boolean isDatabasesProvided() {
+    return has(optionDatabase);
+  }
+
   @CheckForNull
   public String getSchema() {
     return getOptions().valueOf(optionSchema);
@@ -846,6 +850,10 @@ public class ConnectorArguments extends DefaultArguments {
     } else {
       return passwordReader.getOrPrompt();
     }
+  }
+
+  public boolean isPasswordFlagProvided() {
+    return getOptions().has(optionPass);
   }
 
   @CheckForNull
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/Connector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/Connector.java
@@ -35,6 +35,17 @@ public interface Connector {
   @Nonnull
   String getDefaultFileName(boolean isAssessment, Clock clock);
 
+  /**
+   * Validates if the cli parameters passed to the particular connector are expected. The method is
+   * called before {@link Connector#open(ConnectorArguments)} and {@link Connector#addTasksTo(List,
+   * ConnectorArguments)}
+   *
+   * @param arguments cli params
+   * @throws IllegalArgumentException if incorrect set of arguments passed to the particular
+   *     connector
+   */
+  default void validate(ConnectorArguments arguments) throws Exception {}
+
   void addTasksTo(@Nonnull List<? super Task<?>> out, @Nonnull ConnectorArguments arguments)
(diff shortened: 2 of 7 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/Connector.java
```diff
@@ -35,6 +35,17 @@ enum DefaultProperties implements ConnectorProperty {}
   @Nonnull
   String getDefaultFileName(boolean isAssessment, Clock clock);
 
+  /**
+   * Validates if the cli parameters passed to the particular connector are expected. The method is
+   * called before {@link Connector#open(ConnectorArguments)} and {@link Connector#addTasksTo(List,
+   * ConnectorArguments)}
+   *
+   * @param arguments cli params
+   * @throws IllegalArgumentException if incorrect set of arguments passed to the particular
```
sayuzbas-google, line 44: It says throws `IllegalArgumentException`, but in method declaration it throws `Exception`

vladislav-sidorovich: fixed, thanks.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -844,6 +852,10 @@ public String getPasswordOrPrompt() {
     }
   }
 
+  public boolean isPasswordFlagProvided() {
+    return getOptions().has(optionPass);
```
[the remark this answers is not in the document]
vladislav-sidorovich: fixed

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/Connector.java
sayuzbas-google: It is better to throw specific exception, rather than general one.

sayuzbas-google (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/Connector.java
```diff
@@ -35,6 +35,17 @@ enum DefaultProperties implements ConnectorProperty {}
...
+  /**
+   * Validates if the cli parameters passed to the particular connector are expected. The method is
+   * called before {@link Connector#open(ConnectorArguments)} and {@link Connector#addTasksTo(List,
+   * ConnectorArguments)}
+   *
+   * @param arguments cli params
+   * @throws Exception if incorrect set of arguments passed to the particular
+   *     connector
+   */
+  default void validate(ConnectorArguments arguments) throws Exception {}
```
misolt, line 47: If the purpose of the method is to validate, then I'd remove `throws Exception` and communicate success or failure via the return value.

If it has to throw an exception, I'd prefer throwing just `MetadataDumperUsageException`, which matches the conditions in method javadoc.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/Connector.java
vladislav-sidorovich: In general I agree, in the current case there are no differences because this is step No.2 in Dumper and the only handling here is to finish the program execution.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/Connector.java
misolt: There is different handling for `MetadataDumperUsageException` vs plain `Exception`:
- [MDUException handling](https://github.com/google/dwh-migration-tools/blob/main/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/Main.java#L72)
- [Exception handling](https://github.com/google/dwh-migration-tools/blob/main/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/Main.java#L78)

If the end goal is to throw an usage exception, I'd use `throws MetadataDumperUsageException` and throw the exception that we want in the end right in this method.

I wouldn't use `IllegalStateException` or other `RuntimeException` subclasses that we don't define ourselves - these are associated with the programmer making a mistake in the implementation, but in this case the error was caused by the user.

vladislav-sidorovich: Uh, yes that right. there is `if` in handling but it's ok. 

`RuntimeException`s are ok, we use them a lot in our jopt framework to work with command line and usage of `Preconditions` is more convenient in the method implementation. 

To clarify:
1. we can use any type of exception, it will not cause issues. 
2. To follow best practices we can use some specific type of exception (I see only readability benefits here)
3. However, customize an exception type (point No2) will reduce readability of  implementations with `Preconditions` class. 

So, I will chose point No3 as more valuable compare to No2.

misolt: I think it's a general discussion about `Preconditions` usage. I'm resolving the PR not to delay the change anymore, but would be great to also talk about `Preconditions` when there's more time.

PR #766, opened 2025-02-28
[b/390590500] Airflow dates params

vladislav-sidorovich:
Add date range support to airflow connector.

Usage:

1. `--lookback-days=x` -> date range: `[now-x, now)`
2. `--start-date=x --end-date=y` -> date range: `[x, y)`
3. `-start-date=x --lookback-days=y` -> date range: `[x-y, x)`

If date range is provided, sql queries will be generate for each day in the interval.


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
```diff
@@ -95,24 +116,89 @@ public void addTasksTo(@Nonnull List<? super Task<?>> out, @Nonnull ConnectorArg
...
+        "task_instance.csv",
+        "select * from task_instance ",
+        zonedIntervals,
+        (startDate, endDate) ->
+            String.format(
+                " where end_date >= CAST( '%s' as TIMESTAMP) and end_date < CAST( '%s' as TIMESTAMP) ;",
+                dateToSqlFormat(startDate), dateToSqlFormat(endDate)));
     // Airflow v1.6.0
-    addFullTable(out, "dag_run.csv", "select * from dag_run;");
+    addTableTaskWithIntervals(
```
vladislav-sidorovich, line 133: I'm not sure if we should split the query into intervals or just limit the dates. The answer is depends on the data size in DB.
In theory it can be big, not sure if it's a case in practice.

vladislav-sidorovich: I will keep it in one file for now, because consuming multiple files for metadata may require more work.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
```diff
@@ -144,12 +265,45 @@ public void validate(ConnectorArguments arguments) throws IllegalStateException
...
     }
+
+    validateDatesInterval(arguments);
+  }
+
+  private void validateDatesInterval(ConnectorArguments arguments) {
+    // valid inputs:
+    // 1. all null
+    // 2. loockback days
+    // 3. loocback with start
```
kajgol, line 276: shouldn't it be end?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
```diff
@@ -144,12 +265,45 @@ public void validate(ConnectorArguments arguments) throws IllegalStateException
...
+    // 4. start, end
```
kajgol, line 277: wouldn't a start date also be valid meaning [start;  now) ? I think it could be useful if there was a dump taken on some date and then we want to get new data starting from that date.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
```diff
@@ -95,24 +117,123 @@ public void addTasksTo(@Nonnull List<? super Task<?>> out, @Nonnull ConnectorArg
...
     // analog of DAG's python definition in json
-    addFullTable(out, "serialized_dag.csv", "select * from serialized_dag;", TaskCategory.OPTIONAL);
+    addQueryTask(out, "serialized_dag.csv", "select * from serialized_dag;", TaskCategory.OPTIONAL);
     // Airflow v1.10.10
-    addFullTable(out, "dag_code.csv", "select * from dag_code;", TaskCategory.OPTIONAL);
+    addQueryTask(out, "dag_code.csv", "select * from dag_code;", TaskCategory.OPTIONAL);
+  }
+
+  @Nullable
+  private ZonedIntervalIterable fromAirflowConnectorArguments(ConnectorArguments arguments) {
```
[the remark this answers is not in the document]
vladislav-sidorovich: I'm open for any ideas, let's be consistent with `ZonedIntervalIterableGenerator`.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
```diff
@@ -144,12 +265,45 @@ public void validate(ConnectorArguments arguments) throws IllegalStateException
...
       Preconditions.checkState(arguments.getSchema() != null, "--schema is required with --host");
     }
+
+    validateDatesInterval(arguments);
+  }
+
+  private void validateDatesInterval(ConnectorArguments arguments) {
+    // valid inputs:
+    // 1. all null
+    // 2. loockback days
```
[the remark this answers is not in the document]
vladislav-sidorovich: thanks.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -229,6 +232,37 @@ public class ConnectorArguments extends DefaultArguments {
           .withOptionalArg()
           .describedAs("sekr1t");
 
+  private final OptionSpec<ZonedDateTime> optionStartDate =
+      parser
+          .accepts(
+              OPT_START_DATE,
+              "Inclusive start date for data to export, value will be truncated to hour")
```
kpiotrowski, line 239: This description is very generic and suggest that these parameters can be used to limit time range of data exported by the dumper but actually they are used only by the Airflow connector. I suggest to mention that they are valid for airflow connector only or support them in all other connectors (in addition to query-log-*).

Did you think about making the `query-log-*` flags deprecated? The one ones could eventually replace existing ones.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
```diff
@@ -95,24 +117,123 @@ public void addTasksTo(@Nonnull List<? super Task<?>> out, @Nonnull ConnectorArg
...
+            : "select * from task_instance "
+                + String.format(
+                    " where end_date >= CAST( '%s' as TIMESTAMP) and end_date < CAST( '%s' as TIMESTAMP) ;",
+                    dateToSqlFormat(dateRange.getLeft()), dateToSqlFormat(dateRange.getRight())));
+
     // Airflow v1.6.0
-    addFullTable(out, "dag_run.csv", "select * from dag_run;");
+    addQueryTask(
+        out,
+        "dag_run.csv",
```
kpiotrowski, line 161: Looks like `dag_run.csv` and `task_instance.csv` use very similar SQL query. How about moving the logic that build query text to a separate function that takes (table_name, dateRange) as arguments to reduce duplicated code?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
```diff
@@ -95,24 +117,123 @@ public void addTasksTo(@Nonnull List<? super Task<?>> out, @Nonnull ConnectorArg
...
+        "Date range for query generation from {} to {} exclusive and increments of one day.",
+        dateRange.getLeft(),
+        dateRange.getRight());
+    return ZonedIntervalIterableGenerator.forDateRangeWithIntervalDuration(
+        dateRange.getLeft(), dateRange.getRight(), Duration.ofDays(1));
+  }
+
+  @Nullable
+  private Pair<ZonedDateTime, ZonedDateTime> dateRange(ConnectorArguments arguments) {
+    boolean isDateRangeSpecified =
```
kpiotrowski, line 194: nit: I think `isDateRangeSpecified` is redundant. You can immediately return if (arguments.getStartDate() == null || arguments.getLookbackDays() == null)

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
```diff
@@ -144,12 +265,45 @@ public void validate(ConnectorArguments arguments) throws IllegalStateException
       Preconditions.checkState(arguments.getPort() != null, "--port is required with --host");
       Preconditions.checkState(arguments.getSchema() != null, "--schema is required with --host");
     }
+
+    validateDatesInterval(arguments);
+  }
+
+  private void validateDatesInterval(ConnectorArguments arguments) {
```
kpiotrowski, line 272: Shouldn't the validation be executed on higher level than the airflow connector (or at least function moved to some shared space)? New flags are not defined in the scope of the airflow connector and they could be used by other connectors too.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
```diff
@@ -95,24 +117,123 @@ public void addTasksTo(@Nonnull List<? super Task<?>> out, @Nonnull ConnectorArg
...
+    boolean isDateRangeSpecified =
+        arguments.getStartDate() != null || arguments.getLookbackDays() != null;
+    if (!isDateRangeSpecified) {
+      return null;
+    }
+
+    ZonedDateTime startDate;
+    ZonedDateTime endDate;
+    if (arguments.getLookbackDays() != null) {
+      endDate = arguments.getStartDate(ZonedDateTime.now(ZoneOffset.UTC));
```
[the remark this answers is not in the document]
vladislav-sidorovich: It's combined with `lookbackDays` not with `lookforward*`. I have unit-test to cover this logic `validate_startDateAndLookbackDays_success`, feel free to check.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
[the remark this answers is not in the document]
vladislav-sidorovich: sounds good, done.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
vladislav-sidorovich: nope, it's start. There is unit-test for that and the description in this PR as well.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
vladislav-sidorovich: I would say that all the sql queries are similar here, however I would not extract a new function for that reason. 

Downside of removing "duplication" is introduction coupling between the components. In the current case those tables are independent and queries may change in the future independently, so we do not need to couple the components.   

The question: `Will these queries change at the same time for the same reason in the future?`
The answer: `No, the tables are independent.`

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
```diff
@@ -95,24 +117,123 @@ public void addTasksTo(@Nonnull List<? super Task<?>> out, @Nonnull ConnectorArg
...
 
+    ZonedIntervalIterable zonedIntervals = fromAirflowConnectorArguments(arguments);
+    Pair<ZonedDateTime, ZonedDateTime> dateRange = dateRange(arguments);
+
     // Airflow v1.5.0
-    addFullTable(out, "dag.csv", "select * from dag;");
-    addFullTable(out, "task_instance.csv", "select * from task_instance;");
+    addQueryTask(out, "dag.csv", "select * from dag;");
+    // todo think about multiple files
+    // addTableTaskWithIntervals(
```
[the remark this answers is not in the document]
vladislav-sidorovich: Well, it's just and example of API call, but I agree it's `dead code`.

Ok, let's move it to a bit different comment `/**`, so now it's a javadoc 😎 and not a dead code anymore.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
vladislav-sidorovich: That right, but there is a technic of self documented code. The variable is introduced to describe the boolean expression instead of writing a comment.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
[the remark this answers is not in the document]
vladislav-sidorovich: that's true. However `start` alway used with an additional parameter:
1. start + end
2. start + lookback

So, the combination of the parameters doesn't look misleading to me. But let's keep the conversation open to collect more opinions if any.

kpiotrowski:
-start-date=x --lookback-days=y -> date range: [x-y, x)

This usage is not intuitive, at least for me. Usually when I provide start date I want all the data to be later than this. I think `end-date` works better with a `lookback-days` because lookback refers to searching N days in the past.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -229,6 +232,37 @@ public class ConnectorArguments extends DefaultArguments {
...
+          .withOptionalArg()
+          .ofType(Date.class)
+          .withValuesConvertedBy(ZonedParser.withDefaultPattern(DayOffset.START_OF_DAY))
+          .describedAs("2001-01-15[ 00:00:00.[000]]");
+
+  private final OptionSpec<ZonedDateTime> optionEndDate =
+      parser
+          .accepts(
+              OPT_END_DATE,
+              "Inclusive end date for data to export, value will be truncated to hour")
```
zaldis, line 249:
```suggestion
              "Exclusive end date for data to export, value will be truncated to hour")
```
Based on PR description this parameter should not be inclusive.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -229,6 +232,37 @@ public class ConnectorArguments extends DefaultArguments {
...
+          .withOptionalArg()
+          .ofType(Date.class)
+          .withValuesConvertedBy(ZonedParser.withDefaultPattern(DayOffset.START_OF_DAY))
+          .describedAs("2001-01-15[ 00:00:00.[000]]");
+
+  private final OptionSpec<Integer> optionLookbackDays =
+      parser
+          .accepts(
+              OPT_LOOKBACK_DAYS,
+              "The N days back of data to export from current moment. "
```
zaldis, line 259: Please, add more details about `start_date` parameter (it's not always a current moment).

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/annotations/RespectsArgumentQueryLogDays.java
```diff
@@ -35,5 +35,5 @@
     description = RespectsArgumentQueryLogDays.DESCRIPTION)
 public @interface RespectsArgumentQueryLogDays {
 
-  public static final String DESCRIPTION = "The number of days of query history to dump.";
+  String DESCRIPTION = "The number of days of query history to dump.";
```
zaldis, line 38: Shouldn't it be with `final` modifier?
```suggestion
  final String DESCRIPTION = "The number of days of query history to dump.";
```

zaldis: Or looks like it's final by default

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
```diff
@@ -95,24 +117,107 @@ public void addTasksTo(@Nonnull List<? super Task<?>> out, @Nonnull ConnectorArg
...
     out.add(new FormatTask(FORMAT_NAME));
 
+    ZonedIntervalIterable zonedIntervals = queryRangeFromConnectorArguments(arguments);
+    Pair<ZonedDateTime, ZonedDateTime> dateRange = dateRange(arguments);
+
     // Airflow v1.5.0
-    addFullTable(out, "dag.csv", "select * from dag;");
-    addFullTable(out, "task_instance.csv", "select * from task_instance;");
+    addQueryTask(out, "dag.csv", "select * from dag;");
+    // todo think about multiple files with this.addTableTaskWithIntervals()
```
zaldis, line 125: Any bug ticket for this?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
```diff
@@ -95,24 +117,107 @@ public void addTasksTo(@Nonnull List<? super Task<?>> out, @Nonnull ConnectorArg
...
+    addQueryTask(
+        out,
+        "task_instance.csv",
+        dateRange == null
+            ? "select * from task_instance;"
+            : "select * from task_instance "
+                + String.format(
+                    " where end_date >= CAST( '%s' as TIMESTAMP) and end_date < CAST( '%s' as TIMESTAMP) ;",
+                    dateToSqlFormat(dateRange.getLeft()), dateToSqlFormat(dateRange.getRight())));
```
zaldis, line 134: If pre-build the filter query separately it might be easier to read and debug.
```suggestion
    String queryFilter = "";
    if (dateRange != null) {
      queryFilter = String.format(
           "where end_date >= CAST( '%s' as TIMESTAMP) and end_date < CAST( '%s' as TIMESTAMP) ;",
           dateToSqlFormat(dateRange.getLeft()),
           dateToSqlFormat(dateRange.getRight())
       )
    }
    addQueryTask(
        out,
        "task_instance.csv",
        String.format("select * from task_instance %s;", queryFilter)
```

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/annotations/RespectsArgumentQueryLogDays.java
vladislav-sidorovich: yes

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
vladislav-sidorovich: There is next line:
```
+ "Can be combined with --"
                  + OPT_START_DATE)
```

vladislav-sidorovich: the description updated to be more clear

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
vladislav-sidorovich: Yes, thank you. fixed.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
vladislav-sidorovich: I would say it should usage experience and not the ticket. once we will have some feedbacks, we will "think about it"

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
vladislav-sidorovich:
```
String filter =
        dateRange != null
            ? String.format(
                " where end_date >= CAST( '%s' as TIMESTAMP) and end_date < CAST( '%s' as TIMESTAMP) ;",
                dateToSqlFormat(dateRange.getLeft()), dateToSqlFormat(dateRange.getRight()))
            : "";
addQueryTask(out, "task_instance.csv", "select * from task_instance " + filter);
```
vs
```
    addQueryTask(
        out,
        "task_instance.csv",
        "select * from task_instance "
            + (dateRange != null
                ? String.format(
                    " where end_date >= CAST( '%s' as TIMESTAMP) and end_date < CAST( '%s' as TIMESTAMP) ;",
                    dateToSqlFormat(dateRange.getLeft()), dateToSqlFormat(dateRange.getRight()))
                : ""));

```
vs    
```
  addQueryTask(
        out,
        "task_instance.csv",
        dateRange == null
            ? "select * from task_instance;"
            : "select * from task_instance "
                + String.format(
                    " where end_date >= CAST( '%s' as TIMESTAMP) and end_date < CAST( '%s' as TIMESTAMP) ;",
                    dateToSqlFormat(dateRange.getLeft()), dateToSqlFormat(dateRange.getRight())));
```

Not a big difference from my point of view. The variable looks good, but it will require to introduce such variable for  several method or use code blocks `{ ... }`.  

So, I will keep it as is.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
zaldis: We can track it in the ticket for visibility and close it if it will not be relevant in the future.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
zaldis: The main difference that you can log your filter or set a breakpoint for debugging purposes. With current implementation you have to do it in `addQueryTask` which is called multiple times. Hence it will be harder to identify required call.

vladislav-sidorovich:
> -start-date=x --lookback-days=y -> date range: [x-y, x)
> 
> This usage is not intuitive, at least for me. Usually when I provide start date I want all the data to be later than this. I think `end-date` works better with a `lookback-days` because lookback refers to searching N days in the past.

I do the calculation and log the actual time range to show the date range explicitly. I agree that `start-date` might be confusing. However the logic behind it is the following:  I chose `X` date in the calendar and go back for `Y` days, as the result it will be a date range.  In other words, there is an ability to move `now`. 
With such logic go back from `end` is confusing as well. 

I'm open to ideas.

There is a good example: https://grafana.com/docs/grafana/latest/dashboards/use-dashboards/#set-dashboard-time-range  
For example we can rename `start` -> `from` and `end` - `to`. So, it will be: `from` date `X` look back for `Y` days.

does it look better?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
vladislav-sidorovich: added

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
kpiotrowski: If there will be a need to split it into multiple smaller, interval based files I'd consider ingesting it as a time-series data. We have a process for it. Airflow connector could be use by both metadata and time-series loaders.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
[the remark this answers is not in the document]
kajgol: I'm starting to wonder if we need this case at all. I can see a usecase for --lookback=7 days from now, or --start-date=last tuesday because that's when I run the previous dump, or --start-date=last tuesday and --end-date=last-saturday because I want a specific range, but why would I want something like 7 days before last tuesday?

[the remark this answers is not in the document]
vladislav-sidorovich: 1. Can be connector used without this case? - yes, start/end is enough to cover all ranges.
2. is there additional benefit for such use case? - yes, it's simpler way to specify the range.
3. is the benefit is significant? - it's hard to say/measure. 

> --start-date=last tuesday because that's when I run the previous dump

But you actually raised an interesting use case. What if I will run the dumper 3 times and I want cover the ranges [t1 - t2 - t3 - t4]? It doesn't look like a use case, we just need to run dumper once for [t1 - t4] where t4 can be now. 
We can go forward and think about an `incremental assessment`, run dumper each day/week and append data to the existing assessment... It sounds interesting, but it's not the case for now.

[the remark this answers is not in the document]
kajgol: "is there additional benefit for such use case? - yes, it's simpler way to specify the range."
It's not simpler if it's confusing ;-)

shevek-google (commented):
My thoughts from discussion:
* Start should temporally be before end.
* Can we make the names sufficiently generic that we can have one set of arguments for query logs, airflow logs, etc, in the future?
* lookback implies with respect to end, so it doesn't "feel" right to say "start" and lookback (because it would be forwards from start). Can we use interval, period, something like that which is agnostic about which endpoint we anchor?

Overall, I appreciate the philosophy that is driving these clarifications, so thank you for asking.

vladislav-sidorovich:
I will keep only `start` and `end` date parameters for now. `lookback-days` or similar parameter will be introduced in a separate PR.
Eventually these params should be applied to all connectors (metadata and query logs).

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/airflow/AirflowConnector.java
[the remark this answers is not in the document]
vladislav-sidorovich: I remove `lookback` param for now

PR #771, opened 2025-03-04
Feat: b/397181378 Encode PII values

zaldis:
- New CLI flag encode-pii
- Encode user field for YARN applications

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
@@ -365,12 +365,8 @@ public class ConnectorArguments extends DefaultArguments {
           .ofType(File.class)
           .describedAs("script.sql");
 
-  private final OptionSpec<Boolean> optionEncodePII =
-      parser
-          .accepts(OPT_ENCODE_PII, "Encode PII values before storing data into zip file")
-          .withOptionalArg()
-          .ofType(Boolean.class)
-          .defaultsTo(false);
+  private final OptionSpec<Void> optionEncodePII =
+      parser.accepts(OPT_ENCODE_PII, "Encode PII values before storing data into zip file");
 
   // redshift.
   private final OptionSpec<String> optionRedshiftIAMAccessKeyID =
@@ -987,7 +983,7 @@ public class ConnectorArguments extends DefaultArguments {
 
   @Nonnull
   public Boolean isEncodedPII() {
-    return getOptions().valueOf(optionEncodePII);
+    return getOptions().has(optionEncodePII);
   }
 
   public boolean isTestFlag(char c) {
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
@@ -365,7 +365,7 @@ public class ConnectorArguments extends DefaultArguments {
           .ofType(File.class)
           .describedAs("script.sql");
 
-  private final OptionSpec<Void> optionEncodePII =
+  private final OptionSpec<Void> optionEncodePii =
       parser.accepts(OPT_ENCODE_PII, "Encode PII values before storing data into zip file");
 
   // redshift.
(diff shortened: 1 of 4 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -976,6 +981,11 @@ public List<String> getQueryLogAlternates() {
     return getOptions().valuesOf(optionQueryLogAlternates);
   }
 
+  @Nonnull
+  public Boolean isEncodedPII() {
```
misolt, line 985: Here and in the other public method: I'd rename `PII` -> `Pii`.

In tests I can see that camelcase is already used, in private members and local variables it doesn't really matter.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/dto/ApiYARNApplicationDTO.java
```diff
@@ -27,6 +27,12 @@
  */
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class ApiYARNApplicationDTO {
+  private boolean isPIIEncodingEnabled = false;
```
misolt, line 30: Are there any reasons why we would not want to encode PII? Having only one code path (PII always encoded) seems like it would be easier to maintain.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/dto/ApiYARNApplicationDTO.java
```diff
@@ -88,6 +94,9 @@ public String getEndTime() {
   }
 
   public String getUser() {
+    if (isPIIEncodingEnabled) {
+      return String.valueOf(user.hashCode());
```
misolt, line 98: Note that hashCode can have collisions, which would cause different users to be merged together.

misolt: Other things we should probably consider for the encoding function are
- should it be deterministic?
- should it be reversible

zaldis: I think it should be not be a huge deal for this specific case, as we don't expect to have critical information in user field.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/dto/ApiYARNApplicationDTO.java
zaldis: Actually it makes sense to always encode PII, but we can't be sure for 100% that user field contains PII (it might be the OS user which executed the job). That's why it's up to customer to decide if it's required to encode such fields.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
zaldis: I'd create a separated ticket to fix it for all public methods. For now there are public methods which follow upper camel case format.
https://github.com/google/dwh-migration-tools/pull/771/files/e1cc7958a6db776c094b64fc95eb4d62ed43cf51#diff-f059b8c4a28e0cedc00705158b7b7fb830587335cc81c00856f51b4606111a3aR1003

For example:
- `public String getIAMAccessKeyID()`
- `public String getOracleSID()`

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -363,6 +365,9 @@ public class ConnectorArguments extends DefaultArguments {
           .ofType(File.class)
           .describedAs("script.sql");
 
+  private final OptionSpec<Void> optionEncodePii =
+      parser.accepts(OPT_ENCODE_PII, "Encode PII values before storing data into zip file");
```
vladislav-sidorovich, line 369: I prefer `mask` to `encode`.

vladislav-sidorovich:
Let's put this PR on hold and prepare a design documents because PII is very sensitive topic and there are laws regulations which we must to respect.

PR #776, opened 2025-03-05
[b/400931733] always add trailing slash

vladislav-sidorovich:


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerHandle.java
```diff
@@ -41,10 +42,28 @@ public ClouderaManagerHandle(URI apiURI, CloseableHttpClient httpClient) {
...
 
+  /** 1. Remove query params and url fragments 2. Add trailing slash for safety */
+  private static URI unify(URI uri) {
+    try {
+      return new URI(
+              uri.getScheme(),
+              uri.getUserInfo(),
+              uri.getHost(),
+              uri.getPort(),
+              uri.getPath() + "/",
```
sayuzbas-google, line 58: so `getPath` returns path without trailing slash, right? So you will not situation with 2 slashes in the end of uri?

sayuzbas-google (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerHandle.java
```diff
@@ -41,10 +42,28 @@ public ClouderaManagerHandle(URI apiURI, CloseableHttpClient httpClient) {
...
     Preconditions.checkNotNull(httpClient, "httpClient can't be null.");
 
-    this.apiURI = apiURI;
+    // Always add trailing slash for safety
+    this.apiURI = unify(apiURI);
     this.httpClient = httpClient;
   }
 
+  /** 1. Remove query params and url fragments 2. Add trailing slash for safety */
+  private static URI unify(URI uri) {
```
zaldis, line 51: Should not this method be moved outside of Cloudera Handler. For example to `utils` package.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerHandle.java
vladislav-sidorovich: `getPath` return what was provided in the `path`, it can be no slashes, can be 1 or multiple slashes in the end and in the mid of the path. It handled by `normalize()` method.

zaldis: This situation might be fixed with `normalize()` call at the end.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/cloudera/manager/ClouderaManagerHandle.java
vladislav-sidorovich: No, removing params and fragments in unique for our use case with Cloudera. If we will find more usage in the future we will extract it. For now it's Cloudera internal invariant that we can rely on in Cloudera tasks.

PR #777, opened 2025-03-05
[b/377253441] Add extraction of WAREHOUSE_EVENTS_HISTORY

misolt:


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakePlanner.java
```diff
@@ -167,6 +167,38 @@ Task<?> reportDateRangeTask() {
...
+            "cluster_number",
+            "event_name",
+            "event_reason",
+            "event_state",
+            "user_name",
+            "role_name",
+            "query_id");
+    String query =
+        String.format(
+            "SELECT %s FROM SNOWFLAKE.ACCOUNT_USAGE.WAREHOUSE_EVENTS_HISTORY",
```
vladislav-sidorovich, line 185: Why we can't use `select * ...` here? In general it's a good practice for the Dumper to reduce coupling on the details.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakePlanner.java
```diff
@@ -167,6 +167,38 @@ Task<?> reportDateRangeTask() {
     return new LiteTimeSeriesTask("report_date_range.csv", query, header);
   }
 
+  Task<?> warehouseEventsHistoryTask() {
+    ImmutableList<String> selectList =
```
sayuzbas-google, line 171: nit: columnsList

vladislav-sidorovich:
looks good, but I'm afraid that no tests were updated with the code changes.

vladislav-sidorovich (approved):

sayuzbas-google (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakePlanner.java
```diff
@@ -167,6 +167,38 @@ Task<?> reportDateRangeTask() {
...
+            "WarehouseId",
+            "WarehouseName",
+            "ClusterNumber",
+            "EventName",
+            "EventReason",
+            "EventState",
+            "UserName",
+            "RoleName",
+            "QueryId");
+    return new LiteTimeSeriesTask("warehouse_events_history.csv", query, header);
```
sayuzbas-google, line 199: Should not we reduce number of dumped data as part of Lite assessment? I remember you mentioned that, data which you looked is not so big, just want to double check, so we will not have large data here.
So maybe worth to reduce dumping data to some date range, like the last 10 days or 2 weeks?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakePlanner.java
misolt: I think both are fine - "select list" is commonly used to refer to these columns across different dialects, e.g. [here](https://docs.snowflake.com/en/sql-reference/functions/listagg#:~:text=column%20in%20the-,SELECT%20list,-.%20Therefore%2C%20do%20not)

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakePlanner.java
misolt: Let's discuss in chat

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakePlanner.java
misolt: It is a good practice, but here it will be coupled anyway, because we need to name the select list items explicitly in subsequent steps.

vladislav-sidorovich: Do you mean `LiteTimeSeriesTask`? So, why do we need it explicitly there?

misolt:
> looks good, but I'm afraid that no tests were updated with the code changes.

This is a common problem with changes of this type.

I could test the Java parts of the code, but most complexity is in SQL, so it wouldn't lead to interesting tests. I mean for example testing if the item that we add to the list actually ends up being added to the list, but without testing if the item works.

Automatically testing SQLs would address it, but this is more complex to implement. We've discussed it before and didn't come up with a satisfactory solution.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakePlanner.java
misolt: I mean the format of the CSVs in the output.

PR #783, opened 2025-03-11
[b/394577286] Replace the warehouse events task with a lite version

misolt:


vladislav-sidorovich (approved):

vladislav-sidorovich:
the changes are minimal, however I'm wondering that there are no failed tests. 
We have to keep attention on it to avoid unexpected issues in the future

PR #787, opened 2025-03-14
Remove redundant recursion for TaskGroup

misolt:
The methods `countTasks` and `print` are recursive, but the preconditions checks in `ParallelTaskGroup::addTask` mean that they never reach depth >1, making the recursion not that useful.

Since the same effect can be accomplished using just `if` or `:?`, I removed the recursion.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/MetadataDumper.java
@@ -91,11 +91,12 @@ public class MetadataDumper {
     return run(connector, arguments);
   }
 
-  private void print(@Nonnull Task<?> task, int indent) {
-    System.out.println(repeat(' ', indent * 2) + task);
+  private void print(@Nonnull Task<?> task) {
+    String indent = "  ";
+    System.out.println(indent + task);
     if (task instanceof TaskGroup) {
       for (Task<?> subtask : ((TaskGroup) task).getTasks()) {
-        print(subtask, indent + 1);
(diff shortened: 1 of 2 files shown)


vladislav-sidorovich:
I'm not sure we should go this direction. 
1. implementation with recursion looks good and works correct.
2. with the recursion there are no coupling between different parts of the code. 

If we really want to do something like that we should change the api and make it `public` contract that only 1 level structure is supported. 
For example `public void addTask(Task<?> task)` -> `public void addTask(Task<? extend MyNewParraliasableTask> task)`

misolt:
> If we really want to do something like that we should change the api and make it `public` contract that only 1 level structure is supported. For example `public void addTask(Task<?> task)` -> `public void addTask(Task<? extend MyNewParraliasableTask> task)`

From the [preconditions check](https://github.com/google/dwh-migration-tools/blob/8745fd06720206fd48074edd7f0de47afabde9e7/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/ParallelTaskGroup.java#L40) I understand that this is already done. It can technically be circumvented, but the main goal of `TaskGroup` seems to be running multiple things at once in the same time slot, so a nested subtask task would not bring any benefit vs a top-level subtask in the same `TaskGroup`.

I'm not sure about adding a parallelizable abstract class or interface, seems difficult to reconcile it with the current hierarchy. For example, should `AbstractJdbcTask` extend it or not?

misolt:
Closing, because:

- we need to address b/404528097 before this can be implemented
- there's no need to keep the code change until then - it's trivial and can be written again

PR #797, opened 2025-03-18
[b/404164828] Limit time series data to last 14 days

misolt:


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakePlanner.java
```diff
@@ -55,6 +55,9 @@ private enum Format {
     }
   }
 
+  private static final String TIME_PREDICATE =
+      "timestamp > CURRENT_TIMESTAMP(0) - INTERVAL '14 days'";
```
vladislav-sidorovich, line 59: Should it be configurable by dumper parameters? Let's say 14d is default, but users will be able to override it and set 7, 21, 30 and so on.

vladislav-sidorovich (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakePlanner.java
```diff
@@ -210,27 +224,21 @@ Task<?> warehouseEventsHistoryTask() {
...
             "credits_used_cloud_services");
+    String view = "SNOWFLAKE.ACCOUNT_USAGE.WAREHOUSE_METERING_HISTORY";
     String query =
-        String.format(
-            "SELECT %s FROM SNOWFLAKE.ACCOUNT_USAGE.WAREHOUSE_METERING_HISTORY",
-            String.join(", ", selectList));
+        buildQuery(
+            selectList,
+            view,
+            "end_time IS NULL OR end_time > CURRENT_TIMESTAMP(0) - INTERVAL '14 days'");
```
sayuzbas-google, line 241: should not we filter by `start_time`? so to get events which started in the last 14 days, not finished during the last 14 days

misolt: At the moment both seem equally good, but if we use `end_time`, we can still switch to `start_time` by adding filtering in the service.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakePlanner.java
misolt: It might be an useful addition later, but at the current stage choosing the appropriate default seems good enough.

The main factor for this number of days is if it's enough data for Assessment to be accurate. This means that the user won't have the information to choose the value correctly.

PR #798, opened 2025-03-18
[b/404528097] Remove TaskGroup#getTasks

misolt:
As the first step of fixing the issue, remove the method `TaskGroup#getTasks`. This method returns the reference to all subtasks as a mutable list, allowing for arbitrary modification.

Since all used instances of `TaskGroup` belong to the subclass `ParallelTaskGroup`, I also made `TaskGroup` abstract.

Two additional optimizations are possible. **They were not included** to avoid unnecessary complexity:
1. Return an `Iterable` from `TaskGroup#toCallables` to avoid the cost of creating the list
2. Use a one-way linked list to avoid the concatenation in `print`

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/TasksRunner.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/TasksRunner.java
@@ -65,7 +65,7 @@ public class TasksRunner implements TaskRunContextOps {
     context = createContext(sinkFactory, handle, threadPoolSize, arguments);
     this.state = state;
     this.tasks = tasks;
-    totalNumberOfTasks = countTasks(tasks);
+    totalNumberOfTasks = TaskGroup.count(tasks);
     stopwatch = Stopwatch.createStarted();
     numberOfCompletedTasks = new AtomicInteger();
   }
@@ -170,11 +170,5 @@ public class TasksRunner implements TaskRunContextOps {
       }
     }
     return null;
-  }
-
-  private int countTasks(List<Task<?>> tasks) {
-    return tasks.stream()
-        .mapToInt(task -> task instanceof TaskGroup ? countTasks(((TaskGroup) task).getTasks()) : 1)
-        .sum();
   }
 }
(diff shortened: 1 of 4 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/TasksRunner.java
```diff
@@ -65,7 +65,7 @@ public TasksRunner(
     context = createContext(sinkFactory, handle, threadPoolSize, arguments);
     this.state = state;
     this.tasks = tasks;
-    totalNumberOfTasks = countTasks(tasks);
+    totalNumberOfTasks = TaskGroup.count(tasks);
```
vladislav-sidorovich, line 68: It's not obvious why `TaskRunner` should call `TaskGroup` instead of simple `Task` class to count the tasks.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskGroup.java
```diff
@@ -49,11 +63,6 @@ public void addTask(Task<?> task) {
     tasks.add(task);
   }
 
-  @Nonnull
-  public List<Task<?>> getTasks() {
```
vladislav-sidorovich, line 53: What if we will return `Iterable` or `ImmutableList` and do not modify the rest code?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/TasksRunner.java
misolt: I don't think `Task` would be right for it, because then `Task` would depend on `TaskGroup`. It's doable, just more complicated than necessary.

If you prefer it, I can make a `TaskUtil` class (unless there already is one) and move `count` there.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/task/TaskGroup.java
misolt: We don't need to expose so much internals to fill the purpose of this class. The method `toCallables` does something similar, but it's limited to subclasses, making it not as bad.

In general if this is public we would probably need to worry about contents getting out of date, because the internal task list can still be modified. I also don't like having to copy the list just to count the items.

PR #800, opened 2025-03-19
Add the Assessment flag to metadata

misolt:
The `--assessment` flag needs to be present in output archive metadata. It doesn't make sense to have the user provide this flag, because by design all usage of `snowflake-lite` is meant for Assessment.

Modify the `ConnectorArguments` passed to the metadata task in order to make the flag appear in the outputs as if the user provided it.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeLiteConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeLiteConnector.java
@@ -25,8 +25,10 @@ import com.google.edwmigration.dumper.application.dumper.task.DumpMetadataTask;
 import com.google.edwmigration.dumper.application.dumper.task.FormatTask;
 import com.google.edwmigration.dumper.application.dumper.task.Task;
 import com.google.edwmigration.dumper.application.dumper.utils.ArchiveNameUtil;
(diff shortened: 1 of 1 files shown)


vladislav-sidorovich:
It was a decision to keep `--assessment` for consistency in the connectors. The same thing is in Datalake related connectors. The flag is not needed, but exists for consistency. 

- [ ] discussed with the team to change approach for `--assessment` flag usage

misolt:
> It was a decision to keep `--assessment` for consistency in the connectors. The same thing is in Datalake related connectors. The flag is not needed, but exists for consistency.
> 
> * [ ]  discussed with the team to change approach for `--assessment` flag usage

There are two separate points:
1. If the user provides the `--assessment` flag, will extraction succeed?
2. Will it succeed (and produce valid outputs) if the flag is not provided?

The point about consistency applies in 1. Hypothetically, if the user had started with one of the Datalake connectors (using the `--assessment` flag) and then wanted to try `snowflake-lite` with the same, they would get an unexpected error.

We can discuss it, but it is handled by the check:
```java
      String message = "...";
      throw new MetadataDumperUsageException(message);
    }
```
which is not a part of this PR.

This PR is about the second case - skipping the flag shouldn't cause the output to become invalid for `snowflake-lite`. It also doesn't have to cause extraction to fail.

Generally, there's no scenario where the user would want Dumper to fail, if it has all the necessary data to succeed.

PR #804, opened 2025-03-20
[improvement] add test template to verify assessment flag usage

vladislav-sidorovich:
It's very simple forget to add '--assessment' flag to connector. The problem is that the mistake will be find only during uploading zip to server.

So, the idea is to explicitly mark each connector if `assessment` flag is required or not for the connector.   In addition to that the test will verify that connectors are aligned with the annotations.

Please help to double check if the connectors marked properly by `RespectsArgumentAssessment` or `AvoidArgumentAssessment`. 

It should be at least align with our public doc: https://cloud.google.com/bigquery/docs/migration-assessment#teradata_1

Note: the test is failed because the validation logic doesn't exist for all the connector. I would prefer first define the list of connectors and add the validation logic later.

### DIFF
--- a/dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/AssessmentFlagUsageTest.java
+++ b/dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/AssessmentFlagUsageTest.java
@@ -1,0 +1,30 @@
+package com.google.edwmigration.dumper.application.dumper;
+
+import static org.junit.Assert.assertThrows;
+
+import com.google.common.collect.ImmutableCollection;
+import com.google.edwmigration.dumper.application.dumper.annotations.RespectsArgumentAssessment;
+import com.google.edwmigration.dumper.application.dumper.connector.Connector;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class AssessmentFlagUsageTest {
+  private static final Logger logger = LoggerFactory.getLogger(AssessmentFlagUsageTest.class);
+
+  @Test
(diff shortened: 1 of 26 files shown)


dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/AssessmentFlagUsageTest.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class AssessmentFlagUsageTest {
+  private static final Logger logger = LoggerFactory.getLogger(AssessmentFlagUsageTest.class);
+
+  @Test
+  public void assessmentFlagIsExplicitlySpecified() throws Exception {
+    ImmutableCollection<Connector> connectors = ConnectorRepository.getInstance().getAllConnectors();
+    for (Connector connector : connectors) {
+      RespectsArgumentAssessment assessment = AnnotationUtils.findAnnotation(connector.getClass(),
+          RespectsArgumentAssessment.class);
```
misolt, line 21: Many connectors don't have this annotation, despite using the `--assessment` flag and they will not be found by this test.

kpiotrowski:
What about the connectors that are shared by assessment and translation? Depending on the use case they will need `--assessment` flag or not. This is true for most of the "metadata" connectors supported by the assessment. Additionally, some query logs connectors are shared by assessment and lineage.

vladislav-sidorovich:
> What about the connectors that are shared by assessment and translation? Depending on the use case they will need `--assessment` flag or not. This is true for most of the "metadata" connectors supported by the assessment. Additionally, some query logs connectors are shared by assessment and lineage.

In this case such connectors should be able accept and process the flag. But I think we can go case by case.

@shevek-google what do you think?

PR #806, opened 2025-03-20
added gradle google profile

kaxuna:


buildSrc/src/main/groovy/dwh-migration-dumper.java-common-conventions.gradle
```diff
@@ -20,10 +20,17 @@ plugins {
...
     id 'com.diffplug.spotless'
     id 'nebula.info'
+    id "com.google.cloud.artifactregistry.gradle-plugin"
 }
 
 repositories {
-    mavenCentral()
+    if (project.hasProperty('google')) {
+        maven {
+            url "artifactregistry://us-maven.pkg.dev/artifact-foundry-prod/maven-3p-trusted"
```
vladislav-sidorovich, line 29: Well, I think it should be hidden.

- [ ] hide the url, not share it with the all the world

PR #746, opened 2025-02-19
[b/394312229] Dump Snowflake external table metadata

Hackerpilot:

PR #812, opened 2025-03-24
Revert "[b/394312229] Dump Snowflake external table metadata"

misolt:
Reverts google/dwh-migration-tools#746


misolt:
Continued in #813

PR #829, opened 2025-03-26
Small updates on gcsync client

tonyxiaowei:
1. Use bytes instead of string for the 256bit checksum which is more storage efficient.
2. Set a max 1hr timeout limit for the cloud run task which would be needed for larger files
3. Removed an unused import


dbsync/common/src/main/java/com/google/edwmigration/dbsync/common/ChecksumGenerator.java
```diff
@@ -48,7 +48,8 @@ public void generate(ChecksumConsumer<IOException> out, ByteSource in) throws IO
         // Yes, this masks the instance variable.
         int blockSize = rollingChecksum.getBlockSize();
         if (DEBUG) {
```
vladislav-sidorovich, line 50: Do we really need this flag? Looks like we can remove it from the code and just use logging with levels

dbsync/gcsync/src/main/java/com/google/edwmigration/dbsync/gcsync/Constants.java
```diff
@@ -20,4 +23,6 @@ public class Constants {
   public static final String GENERATE_CHECK_SUM_MAIN = GenerateCheckSumMain.class.getName();
 
   public static final String RECONSTRUCT_FILE_MAIN = ReconstructFilesMain.class.getName();
+
+  public static final Duration CLOUD_RUN_TIMEOUT = Durations.fromMinutes(60);
```
vladislav-sidorovich, line 27: [minor] It will be nice to have an ability to configure it somehow. optional parameres, env variables an so on

vladislav-sidorovich (approved):

dbsync/gcsync/src/main/java/com/google/edwmigration/dbsync/gcsync/Constants.java
tonyxiaowei: I have a following up change addressing this. Making it configurable

PR #830, opened 2025-03-27
[Snowflake Connector] Add tests for connector's SQL overrides.

paolomorandini:
- Add tests for SQL overrides

### DIFF
--- a/dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnectorTest.java
+++ b/dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnectorTest.java
@@ -16,6 +16,10 @@
  */
 package com.google.edwmigration.dumper.application.dumper.connector.snowflake;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.assertTrue;
+
 import com.google.common.collect.ImmutableMap;
 import com.google.common.io.Resources;
 import com.google.edwmigration.dumper.application.dumper.ConnectorArguments;
@@ -34,7 +38,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import javax.annotation.Nonnull;
-import org.junit.Assert;
+import org.apache.commons.lang3.ArrayUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.junit.Assume;
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -122,19 +127,19 @@ public class SnowflakeMetadataConnectorTest extends AbstractSnowflakeConnectorEx
     Assume.assumeTrue(isDumperTest());
 
     MetadataDumperUsageException exception =
-        Assert.assertThrows(
+        assertThrows(
             MetadataDumperUsageException.class,
             () -> {
               File outputFile =
                   TestUtils.newOutputFile("compilerworks-snowflake-metadata-fail.zip");
               String[] args = ARGS(connector, outputFile);
 
-              Assert.assertEquals("--database", args[6]);
+              assertEquals("--database", args[6]);
               args[7] = args[7] + "_NOT_EXISTS";
               run(args);
             });
 
-    Assert.assertTrue(exception.getMessage().startsWith("Database name not found"));
+    assertTrue(exception.getMessage().startsWith("Database name not found"));
   }
 
   @Test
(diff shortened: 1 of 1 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeTaskUtil.java
```diff
@@ -24,21 +24,25 @@
...
+  private static final String EMPTY_WHERE_CLAUSE = "";
+
   static AbstractJdbcTask<Summary> withFilter(
       String format,
       String schemaName,
       String zipEntryName,
       String whereClause,
       Class<? extends Enum<?>> header) {
+    if (format.contains(" WHERE ")) {
+      whereClause = EMPTY_WHERE_CLAUSE;
```
vladislav-sidorovich, line 36: Will it construct a valid sql? Do we have test for it? I see the files with sql below, but which line from those should verify this `if` statement is unclear. 

I'm interested how the empty string will be used in sql query. 
`select * from table where 'EMPTY_WHERE_CLAUSE' order by 1;` is not a valid sql.

paolomorandini: This only works because we have full control of the SQL statements and we know there is no SQL containing clauses after `WHERE` and that `format` will only contain a `WHERE` clause if that is specified with an override. In case of `query` override this code is not executed (i.e. the query is used literally).

This fix would've been more evident if this PR came in two steps, as without this change the queries based on `ACCOUNT_USAGE` would all end in `WHERE DELETES IS NOT NULL WHERE SQL_OVERRIDE(...)`.

The purpose of this PR is to set a test to support refactoring of the query building methods.

vladislav-sidorovich: Can we maybe write a test that will fail if some sql queries with clauses after `WHERE` will appear in the future? 

I'm worried about maintainability. This statement might be confusing for other (including external) developers without additional explanation as in the comment above.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeTaskUtil.java
```diff
@@ -24,21 +24,25 @@
...
 
+  private static final String EMPTY_WHERE_CLAUSE = "";
+
   static AbstractJdbcTask<Summary> withFilter(
       String format,
       String schemaName,
       String zipEntryName,
       String whereClause,
       Class<? extends Enum<?>> header) {
+    if (format.contains(" WHERE ")) {
```
misolt, line 35: I think we should handle these `where` clauses in one place, deducing it from the format string like this can be unpredictable.

For example, someone might want to get rid of a bad `where` clause and remove it from the format string, only to discover that this leads to another `where` clause being added, in this method.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeTaskUtil.java
paolomorandini: I updated the PR to not introduce fixes and only add a test to assert the SQL statements produced by the Snowflake connector. Fixes for illegal SQL will be included in the next PR.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnectorTest.java
```diff
@@ -154,10 +156,93 @@ public void connector_generatesExpectedSql() throws IOException {
...
+  public void connector_assessment_generatesExpectedSql() throws IOException {
+    Map<String, String> actualSqls = collectSqlStatements("--assessment");
+    TaskSqlMap expectedSqls =
+        CoreMetadataDumpFormat.MAPPER.readValue(
+            Resources.toString(
+                Resources.getResource("connector/snowflake/assessment-jdbc-tasks-sql.yaml"),
+                StandardCharsets.UTF_8),
+            TaskSqlMap.class);
+
+    Assert.assertEquals(expectedSqls.size(), actualSqls.size());
```
misolt, line 169: This is usually clearer with a static import: `Assert.assertEquals` -> `assertEquals`

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnectorTest.java
```diff
@@ -154,10 +156,93 @@ public void connector_generatesExpectedSql() throws IOException {
...
+              Resources.toString(
+                  Resources.getResource(
+                      String.format(
+                          "connector/snowflake/jdbc-tasks-sql-override-%s.yaml",
+                          propertyAction.value)),
+                  StandardCharsets.UTF_8),
+              TaskSqlMap.class);
+
+      List<String> overrideArguments = new ArrayList<>();
+      for (MetadataView metadataView : MetadataView.values()) {
```
misolt, line 190: There should be no non-trivial logic in tests, which usually includes the loops. Please consider the alternatives:
- Allow for repetition and list all items one-by-one.
- Make the test parameterized, only do a single step per test run.
- Only run the test for a smaller subset of values instead of all values. Most errors can be detected by checking only a subset of the values and this will make the test more readable overall.

dumper/app/src/test/resources/connector/snowflake/assessment-jdbc-tasks-sql-override-query.yaml
```diff
@@ -0,0 +1,10 @@
+databases-au.csv: "SQL_OVERRIDE(databases, query)"
```
misolt, line 1: If the input file is very small, I think it would be better to include it as a literal in the tests, to minimize the need for the reader to switch between the two files.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnector.java
```diff
@@ -50,7 +50,7 @@ public class SnowflakeMetadataConnector extends AbstractSnowflakeConnector
   @SuppressWarnings("UnusedVariable")
   private static final Logger logger = LoggerFactory.getLogger(SnowflakeMetadataConnector.class);
 
-  private enum PropertyAction {
+  enum PropertyAction {
```
misolt, line 53: Instead of making this enum visible in the tests, I'd prefer to list the values explicitly in tests.

The enum's existence and properties are an implementation detail, changing the enum should have no effect on the tests as long as the externally visible API of this class didn't change.

dumper/app/src/test/resources/connector/snowflake/assessment-jdbc-tasks-sql-override-query.yaml
paolomorandini: Good point, done.

misolt (approved):

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnectorTest.java
```diff
@@ -34,6 +38,8 @@
 import java.util.List;
 import java.util.Map;
 import javax.annotation.Nonnull;
+import org.apache.commons.lang3.ArrayUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.junit.Assert;
```
misolt, line 43: This import should be removed, since the static imports replaced it.

PR #832, opened 2025-03-28
More changes for gcsync

tonyxiaowei:
1. Make the timeout of cloud run tasks configurable.
2. Fix the JobClient settings so that it will poll up to task_run_timeout + 5 mins.
3. Add a README.md for dbsync
4. Add trailing slash if client missed specifying it
5. Used buffered stream in copyTo method.


dbsync/README.md
```diff
@@ -0,0 +1,31 @@
...
+`Java 8` or higher. To check Java version run the command`java -version` or
+refer to Java vendor documentation.
+
+## Gcysnc
+
+This tool is in preview mode, it's being actively developed.
+
+### Build the Gcsync Client ###
+
+    ./gradlew :dbsync:gcsync:build
```
vladislav-sidorovich, line 17: The command is failed:
```
    com.google.protobuf.RuntimeVersion.validateProtobufGencodeVersion(
                       ^
  symbol:   class RuntimeVersion
  location: package com.google.protobuf
/Users/vsidorovich/github/dwh-migration-tools/dbsync/common/build/generated/source/proto/main/java/com/google/edwmigration/dbsync/proto/Dbsync.java:83: error: cannot find symbol
    descriptor.resolveAllFeaturesImmutable();
              ^
  symbol:   method resolveAllFeaturesImmutable()
  location: variable descriptor of type FileDescriptor
12 errors

FAILURE: Build failed with an exception.
```

I assume protibuf compiler is needed in the system. Could you please describe it here?

tonyxiaowei: I was able to reproduce the error by removing protobuf-compipler but I'm seeing a slightly different error message. 

Can you try running

```sudo apt-get install protobuf-compiler```

and see if it solves the problem? If so, I'll add it to the documentation.

vladislav-sidorovich (approved):

dbsync/README.md
tonyxiaowei: Updated readme, looks like it's only an issue on macOS which isn't a priority for this project

PR #840, opened 2025-04-01
[b/292044349] Fix dumping xids bigger than 2^63-1

misolt:


sayuzbas-google (commented):
How this change applies PR name? which one are xids bigger than 2^63 - 1?

misolt:
> How this change applies PR name? which one are xids bigger than 2^63 - 1?

It fixes the error with xids. The default for xid is to cast it to a 64-bit integer. This works most of the time, but fails for big xids (original issue author verified it to be 2^63-1 which makes sense, but the actual number is not important).

To prevent the errors, cast xids to VARCHAR so that they become strings instead of fixed-size integers.

sayuzbas-google (approved):

PR #842, opened 2025-04-01
gradle google profile 

kaxuna:


vladislav-sidorovich (approved):
This version is good. thank you

PR #843, opened 2025-04-02
[b/404497459] Remove an integration test from unit tests

misolt:
`HiveMetadataConnectorTest#testLoadedhive312` is an integration test in the Dumper's unit test suite.

The test works by running the Dumper 5 times against in-memory metastore, then checking that no exceptions are thrown and the results are not empty.

- This test doesn't test much of the application logic, it tests integration instead.
- It adds about 5 minutes to every run of the tests.
- This test is run conditionally depending on the developer's system, making the tests less reliable overall.


vladislav-sidorovich:
Well, the removing is an option for sure. 

But did we expose other options? What is a root cause to running for minutes, why it's not 1-2 seconds (what is really more than enough time)

misolt:
> Well, the removing is an option for sure.
> 
> But did we expose other options? What is a root cause to running for minutes, why it's not 1-2 seconds (what is really more than enough time)

I investigated it, of the 5 minutes:
- It takes abt. 2 to run the Dumper 5 times
- It takes almost 3 minutes to set up the test, i.e. start the Hive instance

I think 3 minutes is a reasonable amount of time for an integration test, but far too long to be run as part of the project's unit tests, especially if integration tests exist to test the same functionality.

vladislav-sidorovich (approved):
As discussed we have it covered with the integration tests internally at Google.

PR #847, opened 2025-04-03
[b/404819770] Enable snowflake MFA caching

kpiotrowski:
This PR enables an optional snowflake MFA caching mechanism. This simplifies the usage of the dumper for the users who decide to use the user&password auth flow with the MFA. This allows user to approve MFA request only once instead of receiving the push notification on every single SQL query.

This requires setting the `authenticator` property to `username_password_mfa` and adding JNA classes to the classpath. Changing the `authenticator` property does not affect the authentication if MFA is disabled for the user.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/AbstractSnowflakeConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/AbstractSnowflakeConnector.java
@@ -117,8 +117,12 @@ public abstract class AbstractSnowflakeConnector extends AbstractJdbcConnector {
       throws SQLException {
     Driver driver =
         newDriver(arguments.getDriverPaths(), "net.snowflake.client.jdbc.SnowflakeDriver");
-    String password = arguments.getPasswordOrPrompt();
-    return new SimpleDriverDataSource(driver, url, arguments.getUser(), password);
+    Properties prop = new Properties();
+
+    prop.put("user", arguments.getUser());
+    prop.put("password", arguments.getPasswordOrPrompt());
+    prop.put("authenticator", "username_password_mfa");
+    return new SimpleDriverDataSource(driver, url, prop);
   }
 
   private DataSource createPrivateKeyDataSource(@Nonnull ConnectorArguments arguments, String url)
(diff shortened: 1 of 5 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/AbstractSnowflakeConnector.java
```diff
@@ -117,8 +117,12 @@ private DataSource createUserPasswordDataSource(@Nonnull ConnectorArguments argu
       throws SQLException {
     Driver driver =
         newDriver(arguments.getDriverPaths(), "net.snowflake.client.jdbc.SnowflakeDriver");
-    String password = arguments.getPasswordOrPrompt();
-    return new SimpleDriverDataSource(driver, url, arguments.getUser(), password);
+    Properties prop = new Properties();
+
+    prop.put("user", arguments.getUser());
+    prop.put("password", arguments.getPasswordOrPrompt());
+    prop.put("authenticator", "username_password_mfa");
```
vladislav-sidorovich, line 124: Can we add some test? 

I'm wondering about maintaining.  Let's say in the future some one will just delete `runtimeOnly libs.jna` because it's not used explicitly. How can we prevent this situation?

misolt (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/AbstractSnowflakeConnector.java
kpiotrowski: Good idea, thanks! I added a simple unit test and a comment.

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/AbstractSnowflakeConnectorTest.java
```diff
@@ -105,4 +105,13 @@ public void openConnection_failsForMixedPrivateKeyAndPassword() throws IOExcepti
             .contains(
                 "Private key authentication method can't be used together with user password"));
   }
+
+  @Test
+  public void checkJnaInClasspath_success() {
+    try {
+      Class.forName("com.sun.jna.Library");
+    } catch (ClassNotFoundException e) {
+      Assert.fail("net.java.dev.jna was not found in the classpath.");
```
vladislav-sidorovich, line 114: Please add the motivation for this lib. Something like `required for Snowflake mfa`

vladislav-sidorovich (approved):

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/AbstractSnowflakeConnectorTest.java
kpiotrowski: done, I added comment and note in a message.

PR #855, opened 2025-04-07
Fix: type errors related to xids in Redshift

misolt:
b/292044349


vladislav-sidorovich:
Can we somehow say the driver to use `BigInterger` instead of `chars`?

misolt:
> Can we somehow say the driver to use `BigInterger` instead of `chars`?

We probably could, but there's no real use for it. We don't do any math with these IDs, only write them to a csv file.

vladislav-sidorovich (approved):

PR #856, opened 2025-04-07
[OPS] fix banner load

vladislav-sidorovich:
Verified for:
1. Java 8, 11, 17, 21.
2. Single and multiple jars in class path.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/StartUpMetainformationPrinter.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/StartUpMetainformationPrinter.java
@@ -16,38 +16,74 @@
  */
 package com.google.edwmigration.dumper.application.dumper;
 
-import java.io.IOException;
-import java.net.URISyntaxException;
+import static java.util.jar.Attributes.Name.IMPLEMENTATION_TITLE;
+
+import java.io.BufferedReader;
+import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.net.URL;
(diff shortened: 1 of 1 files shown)


misolt (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/StartUpMetainformationPrinter.java
```diff
@@ -16,38 +16,74 @@
...
+    System.out.println("App version: [" + version + "], change: [" + change + "]");
+    System.out.println("Build date: " + buildDate);
+    System.out.println();
+  }
+
+  private static Manifest loadCurrentClassManifest() {
+    try {
+      final String implementationTitle =
+          StartUpMetainformationPrinter.class.getPackage().getImplementationTitle();
+      final ClassLoader classLoader = StartUpMetainformationPrinter.class.getClassLoader();
```
misolt, line 71: nit: usually final is unnecessary for local variables

vladislav-sidorovich: In the current implementation we don't need it. Sometimes I use `final` if code is not trivial and I want to highlight some main variables.

To me it makes the code more readable. What do you think?

vladislav-sidorovich: ok, let's keep it as is. Maybe we will change it in the future

PR #859, opened 2025-04-08
Gcsync SubDirectory Scanning and Upload Jar Fix

amitgoel248:
Updated the scan files logic to scan sub directories.
Also applied a regex filter on the files to scan.

Updated the jar upload logic to pick the jar from the build target folder instead of picking from the local directory.


dbsync/gcsync/src/main/java/com/google/edwmigration/dbsync/gcsync/GcsyncClient.java
```diff
@@ -97,56 +105,78 @@ public void syncFiles()
...
-          continue;
-        }
-        if (Files.size(path) < Constants.RSYNC_SIZE_THRESHOLD) {
-          // If file is small, we just upload it.
-          filesToUpload.add(path);
+  public void scanDirectory(final Path directory) throws IOException, URISyntaxException {
+    try (DirectoryStream<Path> directoryStream = Files.newDirectoryStream(directory)) {
+      for (Path entry : directoryStream) {
+        String fileName = entry.getFileName().toString();
+        logger.log(Level.INFO, "Found entry: {0}", entry);
```
tonyxiaowei, line 117: I think this is too verbose to log at INFO

dbsync/gcsync/src/main/java/com/google/edwmigration/dbsync/gcsync/GcsyncClient.java
```diff
@@ -97,56 +105,78 @@ public void syncFiles()
...
-    Path sourceJar = Paths.get(Constants.JAR_FILE_NAME);
-    if (sourceJar == null) {
-      throw new IllegalArgumentException("The jar file has been deleted");
+    Blob blob = gcsStorage.getBlob(tmpBucket, Constants.GCS_JAR_FILE_NAME);
+    InputStream inputStream = this.getClass().getClassLoader().getResourceAsStream(Constants.PUBLISHED_JAR_FILE_NAME);
+    if (inputStream == null) {
+      throw new IllegalArgumentException("The " + Constants.GCS_JAR_FILE_NAME + " file not found.");
+    }
+    // Create a temporary file to store the jar content
+    Path tempJarPath = Files.createTempFile("temp", ".jar");
```
tonyxiaowei, line 168: I don't understand the purpose of copying the jar and immediately upload it, why does it help? In case the client modifies the jar in this short span of time?

dbsync/gcsync/src/main/java/com/google/edwmigration/dbsync/gcsync/Constants.java
```diff
@@ -7,7 +7,9 @@ public class Constants {
 
   public static final String FILES_TO_RSYNC_FILE_NAME = "filesToRsync.txt";
 
-  public static final String JAR_FILE_NAME = "gcsync-all.jar";
+  public static final String GCS_JAR_FILE_NAME = "gcsync-all.jar";
+
+  public static final String PUBLISHED_JAR_FILE_NAME = "gcsync-1.0.0-all.jar";
```
tonyxiaowei, line 12: The jar generated is gcsync-call.jar using the tool in standalone. I don't think we should hard code the version here as it might change. I think we can try some changes to https://source.corp.google.com/h/team/cloud-dataengine-connectors-eng/mirroring-agent/+/rsync-integration:dbsync-lib/build.gradle to remove the version from the jar name.

Instead of having DTS starting a separate process, we should probably expose a method as putFiles() instead?

dbsync/gcsync/src/main/java/com/google/edwmigration/dbsync/gcsync/GcsyncClient.java
amitgoel248: Removed.

dbsync/gcsync/src/main/java/com/google/edwmigration/dbsync/gcsync/GcsyncClient.java
amitgoel248: Updated.

dbsync/gcsync/src/main/java/com/google/edwmigration/dbsync/gcsync/Constants.java
amitgoel248: I have update the overall logic a little bit. Basically the requirement was to embed the gcsync-all.jar with in the the mirroring agent jar, so that we do not have to provide 2 jars to the customers.
For that I have not updated the logic as follows.

Change the gcsync build file to push the jar to resources folder (it pushes all the jars present in the libs of gcsync package)
Introduces a local variable to define the jar name, so that DTS can pass in their own jar name.
Update the upload jar method to fetch the resource as stream, create a temp file where the agent is running, upload it to GCS and delete the file.

vladislav-sidorovich (approved):

PR #864, opened 2025-04-11
[OPS] Add some fun to a developer's life

vladislav-sidorovich:
Print banner during build


kajgol (approved):
Is there a PRD and design doc for this?

vladislav-sidorovich:
> Is there a PRD and design doc for this?

I'm afraid it will be too many comments that I will not be able to address. So, I use a trick, no design doc = no comment.

kpiotrowski (approved):

PR #872, opened 2025-04-16
Implement parallel processing for gcsync

tonyxiaowei:
1. create a new task implementing callable and move the logic of starting cloud run tasks and generating rsync instructions there
2. Implemented a greedy based load balancing strategry distribute files into the tasks
3. Use log4j logger instead of java default logger
4. Check if md5 is null (it seems like gcs can give null md5 even when the blob exists in some rare scenarios)
5. Added a parameter for client to specify the number of concurrent tasks to run
6. Give the cloud run tasks 4g memory and 2 cpu by default


vladislav-sidorovich (approved):

dbsync/gcsync/src/main/java/com/google/edwmigration/dbsync/gcsync/GcsyncTask.java
```diff
@@ -0,0 +1,262 @@
...
+    jobsClient.deleteJobAsync(jobName).get();
+  }
+
+  private static List<Checksum> getChecksumsFromFile(InputStream inputStream) throws IOException {
+    List<Checksum> checksums = new ArrayList<>();
+    Checksum checksum;
+    while ((checksum = Checksum.parseDelimitedFrom(inputStream)) != null) {
+      checksums.add(checksum);
+    }
+    inputStream.close();
```
vladislav-sidorovich, line 258: It's a general practice to close a stream where it was open. In this case you use `try with resource` above. so you don't need to close it here.

dbsync/gcsync/src/main/java/com/google/edwmigration/dbsync/gcsync/GcsyncTask.java
```diff
@@ -0,0 +1,262 @@
...
+    for (Path file : filesToRsync) {
+      // Check if we already have an instruction file with a header md5 that matches with the source
+      // file's md5. Meaning the file has been changes since we last generated the instruction file.
+      String sourceFileMd5 = checkNotNull(fileToMd5.get(file));
+      URI instructionFile =
+          new URI(tmpBucket).resolve(Util.getInstructionFileName(file.getFileName().toString()));
+      Blob blob = gcsStorage.getBlob(instructionFile);
+      if (blob != null
+          && verifyMd5Header(gcsStorage.newByteSource(instructionFile), sourceFileMd5)) {
+        logger.info(
```
vladislav-sidorovich, line 167:
```suggestion
        logger.info("Skip generating instructions for file {} which already exists", file));
```

dbsync/gcsync/src/main/java/com/google/edwmigration/dbsync/gcsync/GcsyncClient.java
```diff
@@ -64,54 +51,109 @@ public class GcsyncClient {
...
+   * A comparator that sorts a list of files by its total size.
+   */
+  private static class FileListComparator implements Comparator<List<Path>> {
+
+    @Override
+    public int compare(List<Path> o1, List<Path> o2) {
+      try {
+        long size1 = 0;
+        long size2 = 0;
+        for (Path file : o1) {
```
vladislav-sidorovich, line 142: If I understand correctly, this comparator is `linear` in term of time complexity and will be called at least `n` times, so it makes an algorithm `n^2`.
I'm not sure what is expected files count. but it's possible to make replace `List<Path>` with new private class and keep accumulated size in a filed. It will reduce time complexity for comparator to `O(1)` and for an algorithm to `O(nlogn)`.

dbsync/gcsync/src/main/java/com/google/edwmigration/dbsync/gcsync/GcsyncTask.java
tonyxiaowei: Yay, you're right

dbsync/gcsync/src/main/java/com/google/edwmigration/dbsync/gcsync/GcsyncClient.java
tonyxiaowei: Yes, you're right & I'm aware of this, intentionally left it this one as this won't make any difference unless we start uploading more than 10k files which would be unlikely and this leaves the code slightly cleaner.

I'll make follow up change to improve it. Using Pair<k,v> would likely be sufficient here.

PR #873, opened 2025-04-17
[b/406436446] add kerberos authentication for ranger

kajgol:


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ranger/RangerClient.java
```diff
@@ -105,60 +147,202 @@ public ImmutableList<Object> findPolicies(Map<String, String> args) throws Range
...
+                    queryParams.entrySet().stream()
+                        .map(entry -> new BasicNameValuePair(entry.getKey(), entry.getValue()))
+                        .collect(toImmutableList()))
+                .build()
+                .toURL();
+      } catch (Exception ex) {
+        throw new RangerException("Failed to build URL", ex);
       }
-      HttpEntity entity = httpResponse.getEntity();
+      HttpURLConnection connection = null;
```
zaldis, line 287: Short version should work as well.
```suggestion
      HttpURLConnection connection;
```

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ranger/RangerClient.java
```diff
@@ -105,60 +147,202 @@ public ImmutableList<Object> findPolicies(Map<String, String> args) throws Range
...
+        String basicAuth,
+        SSLSocketFactory socketFactory,
+        HostnameVerifier hostnameVerifier) {
+      this.basicAuth = basicAuth;
+      this.baseUri = baseUri;
+      this.socketFactory = socketFactory;
+      this.hostnameVerifier = hostnameVerifier;
+    }
+
+    public String doGet(String path, Map<String, String> queryParams) {
```
zaldis, line 272: I would suggest to split this method as it's quite long and responsible for different things.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ranger/RangerClient.java
```diff
@@ -105,60 +147,202 @@ public ImmutableList<Object> findPolicies(Map<String, String> args) throws Range
...
-    try (CloseableHttpResponse httpResponse = httpClient.execute(httpRequest)) {
-      logger.debug("Response from Ranger({}): {}", path, httpResponse);
-      if (httpResponse.getStatusLine().getStatusCode() != SC_OK) {
-        throw new RangerException(
-            String.format(
-                "Failed to fetch data from Ranger internal API: '%s'",
-                httpResponse.getStatusLine()));
+
+    if (arguments.useKerberosAuthForHadoop()) {
+      String configurationName = "MyKerberosHttpClient";
```
zaldis, line 181: Could you please explain why configuration name is `MyKerberosHttpClient`? Should not it be something dynamic?

kajgol: No, it should be something specific to the application, but tbh from what I understand we don't really need this in this case, it's just standard good practice. If we were explicitly creating a login context we would need to specify the configuration by name and use the same name then, but in our case it's handled under the hood.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ranger/RangerClient.java
kajgol: It would but with a null check in the finally block I think this is more readable.

vladislav-sidorovich:
I just want to leave it here for history https://en.wikipedia.org/wiki/Cerberus

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ranger/RangerClient.java
kajgol: done

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ranger/RangerClient.java
```diff
@@ -105,60 +147,216 @@ public ImmutableList<Object> findPolicies(Map<String, String> args) throws Range
...
+
+      Configuration customConfig = new KerberosTicketCacheConfig(configurationName);
+
+      // Set this as the default JAAS configuration for the JVM
+      // **Warning:** This is a global setting for the JVM!
+      Configuration.setConfiguration(customConfig);
+
+      System.setProperty("javax.security.auth.useSubjectCredsOnly", "false");
```
vladislav-sidorovich, line 189: I'm curious, is it robust operation? Are there some additional requirements for Security Manager that we should be aware?

kajgol: Yes, if there is a security manager it may block GSS-API operations that are happening under the hood. We could add a check here to see if there is a manager and warn the users they could need to add permissions.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ranger/RangerClient.java
```diff
@@ -105,60 +147,216 @@ public ImmutableList<Object> findPolicies(Map<String, String> args) throws Range
...
+      clientBuilder.withBasicAuth(
+          "Basic "
+              + Base64.getEncoder()
+                  .encodeToString(
+                      String.format("%s:%s", user, password).getBytes(StandardCharsets.UTF_8)));
+    }
+    return clientBuilder.build();
+  }
+
+  static class ConnectionWrapper {
```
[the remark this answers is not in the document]
kajgol: It is low level, but nothing here is for kerberos, the low level stuff is to set https as the protocol, disable tls validation or use basic auth if the user specified that - would KerberosAuthenticator help with that?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ranger/RangerClient.java
kajgol: added check and warning

vladislav-sidorovich (approved):
Let's create a task to try use, some HttpClient.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ranger/RangerClient.java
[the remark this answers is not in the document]
vladislav-sidorovich: It seems we have to work directly with `connection` for Hadoop.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ranger/RangerClient.java
```diff
@@ -105,60 +123,235 @@ public ImmutableList<Object> findPolicies(Map<String, String> args) throws Range
...
       }
-    } catch (IOException e) {
-      throw new RangerException("Failed to connect to Ranger API", e);
     }
-  }
 
-  @Override
-  public void close() throws Exception {
-    httpClient.close();
+    private HttpURLConnection setupConnection(URL url) throws IOException, AuthenticationException {
```
vladislav-sidorovich, line 284: maybe `openAuthenticatedConnection(..)` ?

vladislav-sidorovich (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/ranger/RangerClient.java
kajgol: It's only authenticated if the user specifies they want to use kerberos, otherwise it's a regular connection. This also disables tls validation if needed, so it doesn't just open the connection.

PR #879, opened 2025-04-23
Update README.md

tonyxiaowei:
Update READ me to be more clear and add instruction with the new option

### DIFF
--- a/dbsync/README.md
+++ b/dbsync/README.md
@@ -27,7 +27,13 @@ this [page](https://protobuf.dev/installation/) on how to install it.
 
 ### Run the Gcsync Client ###
 
-    java -jar gcsync-all.jar --target_bucket <target_bucket> --tmp_bucket <tmp_bucket> --project <project_id> --location <location> --source_directory <source_directory> --task_timeout <task_time_out>
+After you successfully compile the source, you will find a fat jar named as gcsync-all.jar under
+
+    dbsync/gcsync/build/libs
+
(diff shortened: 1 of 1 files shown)


dbsync/README.md
```diff
@@ -27,14 +27,27 @@ this [page](https://protobuf.dev/installation/) on how to install it.
...
 ### Run the Gcsync Client ###
 
-    java -jar gcsync-all.jar --target_bucket <target_bucket> --tmp_bucket <tmp_bucket> --project <project_id> --location <location> --source_directory <source_directory> --task_timeout <task_time_out>
+After you successfully compile the source, you will find a fat jar named as gcsync-all.jar under
+
+    dbsync/gcsync/build/libs
+
+You can then run the tool with the following command:
+
+    java -jar gcsync-all.jar --target_bucket <target_bucket> --tmp_bucket <tmp_bucket> --project <project_id> --location <location> --source_directory <source_directory> --task_timeout <task_time_out> –num_concurrent_tasks <num_concurrent_tasks>
```
vladislav-sidorovich, line 36: For Dumper we use Gradle application plugin to create shell/bash scripts to run https://github.com/google/dwh-migration-tools/blob/b598ba4cdb71863fc08d699ad2534742d7b2161c/dumper/app/build.gradle#L165 

You might consider the same option here.

tonyxiaowei: That would be useful. I'll add it as a todo

PR #888, opened 2025-05-12
Clean unused csv in SF Lite

sayuzbas-google:
Removing creating csv with clusters spin on and spin off. This logic is already done inside materialized view.
And there is no ingestion logic for these files.


vladislav-sidorovich (approved):

misolt:
IMO the cleanest solution here would be switching to use these files instead of extracting the entirety of the corresponding time series view. We should focus on extracting the smallest amount of data rather than data kept in the smallest number of CSVs.

OTOH if these files interfere with processing or development in any way then feel free to remove them.

PR #890, opened 2025-05-14
[b/404190134] add permissions-migration

kajgol:


kajgol:
> There are too many files to review. Can we split this PR in 2 steps?
> 
> 1. Add this code as independed module.
> 2. Add the module into the top-level build script.
> 
> So the step 1 will be safe.

Done

vladislav-sidorovich (approved):

PR #903, opened 2025-05-21
Revert an optimization in GcsBytesource (and some automatic formatting changes)

tonyxiaowei:
This optimization improves performance in some scenarios but makes it worse when blocks are swapped.

### DIFF
--- a/dbsync/gcsync/src/main/java/com/google/edwmigration/dbsync/gcsync/GcsyncTask.java
+++ b/dbsync/gcsync/src/main/java/com/google/edwmigration/dbsync/gcsync/GcsyncTask.java
@@ -65,14 +65,18 @@ public class GcsyncTask implements Callable<Void> {
 
   private final String filesToRsyncFileName;
 
-
-  /**
-   * A task that executes cloud run task to rsync the files and then upload the files that
-   */
-  public GcsyncTask(List<Path> filesToRsync, String targetBucket,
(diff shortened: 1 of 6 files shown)


dbsync/gcsync/src/main/java/com/google/edwmigration/dbsync/gcsync/GcsyncTask.java
```diff
@@ -218,8 +221,10 @@ private void runCloudRunJob(String command, String project)
                                 Container.newBuilder()
                                     .setImage("docker.io/getbamba/google-cloud-sdk-java:latest")
                                     .setResources(
-                                        ResourceRequirements.newBuilder().putLimits("memory", "4Gi")
-                                            .putLimits("cpu", "2").build())
+                                        ResourceRequirements.newBuilder()
+                                            .putLimits("memory", "4Gi")
+                                            .putLimits("cpu", "2")
```
vladislav-sidorovich, line 226: Should it be specified by user or at least some configuration?

vladislav-sidorovich (approved):

dbsync/gcsync/src/main/java/com/google/edwmigration/dbsync/gcsync/GcsyncTask.java
tonyxiaowei: We can make it configurable but probably not every important for now as this task is I/O bound, adding more cpu/memory would not yield any real benefits

PR #905, opened 2025-05-22
[b/411631019] Added persistent metrics for dumper runs

kaxuna:
Added SummaryYamlGenerator that takes responsibility of persisting run results


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/Main.java
```diff
@@ -64,7 +64,7 @@ public static void main(String... args) throws Exception {
     try {
       StartUpMetainformationPrinter.printMetainfo();
 
-      Main main = new Main(new MetadataDumper());
+      Main main = new Main(new MetadataDumper(new SummaryYamlGenerator()));
```
vladislav-sidorovich, line 67: Is any specific reason to use Yaml?

Let's use JSON instead, it's good enough and we have all tools to work with it.

kaxuna: There is a good reason why Yaml was chosen (In fact Json was used originally)

Appending to Json Array needs :

1. Checking if it is a initial log to open an array
2. Every time logs are appended
    a. last token  "]" should be deleted 
    b.  "," + new json should be appended 
    c.  "]" should be reappended in the end

while in Yaml we simply append new logs without care

This kind of string manipulation is of course doable but gives no real advantage especially when both assessment service and dumper already has tools to handle yaml files

misolt:
Seems like this is a new feature - is there a design that I could compare this PR with?

Without that I can't give a good review, because I don't have a clear idea of how this is intended to work.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/Main.java
vladislav-sidorovich: I see, you did a lot of work to build the structure. It can be simplified. Let's discuss it in-person.

dumper/app/gradle.lockfile
```diff
@@ -170,6 +170,8 @@ junit:junit:4.13.2=testCompileClasspath,testRuntimeClasspath
 net.bytebuddy:byte-buddy-agent:1.12.19=testCompileClasspath,testRuntimeClasspath
 net.bytebuddy:byte-buddy:1.12.19=testCompileClasspath,testRuntimeClasspath
 net.java.dev.jna:jna:5.17.0=runtimeClasspath,testFixturesRuntimeClasspath,testRuntimeClasspath
+net.java.dev.jna:jna-platform:5.16.0=compileClasspath,runtimeClasspath,testFixturesRuntimeClasspath,testRuntimeClasspath,testCompileClasspath
```
vladislav-sidorovich, line 173: You excluded this deps but add it to the lock file. It's unusual. Please double check

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/DumperRunMetricsGenerator.java
```diff
@@ -0,0 +1,179 @@
...
+        .setTaskExecutionSummary(taskExecutionSummaries)
+        .setTaskDetailedSummary(taskDetailedSummaries)
+        .setArguments(arguments)
+        .build();
+  }
+
+  /** Appends the given summary lines for the current run to the external log file. */
+  private void appendToCacheOnDisk(Path externalLogPath, String summaryLines) {
+    try (BufferedWriter writer =
+            java.nio.file.Files.newBufferedWriter(
```
dominik-klimczak, line 166: why not import Files?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/DumperRunMetricsGenerator.java
```diff
@@ -0,0 +1,179 @@
...
+  private static void copyCachedAsCurrent(FileSystem zipFs, Path externalLogPath) {
+    Path snapshotInZipPath = zipFs.getPath(DUMPER_RUN_METRICS);
+    try {
+      Path parentInZip = snapshotInZipPath.getParent();
+      if (parentInZip != null && java.nio.file.Files.notExists(parentInZip)) {
+        java.nio.file.Files.createDirectories(parentInZip);
+      }
+      java.nio.file.Files.copy(
+          externalLogPath, snapshotInZipPath, StandardCopyOption.REPLACE_EXISTING);
+      logger.info(
```
dominik-klimczak, line 110: IMO debug is better - this is not a message that the user is interested in

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/DumperRunMetricsGenerator.java
```diff
@@ -0,0 +1,179 @@
...
+          "Copied cumulative run summary from {} to {} in the output ZIP.",
+          externalLogPath,
+          snapshotInZipPath);
+    } catch (IOException e) {
+      logger.error(
```
dominik-klimczak, line 115: IMO warn is better - this is not an error in extraction

dominik-klimczak (approved):

PR #910, opened 2025-05-27
[bug] fix NPE for Oozie end date

vladislav-sidorovich:
Handle null end date.

We collect only jobs the already ended and sort the jobs history by end date. So, if the end date is null for a job, this job should be not be added to the output and processing should go to the next job.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hadoop/oozie/AbstractOozieJobsTask.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hadoop/oozie/AbstractOozieJobsTask.java
@@ -34,6 +34,7 @@ import java.util.Date;
 import java.util.List;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import org.apache.commons.beanutils.PropertyUtils;
 import org.apache.commons.csv.CSVFormat;
 import org.apache.commons.csv.CSVPrinter;
@@ -100,7 +101,8 @@ public abstract class AbstractOozieJobsTask<J> extends AbstractTask<Void> {
         for (J job : jobs) {
           Date currentJobEndTime = getJobEndTime(job);
           boolean inDateRange =
-              minJobEndTimeTimestamp <= currentJobEndTime.getTime()
+              currentJobEndTime != null
+                  && minJobEndTimeTimestamp <= currentJobEndTime.getTime()
                   && currentJobEndTime.getTime() < maxJobEndTimeTimestamp;
           if (!inDateRange) {
             // It's client side filtering. It's inefficient.
@@ -147,6 +149,7 @@ public abstract class AbstractOozieJobsTask<J> extends AbstractTask<Void> {
   abstract List<J> fetchJobsWithFilter(
       XOozieClient oozieClient, String oozieFilter, int start, int len) throws OozieClientException;
 
+  @Nullable
   abstract Date getJobEndTime(J job);
 
   private static String toISO(ZonedDateTime dateTime) {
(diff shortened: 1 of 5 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hadoop/oozie/AbstractOozieJobsTask.java
```diff
@@ -120,10 +122,10 @@ protected Void doRun(TaskRunContext context, @Nonnull ByteSink sink, @Nonnull Ha
 
         J lastJob = jobs.get(jobs.size() - 1);
         Date endTime = getJobEndTime(lastJob);
-        if (endTime == null) {
-          break;
+        boolean isLastJobInProgress = endTime == null;
```
tonyxiaowei, line 125: The style is kind of weird. This boolean variable seems to be used for 1th only.

Why not.

if(endTime != null){
  xxx
}

vladislav-sidorovich: To explain what `null` means in this case and why it's ok to skip this job.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hadoop/oozie/AbstractOozieJobsTask.java
```diff
@@ -100,7 +101,8 @@ protected Void doRun(TaskRunContext context, @Nonnull ByteSink sink, @Nonnull Ha
         for (J job : jobs) {
           Date currentJobEndTime = getJobEndTime(job);
           boolean inDateRange =
```
tonyxiaowei, line 103: Why do we need this boolean if it's only used 1th?

why not 

if(currentJobEndTime != null && minJobEndTimeTimestamp <= currentJobEndTime.getTime()
                  && currentJobEndTime.getTime() < maxJobEndTimeTimestamp){
}

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/hadoop/oozie/OozieWorkflowJobsTaskTest.java
```diff
@@ -189,6 +190,55 @@ public void doRun_requestBatchesFilterOnClient_ok() throws Exception {
     assertEquals(expected, actual);
   }
 
+  @Test
+  public void doRun_requestBatchesFilterOnClient_nullDate_success() throws Exception {
+    final ZonedDateTime endTime =
+        ZonedDateTime.ofInstant(Instant.ofEpochMilli(timestampInMockResponses), UTC).plusHours(1);
+    final ZonedDateTime startTime = endTime.minusDays(7);
+    Date lastCapturedDate = Date.from(startTime.toInstant());
+
```
tonyxiaowei, line 199: Nit: I think you can remove this empty line to group setups together.

tonyxiaowei (approved):

dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/hadoop/oozie/OozieWorkflowJobsTaskTest.java
```diff
@@ -189,6 +190,55 @@ public void doRun_requestBatchesFilterOnClient_ok() throws Exception {
...
+    OozieWorkflowJobsTask task = new OozieWorkflowJobsTask(startTime, endTime);
+
+    // Act
+    task.doRun(context, sink, new OozieHandle(oozieClient));
+
+    // Assert
+    String actual = sink.getContent();
+    String expected = readFileAsString("/oozie/expected-jobs-one-job-from-template.csv");
+    assertEquals(expected, actual);
```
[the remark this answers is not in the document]
vladislav-sidorovich: yes, it's possible. To check full result is ok as well.

Taking into account that all the methods to check full content are in place, this approach is simpler.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/hadoop/oozie/AbstractOozieJobsTask.java
vladislav-sidorovich: to give a name to expression. It's always preferable to extract a variable with reasonable name for such expressions compare to a comment. For trivial expressions such practice can be ignore. 

I think this is a borderline case because the names of variables are long, so I followed the practice and extracted the variable

PR #926, opened 2025-07-25
Filter databases for snowflake

bjankie1:
Limited to metadata connector

Observation 1: --database for metadata parameter MUST always contain a single database name despite the fact that some parts of the code expect that to be a list of database to filter with

Observation 2: if SNOWFLAKE.ACCOUNT_USAGE is accessible with the role that the dumper is run with then: --database parameter is used only to filter external tables https://github.com/google/dwh-migration-tools/blob/c7795917cf86e41e59165c137d0610532d3e9593/dumper/app/src/test/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnectorTest.java#L202

Observation 3: If SNOWFLAKE.ACCOUNT_USAGE is NOT accessible with the role that the dumper is run with then INFORMATION_SCHEMA is used and --database is used first to select a database for accessing INFORMATION_SCHEMA, and then to filter results by this database.

Observation 4: All Assessment specific queries in SnowflakePlanner.java do not respect any filtering

Solution:

Make --database optional. Select a database dynamically if no database is not provided (for INFORMATION_SCHEMA access purpose).
Create a separate optional parameter --limit-to-databases=X,Y,Z for both connectors that will limit the scope of extracted data
Replace all filtering logic that uses --database parameter (arguments.getDatabase()) with --limit-to-databases
Extract external tables respecting the --limit-to-databases parameter. If it is not provided then extract for all database (SHOW DATABASES).
Add filtering to SnowflakePlanner queries by respecting --limit-to-databases parameter.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
@@ -103,6 +103,7 @@ public class ConnectorArguments extends DefaultArguments {
   public static final String OPT_WAREHOUSE = "warehouse";
   public static final String OPT_DATABASE = "database";
   public static final String OPT_SCHEMA = "schema";
+  public static final String OPT_LIMIT_TO_DATABASES = "limit-to-databases";
   public static final String OPT_OUTPUT = "output";
   public static final String OPT_CONFIG = "config";
   public static final String OPT_ASSESSMENT = "assessment";
@@ -216,6 +217,13 @@ public class ConnectorArguments extends DefaultArguments {
   private final OptionSpec<String> optionDatabase =
       parser
           .accepts(OPT_DATABASE, "Database(s) to export")
+          .withRequiredArg()
+          .ofType(String.class)
+          .withValuesSeparatedBy(',')
+          .describedAs("db0,db1,...");
+  private final OptionSpec<String> optionLimitToDatabases =
+      parser
+          .accepts(OPT_LIMIT_TO_DATABASES, "Limit data extraction to specific databases")
           .withRequiredArg()
           .ofType(String.class)
           .withValuesSeparatedBy(',')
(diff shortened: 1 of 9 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeLogsConnector.java
```diff
@@ -333,6 +359,15 @@ private String createExtendedQueryFromAccountUsage(ConnectorArguments arguments)
                 + "WHERE end_time >= to_timestamp_ltz('%s')\n"
                 + "AND end_time <= to_timestamp_ltz('%s')\n"
                 + "AND is_client_generated_statement = FALSE\n");
+    
+    // Add database filtering if limit-to-databases is provided
```
dominik-klimczak, line 363: you can extract this block to a method - you will remove the repetition and you can use meaningful name instead of a comment

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakePlanner.java
```diff
@@ -62,16 +65,53 @@ private enum Format {
...
 
   ImmutableList<Task<?>> generateLiteSpecificQueries() {
+    return generateLiteSpecificQueries(null);
+  }
+
+  ImmutableList<Task<?>> generateLiteSpecificQueries(@Nullable ConnectorArguments arguments) {
     String view = "SNOWFLAKE.ACCOUNT_USAGE";
     String filter = " WHERE DELETED IS NULL";
+    
+    // Add database filtering if limit-to-databases is provided
```
dominik-klimczak, line 107: This seems like yet another instance of the pattern repeated in the other file (although with a string concatenation instead of a builder). Maybe you can just create a helper function that return Optional<String> that can be applied to all these instances.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -103,6 +103,7 @@ public class ConnectorArguments extends DefaultArguments {
   public static final String OPT_WAREHOUSE = "warehouse";
   public static final String OPT_DATABASE = "database";
   public static final String OPT_SCHEMA = "schema";
+  public static final String OPT_LIMIT_TO_DATABASES = "limit-to-databases";
```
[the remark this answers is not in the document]
kaxuna: nit: rename to: databases-limited, databases-with-limit

IMO limit-to-database suggest that this attribute is helper to OPT_DATABASE.

PR #933, opened 2025-08-19
[b/440295519] Create gradle build & test setup

kaxuna:
Workflow for testing and building

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -1,0 +1,52 @@
+name: Java CI with Gradle
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/gradle.yml
```diff
@@ -0,0 +1,52 @@
...
+    - name: Test
+      run: ./gradlew :dumper:app:test
+      
+  dependency-submission:
+    name: create-dependabot-graph
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+
+    steps:
```
misolt, line 41: Don't we need to set up Gradle here as well?

vladislav-sidorovich (approved):

PR #938, opened 2025-08-21
[b/440295519] Github actions release script

kaxuna:
v1 for release script


vladislav-sidorovich:
Looks good but we also need license check and update a pipeline. Let's mark it as draft for now and do not use.

PR #945, opened 2025-08-26
[b/440495128] Add Oracle TCO discovery script

kpiotrowski:
This PR adds Oracle TCO discovery script that is an alternative method of extracting Oracle metadata and statistics for BigQuery Migration Assessment.

Script uses the same SQL files as dwh-migration-dumper, but executes them with SQL*Plus. This is helpful for the users who cannot run java application due to security or compliance reasons. Currently, it extracts only set of metadata and statistics required to TCO calculation. Later this can be extended with full version.


vladislav-sidorovich (approved):
I like it.

misolt (approved):

warehouse-discovery-scripts/oracle/README.md
```diff
@@ -0,0 +1,46 @@
+# 💻 Oracle Database Discovery Scripts
```
misolt, line 1: Maybe this can also be added as a page on the [Wiki](https://github.com/google/dwh-migration-tools/wiki)?

kpiotrowski: Yes, I think this is a good idea. Let me verify this solution with the customer first and I'll update wiki after that.

PR #971, opened 2025-09-18
Streaming telemetry implementation

kaxuna:


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/DiskTelemetryWriteStrategy.java
```diff
@@ -0,0 +1,187 @@
...
+
+  @Override
+  public void flush() {
+    // this implementation uses buffer until zip file is created afterwords it is flushed per
+    // process
+    flushOsCache();
+    flushZip();
+  }
+
+  private void flushOsCache() {
```
misolt, line 101: For both `flushOsCache` and `flushZip`:
- The value on the left of `&&` doesn't get updated in the loop. It can be replaced with an early return:

```java
if (fileSystem == null) {
  return;
}
// loop follows
```
- A for-each loop and then `clear` can be used instead of `poll` in a `while` loop.
- After the above change the dequeues can be replaced with lists.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/DiskTelemetryWriteStrategy.java
```diff
@@ -0,0 +1,187 @@
...
+  private final Queue<ClientTelemetry> bufferOs;
+  private final Queue<ClientTelemetry> bufferZip;
+  private FileSystem fileSystem;
+  private boolean telemetryOsCacheIsAvailable = true;
+
+  public DiskTelemetryWriteStrategy() {
+    bufferOs = new ArrayDeque<>();
+    bufferZip = new ArrayDeque<>();
+    MAPPER = createObjectMapper();
+    telemetryOsCachePath = Paths.get(createTelemetryOsDirIfNotExists(), ALL_DUMPER_RUN_METRICS);
```
misolt, line 58: In general it's better to avoid doing work in constructors, especially if there are significant side effects, like creating a directory in this case.

For example, this could be a constructor parameter so that tests can mock it.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/DiskTelemetryWriteStrategy.java
```diff
@@ -0,0 +1,187 @@
+/*
```
misolt, line 1: Is there a matching test for this class?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/metrics/ClientTelemetry.java
```diff
@@ -20,35 +20,94 @@
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.UUID;
 
 public class ClientTelemetry {
 
+  // Uniquer run Id. Is Set by the TelemetryProcessor.
   @JsonProperty private String id;
```
vladislav-sidorovich, line 27: Let's use more meaningful name.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/TasksRunner.java
```diff
@@ -93,10 +99,41 @@ public <T> T runChildTask(@Nonnull Task<T> task) throws MetadataDumperUsageExcep
...
 
+  private void processTaskStartTelemetry(Task<?> task, String taskEventId) {
+    TaskRunMetrics taskRunMetrics =
+        new TaskRunMetrics(task.getName(), task.getClass().toString(), null, null);
+
+    telemetryProcessor.process(
+        ClientTelemetry.builder()
+            .setEventId(taskEventId)
+            .setEventType(EventType.TASK_RUN_START)
+            .setPayload(Collections.singletonList(taskRunMetrics))
```
vladislav-sidorovich, line 120: Why do you need an immutable list of 1 object ?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/metrics/ClientTelemetry.java
```diff
@@ -20,35 +20,94 @@
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.UUID;
 
 public class ClientTelemetry {
```
vladislav-sidorovich, line 24: Let's name it somehow with `*event`

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/metrics/ClientTelemetry.java
```diff
@@ -20,35 +20,94 @@
...
   @JsonProperty private String id;
 
+  @JsonProperty private String eventId;
+
   @JsonProperty private EventType eventType;
 
-  @JsonProperty private ZonedDateTime dumperRunStartClientTime;
+  @JsonProperty private ZonedDateTime timestamp;
 
   @JsonProperty private List<TelemetryPayload> payload;
```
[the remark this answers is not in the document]
vladislav-sidorovich: @Data
public class TelemetryEvent {
    String eventId = UUID.randomUUID().toString();
    String dumperRunId;
    String eventType;
    long timestamp;
    TelemetryPayload payload;

}

PR #979, opened 2025-09-22
[b/446687185] Extract new metrics from Snowflake

piechotam:
Extract metrics from the `QUERY_ATTRIBUTION_HISTORY` view.
Extract `warehouse_size` from the `QUERY_HISTORY` view.

These metrics will be used by the ML model developed in the project [b/432667540].


kpiotrowski (approved):
LGTM.  I've tested this change together with ingestion and it worked as expected.

PR #1030, opened 2025-11-21
[b/462688456] Add known CVEs files and schema.

vladislav-sidorovich:
We already have some know CVEs in the project. These CVEs must be explicitly mentioned in this file with a justification. 

Both `Yaml` and `Json` formats can be supported with the single JSONSchema. In this example I use `yml` because it seems more human redable in case of multiple lines `justification`.

Examples:
```
# yaml-language-server: $schema=./known_cves_schema.json

- CVE: CVE-2025-52999
  artifact: org.example:vulnerability-lib:3.18.0
  justification: |
    Some text
    with very nice and clear explanation
  expiration_date: 2030-05-18
```

```
[
  {
    "CVE": "CVE-2025-52999",
    "artifact": "org.apache.commons:commons-lang3:3.18.0",
    "justification": "We can and we do",
    "expiration_date": "ISO something"
  },
...
]
```

### DIFF
--- a/.security/known_cves.yml
+++ b/.security/known_cves.yml
@@ -1,0 +1,8 @@
+# yaml-language-server: $schema=./known_cves_schema.json
+
+- CVE: CVE-2025-52999
+  artifact: org.example:vulnerability-lib:3.18.0
+  justification: >
+    Some text
+    with very nice and clear explanation
+  expiration_date: 2030-05-18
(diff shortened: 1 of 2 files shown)


sayuzbas-google (approved):

vladislav-sidorovich:
> Looks good, though it would be nice to add integration with an automated tool for detecting vulnerabilities.
> 
> I'd like to have the option to run a script that will check the project and show me only the findings that are not already excluded.

Yes, exactly it's the next step after defining the schema.

PR #1053, opened 2025-12-16
Make AbstractTaskTest a test instead of util

misolt:
AbstractTaskTest isn't the test of AbstractTask. Instead, it's a utility class where the helpers are accessed via inheritance.

Make AbstractTaskTest a test of AbstractTask.

- Inline the `HEADER` constant.
- Move MemoryByteSink from the util to AbstractTask.
- Remove the now-redundant inheritance of AbstractTaskTest by other test classes and make it a test.

Currently the class is empty, but tests will be added in future PRs.


vladislav-sidorovich (approved):
I like the simplification, thank you.

PR #1113, opened 2026-04-21
Add labeler

vladislav-sidorovich:
Add labeler to mark PR. Let's discuss labels

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -31,6 +31,13 @@ Dumper:
   - changed-files:
     - any-glob-to-any-file: 'dumper/**/*'
 
+Dumper CORE:
+  - changed-files:
+      - any-glob-to-any-file:
+          - 'dumper/**/*.java'
+          - '!dumper/app/src/*/java/com/google/edwmigration/dumper/application/dumper/connector/*/**/*'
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,72 @@
...
+      - any-glob-to-any-file:
+          - 'docs/**/*'
+          - '**/*.md'
+
+Client:
+  - changed-files:
+      - any-glob-to-any-file: 'client/**/*'
+Dumper:
+  - changed-files:
+    - any-glob-to-any-file: 'dumper/**/*'
```
misolt, line 32: Let's also add a Dumper core label, which would be applied if there was any change of Dumper code outside the data-source-specific directories: `connector/airflow`, `connector/hive`, etc.

misolt (approved):

.github/labeler.yml
misolt: The reasoning is that if a PR has this label, the review should be extra careful because it has the potential to break more than just one thing.

vladislav-sidorovich: Sounds good to me

vladislav-sidorovich: Done

PR #1156, opened 2026-06-29
[Snowflake] Allow list of databases to be passed to --database flag

tanclary:
Generate a query for each database's info schema, allowing the --database flag to accept a list rather than just one at a time.

We are doing this to reduce customer toil. Previously if they wanted to use the database filter they could only use one at a time, meaning the metadata could not be extracted in a single run. This change makes that more feasible.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnector.java
@@ -152,7 +152,7 @@ public class SnowflakeMetadataConnector extends AbstractSnowflakeConnector
     String globalDatabaseFilter =
         getInformationSchemaWhereCondition(databaseFilterColumnName, databases);
     AbstractJdbcTask<Summary> usageTask =
-        SnowflakeTaskUtil.withFilter(
+        SnowflakeTaskUtil.createJdbcSelectTask(
             format,
             ACCOUNT_USAGE_SCHEMA_NAME,
             accountUsageFileName,
@@ -165,7 +165,7 @@ public class SnowflakeMetadataConnector extends AbstractSnowflakeConnector
 
     if (databases.isEmpty()) {
       AbstractJdbcTask<Summary> schemaTask =
-          SnowflakeTaskUtil.withFilter(
+          SnowflakeTaskUtil.createJdbcSelectTask(
               format,
               "INFORMATION_SCHEMA",
               informationSchemaFileName,
@@ -182,6 +182,14 @@ public class SnowflakeMetadataConnector extends AbstractSnowflakeConnector
     }
     if (inputSource == SnowflakeInput.SCHEMA_ONLY_SOURCE
         || inputSource == SnowflakeInput.USAGE_THEN_SCHEMA_SOURCE) {
+      // INFORMATION_SCHEMA is database-scoped. To fetch metadata from multiple databases,
+      // we must query each database's INFORMATION_SCHEMA individually (e.g.,
+      // db.INFORMATION_SCHEMA.TABLES).
+      // Prefixing the database name also helps the query optimizer scope the metadata scan,
+      // avoiding performance issues that occur when querying INFORMATION_SCHEMA without a database
+      // scope.
+      //
+      // The first task overwrites the output file; subsequent tasks append to it.
       TaskOptions taskOptions = TaskOptions.DEFAULT;
       for (String database : databases) {
         String schemaPrefix = sanitizeDatabaseName(database) + ".INFORMATION_SCHEMA";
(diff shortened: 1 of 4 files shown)


kaxuna (approved):
lgtm

shevek-google:
It would be nice to record _why_ we made this change, not just that we made it, probably in the code somewhere. Did we overflow a max query length in an IN() clause? Did we prefer to concat the database name as a bareword than to do string comparison?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnector.java
```diff
@@ -214,12 +249,10 @@ public final void addTasksTo(
...
             MetadataView.TABLES),
         TablesFormat.IS_ZIP_ENTRY_NAME,
-        schemaPrefix,
         TablesFormat.AU_ZIP_ENTRY_NAME,
         ACCOUNT_USAGE_WHERE_CONDITION,
-        isAssessment,
-        getInformationSchemaWhereCondition(
-            "table_catalog", arguments.getDatabases())); // Painfully slow.
+        arguments,
+        "table_catalog");
```
shevek-google, line 263: Is it still painfully slow? Why?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnector.java
```diff
@@ -142,30 +143,67 @@ private void addSqlTasksWithInfoSchemaFallback(
...
+    }
+
+    List<Task<?>> tasks = new ArrayList<>();
+    if (inputSource == SnowflakeInput.USAGE_ONLY_SOURCE
+        || inputSource == SnowflakeInput.USAGE_THEN_SCHEMA_SOURCE) {
+      tasks.add(usageTask);
+    }
+    if (inputSource == SnowflakeInput.SCHEMA_ONLY_SOURCE
+        || inputSource == SnowflakeInput.USAGE_THEN_SCHEMA_SOURCE) {
+      TaskOptions taskOptions = TaskOptions.DEFAULT;
```
shevek-google, line 193: If this is DEFAULT here, and we never pass any other value, why did we add the extra overload in SnowflakeTaskUtil?

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeTaskUtil.java
```diff
@@ -37,8 +39,20 @@ static AbstractJdbcTask<Summary> withFilter(
       String zipEntryName,
       Collection<String> whereConditions,
       Class<? extends Enum<?>> header) {
+    return withFilter(
+        format, schemaName, zipEntryName, whereConditions, header, TaskOptions.DEFAULT);
+  }
+
+  static AbstractJdbcTask<Summary> withFilter(
```
shevek-google: A 'with' name implies a mutation of an existing object, whereas this looks like a static factory method, which might be better exposed as two: newSql...() and newJdbcSelectTask(...).

shevek-google (approved):
Approved with definite observations. :) Do as you will.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeTaskUtil.java
tanclary: Ack. Renamed to `createJdbcSelectTask`

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnector.java
tanclary: It actually can change, It starts as  DEFAULT  for the very first task so we overwrite the file if it exists, but for the subsequent loops we update it to APPEND_EXISTING(via  taskOptions =           taskOptions.withWriteMode(WriteMode.APPEND_EXISTING) ). So the helper method does end up receiving a non-default value

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnector.java
tanclary: By prefixing the database name (e.g.,  db.INFORMATION_SCHEMA.TABLES ), it scopes the metadata scan and completely avoids the performance hit we'd get from querying  INFORMATION_SCHEMA  without a scope. I've added a block comment in the code explaining this.

misolt:
Hi, I see that this is already reviewed - just letting you know that we're having an issue in integration tests where they don't start automatically for some authors.

We're working on resolving this, in the meantime you can use the label `kokoro:force-run` to run the required tests before merging.

PR #1158, opened 2026-07-15
Fix: Exclude custom queries from metadata

zaldis:
This change excludes custom queries from `compilerworks-metadata.yaml` file. Custom queries are not used now in the assessment pipeline.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
@@ -1157,6 +1157,15 @@ public class ConnectorArguments extends DefaultArguments {
   @Override
   @Nonnull
   public String toString() {
+    return buildToString(/* excludeQueries= */ false);
+  }
+
+  @Nonnull
+  public String toStringWithoutCustomQueries() {
+    return buildToString(/* excludeQueries= */ true);
+  }
+
+  private String buildToString(boolean excludeQueries) {
     // We do not include password here b/c as of this writing,
     // this string representation is logged out to file by ArgumentsTask.
     ToStringHelper toStringHelper =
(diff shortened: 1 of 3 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
```diff
@@ -1178,7 +1187,9 @@ public String toString() {
             .add(OPT_SPARK_HISTORY_SERVICE_NAMES, getSparkHistoryServiceNames())
             .add(OPT_ASSESSMENT, isAssessment())
             .add(OPT_TELEMETRY, isTelemetryOn());
-    getConnectorProperties().getDefinitionMap().forEach(toStringHelper::add);
+    getConnectorProperties().getDefinitionMap().entrySet().stream()
+        .filter(entry -> !excludeQueries || !entry.getKey().contains(".query"))
+        .forEach(entry -> toStringHelper.add(entry.getKey(), entry.getValue()));
```
sayuzbas-google, line 1192: nit: will it work, as had previously?
```
forEach(toStringHelper::add);
```

sayuzbas-google (approved):

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/ConnectorArguments.java
zaldis: It would result in a compilation error:

1. The stream is iterating over a collection of `Map.Entry` objects, so `forEach` passes a single `Map.Entry` argument to its consumer.
2. Using the method reference `toStringHelper::add` would tell the compiler to look for an add method on `ToStringHelper` that takes exactly one parameter (a Map.Entry or an Object).
3. However, Guava's `ToStringHelper.add()` method requires two parameters (String name, Object value).

PR #1159, opened 2026-07-21
Fixing features.csv file logic for appending values

sayuzbas-google:
Fixing b/534669263.

Fixed WriteMode for features sqls, which are used in features.csv. For the first sql:`SIMPLE("account-usage-simple.sql")` it should be default write mode, for the rest append to existing.


kaxuna (approved):
LGTM

PR #1161, opened 2026-07-24
[Snowflake] Support schema filtering in Snowflake metadata connector

jin-jj:
1. Add support for --schema "schema1, schema2"
2. Fix a minor issue that we add extra database where condition for information_schema query.

### DIFF
--- a/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnector.java
+++ b/dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnector.java
@@ -29,6 +29,7 @@ import com.google.edwmigration.dumper.application.dumper.ConnectorArguments;
 import com.google.edwmigration.dumper.application.dumper.annotations.RespectsArgumentAssessment;
 import com.google.edwmigration.dumper.application.dumper.annotations.RespectsArgumentDatabaseForConnection;
 import com.google.edwmigration.dumper.application.dumper.annotations.RespectsArgumentDatabasePredicate;
+import com.google.edwmigration.dumper.application.dumper.annotations.RespectsArgumentSchemaPredicate;
 import com.google.edwmigration.dumper.application.dumper.connector.Connector;
 import com.google.edwmigration.dumper.application.dumper.connector.ConnectorProperty;
 import com.google.edwmigration.dumper.application.dumper.connector.MetadataConnector;
@@ -57,6 +58,7 @@ import javax.annotation.Nonnull;
 @RespectsArgumentAssessment
 @RespectsArgumentDatabaseForConnection
 @RespectsArgumentDatabasePredicate
+@RespectsArgumentSchemaPredicate
 public class SnowflakeMetadataConnector extends AbstractSnowflakeConnector
     implements MetadataConnector, SnowflakeMetadataDumpFormat {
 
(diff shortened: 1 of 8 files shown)


dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnector.java
```diff
@@ -327,16 +340,42 @@ public final void addTasksTo(
...
+        String quotedName = identifierNameQuoted(schema);
+        AssessmentQuery query = planner.externalTablesInSchema(quotedName);
+        Task<?> task =
+            new JdbcSelectTask(
+                    query.zipEntryName, query.formatString, TaskCategory.REQUIRED, taskOptions)
+                .withHeaderTransformer(query.transformer());
+        out.add(task);
+        taskOptions = taskOptions.withWriteMode(WriteMode.APPEND_EXISTING);
+      }
+    } else {
```
mshetairy: Problem **#1 Cartesian Product & Task Failures**.

dumper/app/src/main/java/com/google/edwmigration/dumper/application/dumper/connector/snowflake/SnowflakeMetadataConnector.java
```diff
@@ -327,16 +340,42 @@ public final void addTasksTo(
...
+        AssessmentQuery query = planner.externalTablesInDatabase(quotedName);
+        Task<?> task =
+            new JdbcSelectTask(
+                    query.zipEntryName, query.formatString, TaskCategory.REQUIRED, taskOptions)
+                .withHeaderTransformer(query.transformer());
+        out.add(task);
+        // Next tasks will append to the same file.
+        taskOptions = taskOptions.withWriteMode(WriteMode.APPEND_EXISTING);
+      }
+    } else if (databases.isEmpty() && !schemata.isEmpty()) {
```
mshetairy: Problem **#2 Unqualified Schema Scoping when `--database` is omitted** .

mshetairy:
Thanks for working on this! The overall wiring for `@RespectsArgumentSchemaPredicate` and the `WHERE` filters on `ACCOUNT_USAGE` / `INFORMATION_SCHEMA` looks great. 

I did notice two critical risks around how `SHOW EXTERNAL TABLES` is handled, along with some comment cleanup, that we should address before merging:

### 1. Cartesian Product & Task Failures (`addTasksTo`, lines 366-378)
When both `--database` and `--schema` are provided, the loop creates a separate task for every possible pair (`SHOW EXTERNAL TABLES IN SCHEMA "DB_i"."SCHEMA_j"`). 
* **Task Failures:** In Snowflake, running `SHOW ... IN SCHEMA "DB1"."SCHEMA_B"` throws a SQL compilation error if `SCHEMA_B` belongs to `DB2` and doesn't exist in `DB1`. Because this pairs every database with every schema in the filter lists, any schema not present in *every* filtered database will cause JDBC task failures and pollute the error logs.
* **Scaling:** A filter of 10 databases and 20 schemas explodes into 200 separate JDBC tasks.
* **Suggestion:** Instead of generating schema-scoped `SHOW` commands in a nested loop, keep executing `SHOW EXTERNAL TABLES IN DATABASE "DB_i"` (one task per database, as before) and filter the resulting rows by schema in memory using the `HeaderTransformer`.

### 2. Unqualified Schema Scoping when `--database` is omitted (`addTasksTo`, lines 355-365)
When `--database` is empty and only `--schema` is used, this generates `SHOW EXTERNAL TABLES IN SCHEMA "SCHEMA1"`. Because `"SCHEMA1"` isn't qualified with a database prefix, Snowflake will only search inside the **currently selected default database of the JDBC session**. If no default database is set on the connection—or if `"SCHEMA1"` resides in a different database—this query will fail or silently miss external tables across the rest of the account.

### 3. Comment Cleanup (`SnowflakeMetadataConnector.java`)
It looks like there are a few AI assistant comment artifacts left over from generating the code modifications (e.g., `// Changed to EMPTY_WHERE_CONDITION` and `// Changed to String literal`). Let's be sure to clean those inline comments up before merging so the codebase stays clean!

kaxuna:
lint fails. apply spotlessApply

jin-jj:
1. Added a in-memory filter/predict mechanism for jdbc task
2. Did the spotless apply

kaxuna (approved):

PR #1166, opened 2026-08-04
[Snowflake] Introduce a flag to filter out databases that only contain cloned tables.

jin-jj:
1. Adds the --ignore-clone-only-database argument to skip databases with only cloned tables.

2. Implements a fallback check when the argument isn't used, which logs a warning and suggests the flag if clone-only databases are found.


kaxuna (commented):
LGTM

kaxuna (approved):

PR #1169, opened 2026-08-28
Remove Databricks README section

mingjialiu:
Remove the Databricks-specific section from the Dumper README so the documentation reflects the current supported content.

The change removes the outdated Unity Catalog and Hive Metastore usage examples.


kaxuna:
@mingjialiu 

Is there a design doc for this? We do have discovery-notebook for databricks extraction. What is the reason for duplication?

